PR #150, opened 2024-05-21
* cleanTables now uses truncate instead of deleteMany

Dosexe:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/prisma-utils/test/DbCleaner.ts
+++ b/packages/app/prisma-utils/test/DbCleaner.ts
@@ -2,15 +2,13 @@ import type { PrismaClient } from '@prisma/client'
 
 export class DbCleaner<DbModel extends object> {
 	private readonly schema: string
-	private readonly tableNames: string[]
 
-	constructor(schema: string, dbModel: DbModel) {
+	constructor(schema: string) {
 		this.schema = schema
-		this.tableNames = Object.values(dbModel)
 	}
 
-	async cleanTables(prisma: PrismaClient) {
-		for (const table of this.tableNames) {
+	async cleanTables(prisma: PrismaClient, dbModel: DbModel) {
+		for (const table of Object.values(dbModel)) {
 			await prisma.$executeRawUnsafe(`TRUNCATE TABLE ${this.schema}.${table} CASCADE;`)
 		}
 	}
--- a/packages/app/prisma-utils/test/DbCleaner.ts
+++ b/packages/app/prisma-utils/test/DbCleaner.ts
@@ -5,12 +5,12 @@ export enum DB_MODEL {
 	item2 = 'item2',
 }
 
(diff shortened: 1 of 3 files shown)


packages/app/prisma-utils/test/DbCleaner.ts
```diff
@@ -5,12 +5,12 @@ export enum DB_MODEL {
...
-export async function cleanTables(prisma: PrismaClient, modelNames: readonly DB_MODEL[]) {
-	const delegates = modelNames.map<{ deleteMany: () => Promise<unknown> }>(
-		(modelName) => prisma[modelName],
-	)
-
-	for (const delegate of delegates) {
-		await delegate.deleteMany()
+export async function cleanTables(
+	prisma: PrismaClient,
+	modelNames: readonly DB_MODEL[],
```
Dosexe, line 10: @kibertoad Technically we can make `DB_MODEL` generic and use function from lib instead of copying it every time, WDYT?

packages/app/prisma-utils/test/DbCleaner.ts
```diff
@@ -5,12 +5,12 @@ export enum DB_MODEL {
...
 }
 
-export async function cleanTables(prisma: PrismaClient, modelNames: readonly DB_MODEL[]) {
-	const delegates = modelNames.map<{ deleteMany: () => Promise<unknown> }>(
-		(modelName) => prisma[modelName],
-	)
-
-	for (const delegate of delegates) {
-		await delegate.deleteMany()
+export async function cleanTables(
```
kibertoad, line 8: DbCleaner isn't currently exported from prisma-utils.
Considering that we no longer rely on Prisma models for truncation, we probably can export this as a generic reusable DB cleaner.
If we want to keep it type-safe, we probably can wrap it into a class, which would accept a list of supported modules as a constructor param.

packages/app/prisma-utils/test/DbCleaner.ts
kibertoad: yup, that would be great

packages/app/prisma-utils/test/DbCleaner.ts
```diff
@@ -1,16 +1,17 @@
...
-		(modelName) => prisma[modelName],
-	)
+	constructor(schema: string, dbModel: DbModel) {
+		this.schema = schema
+		this.tableNames = Object.values(dbModel)
+	}
 
-	for (const delegate of delegates) {
-		await delegate.deleteMany()
+	async cleanTables(prisma: PrismaClient) {
```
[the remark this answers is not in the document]
Dosexe: Yeah, overoptimized the code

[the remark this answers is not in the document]
Dosexe: Now it look to me, that we can simply expose one function without a class:) Unless we want to enclose the schema

PR #152, opened 2024-05-24
Crdb transaction retry

CarlosGamero:
## Changes

We have the following error:

```
Transaction failed with unrecoverable error: {"name":"PrismaClientKnownRequestError","code":"P2010","clientVersion":"5.13.0","meta":{"code":"40001","message":"restart transaction: TransactionRetryWithProtoRefreshError: WriteTooOldError: write for key ...
```

Following [crdb doc](https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference#:~:text=To%20indicate%20that%20a%20transaction,the%20string%20%22restart%20transaction%22%20.)
> All transaction retry errors use the SQLSTATE error code 40001, and emit error messages with the string [restart transaction](https://www.cockroachlabs.com/docs/v24.1/common-errors#restart-transaction). Further, each error includes a [specific error code](https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference#transaction-retry-error-reference) to assist with targeted troubleshooting.All transaction retry errors use the SQLSTATE error code 40001, and emit error messages with the string [restart transaction](https://www.cockroachlabs.com/docs/v24.1/common-errors#restart-transaction). Further, each error includes a [specific error code](https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference#transaction-retry-error-reference) to assist with targeted troubleshooting.

## Checklist

### DIFF
--- a/packages/app/prisma-utils/src/cockroachdbError.ts
+++ b/packages/app/prisma-utils/src/cockroachdbError.ts
@@ -1,0 +1,26 @@
+import type { PrismaClientKnownRequestError } from '@prisma/client/runtime/library'
+
+/**
+ * https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference#:~:text=To%20indicate%20that%20a%20transaction,the%20string%20%22restart%20transaction%22%20.
+ *
+ * All transaction retry errors use the SQLSTATE error code 40001, and emit error messages with the string restart transaction. Further, each error includes a specific error code to assist with targeted troubleshooting
+ */
+const COCKROACHDB_RETRY_TRANSACTION_CODE = '40001'
+const COCKROACHDB_RETRY_TRANSACTION_MESSAGE = 'restart transaction'
+
+/**
+ * Check if the error is a CockroachDB transaction retry error
+ *
+ * @param error
+ * TODO: add tests
+ */
+export const isCockroachDBRetryTransaction = (error: PrismaClientKnownRequestError): boolean => {
+	const meta = error.meta
+	if (!meta) return false
+
+	return (
+		meta.code === COCKROACHDB_RETRY_TRANSACTION_CODE &&
+		typeof meta.message === 'string' &&
+		meta.message.includes(COCKROACHDB_RETRY_TRANSACTION_MESSAGE)
+	)
+}
--- a/packages/app/prisma-utils/src/cockroachdbError.ts
+++ b/packages/app/prisma-utils/src/cockroachdbError.ts
@@ -12,7 +12,6 @@ const COCKROACHDB_RETRY_TRANSACTION_MESSAGE = 'restart transaction'
  * Check if the error is a CockroachDB transaction retry error
  *
  * @param error
- * TODO: add tests
  */
 export const isCockroachDBRetryTransaction = (error: PrismaClientKnownRequestError): boolean => {
 	const meta = error.meta
--- a/packages/app/prisma-utils/src/cockroachdbError.ts
+++ b/packages/app/prisma-utils/src/cockroachdbError.ts
@@ -3,10 +3,9 @@ import type { PrismaClientKnownRequestError } from '@prisma/client/runtime/libra
 /**
  * https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference#:~:text=To%20indicate%20that%20a%20transaction,the%20string%20%22restart%20transaction%22%20.
  *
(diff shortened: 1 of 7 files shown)


kjamrog (approved):

packages/app/prisma-utils/src/prismaTransaction.ts
```diff
@@ -45,11 +46,7 @@ export const prismaTransaction = (async <T, P extends PrismaClient>(
 	while (retries < options.retriesAllowed) {
 		result = await executeTransactionTry(prisma, arg, options)
 
-		if (
-			result.result ||
-			!isPrismaClientKnownRequestError(result.error) ||
-			result.error.code !== PRISMA_SERIALIZATION_ERROR
-		) {
+		if (result.result || !isRetryAllowed(result)) {
```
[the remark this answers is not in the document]
CarlosGamero: We can implement it, yes, but delay here shouldn’t be high so base default delay should be low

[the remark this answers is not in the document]
CarlosGamero: Not sure how should be that increment, exponential backoff sound like too much 🤔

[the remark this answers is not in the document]
kibertoad: what delay would you use?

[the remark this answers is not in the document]
CarlosGamero: I am not sure tbh, but I would start using the same approach CockroachDB is using in their [example](https://www.cockroachlabs.com/docs/stable/transaction-retry-error-example), exponential backoff with 100ms as default value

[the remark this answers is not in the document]
kibertoad: sounds good!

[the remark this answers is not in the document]
CarlosGamero: Nice! I will take independently to not delay this PR 🙇

packages/app/prisma-utils/src/cockroachdbError.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param error
+ */
+export const isCockroachDBRetryTransaction = (error: PrismaClientKnownRequestError): boolean => {
+	const meta = error.meta
+	if (!meta) return false
+
+	return (
+		meta.code === COCKROACHDB_RETRY_TRANSACTION_CODE &&
+		typeof meta.message === 'string' &&
+		meta.message.includes(COCKROACHDB_RETRY_TRANSACTION_MESSAGE)
```
kibertoad, line 23: is this part necessary? coupling to specific messages is not very reliable. we can't rely on code alone?

CarlosGamero: We can yes, but I wanted to be as explicit as possible, this is a CockroachDB-specific error and we may want to use this library with other DBs and not sure if we can have the same code with different meanings

About that coupling to specific messages is not very reliable, I completely agree, but on doc, they explicitly say that the message should contain that part so I would say it is fine.

But I have no strong opinion, if you prefer to remove that check I am more than happy to do it

kibertoad: fair point that this is a cockroach-specific error. can we do a more reliable check? does prisma client expose the used driver name?

CarlosGamero: That was the first idea, but I couldn't find it exposed 😓

kibertoad: Main thing that concerns me is that we don't have a real integration test throwing this error, so if CockroachDB ever decides to change the message, no tests will be failing.
what do you think about passing a db driver enum as a part of PrismaTransactionOptions?

CarlosGamero: Fair point and good idea! yeah, will include the DB driver as part of the options.

I think I will make it mandatory which means a major release, so in the future when we add more drivers we don't need to migrate :D

kibertoad: sounds good! we can also default to CRDB to retain backwards compatibility :D

CarlosGamero: yeah, finally decided to use it as default, and later when we need to add more Db drivers we can see if make sense to make it mandatory

kibertoad (approved):

PR #155, opened 2024-05-30
Prisma transaction retry delay

CarlosGamero:
## Changes

Prisma transaction incorporates a exponential backoff retry with 100ms as default base delay

## Checklist

### DIFF
--- a/packages/app/prisma-utils/src/prismaTransaction.ts
+++ b/packages/app/prisma-utils/src/prismaTransaction.ts
@@ -22,7 +22,11 @@ const DEFAULT_OPTIONS = {
 	retriesAllowed: 3,
 	DbDriver: 'CockroachDb',
 	baseRetryDelayMs: 100,
-} satisfies Pick<PrismaTransactionOptions, 'retriesAllowed' | 'DbDriver' | 'baseRetryDelayMs'>
+	maxRetryDelayMs: 30000, // 30s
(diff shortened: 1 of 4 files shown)


packages/app/prisma-utils/src/prismaTransaction.ts
```diff
@@ -72,6 +84,11 @@ const executeTransactionTry = async <T, P extends PrismaClient>(
 	}
 }
 
+const calculateRetryDelay = (retries: number, baseRetryDelayMs: number): number => {
+	// exponential backoff -> 2^(retry-1) * baseRetryDelayMs
+	return Math.pow(2, retries - 1) * baseRetryDelayMs
```
kibertoad, line 89: maybe we should support optional cap for it too? I reckon that transaction that is executed hours later is not particularly a good idea.

CarlosGamero: Added `maxRetryDelayMs`, it is optional but with a default value of the 30s. let me know if you think the naming or default values are not correct 🙏 

Also, check with Anton and he is fine as long as the retry mechanism follows this rules:
1. Don’t retry forever
2. Don’t create heavy load with retries
3. Skip retries after a reasonable number of tries/time

So I think all good 🎉

kibertoad (approved):
great job!

PR #156, opened 2024-05-30
Bugsnag report error details

CarlosGamero:
## Changes

Adding internal error and public error details to Bugsnag context by default so we can avoid doing it manually

## Checklist


CarlosGamero:
@kibertoad finally decided to do it here because, if I am not wrong, it will solve the issue for all the services/libraries where we are using Bugsnag, please correct me if it is not right 🙏

kibertoad (approved):

PR #158, opened 2024-06-02
Fix error-utils exports

kibertoad:
## Changes

vite-based build configuration wasn't working well for a node-only library, and a big bunch of dependencies from node_modules were included in the generated dist bundle

this probably can be fixed via vite configuration, but as a quick fix I've replaced the library config with what we use for other node-only libraries.

## Checklist

### DIFF
--- a/packages/app/error-utils/package.json
+++ b/packages/app/error-utils/package.json
@@ -1,29 +1,22 @@
 {
 	"name": "@lokalise/error-utils",
-	"version": "1.2.0",
-	"type": "module",
+	"version": "1.2.1",
 	"license": "Apache-2.0",
 	"files": [
-		"dist"
+		"dist/**",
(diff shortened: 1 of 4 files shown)


packages/app/error-utils/package.json
```diff
@@ -1,29 +1,22 @@
 {
 	"name": "@lokalise/error-utils",
-	"version": "1.2.0",
-	"type": "module",
+	"version": "1.2.1",
```
CarlosGamero, line 3: I thin version will be updated automatically on release action

CarlosGamero (approved):

packages/app/error-utils/package.json
kibertoad: true. fixed!

PR #196, opened 2024-06-26
poc -> transaction auto-incremental timeout

CarlosGamero:
**THIS IS POC** to start the discussion about this proposed feature, if we decide to implement this we will create a new branch adding the proper code with the discussed solution and tests

### DIFF
--- a/packages/app/prisma-utils/src/prismaTransaction.ts
+++ b/packages/app/prisma-utils/src/prismaTransaction.ts
@@ -18,14 +18,16 @@ type PrismaTransactionReturnType<T> = Either<
 	T | runtime.Types.Utils.UnwrapTuple<Prisma.PrismaPromise<unknown>[]>
 >
 
+
 const DEFAULT_OPTIONS = {
 	retriesAllowed: 3,
 	DbDriver: 'CockroachDb',
 	baseRetryDelayMs: 100,
 	maxRetryDelayMs: 30000, // 30s
+	timeout: 5000, // 5s
 } satisfies Pick<
 	PrismaTransactionOptions,
-	'retriesAllowed' | 'DbDriver' | 'baseRetryDelayMs' | 'maxRetryDelayMs'
+	'retriesAllowed' | 'DbDriver' | 'baseRetryDelayMs' | 'maxRetryDelayMs' | 'timeout'
 >
 
 /**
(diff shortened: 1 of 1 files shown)


packages/app/prisma-utils/src/prismaTransaction.ts
```diff
@@ -58,9 +60,12 @@ export const prismaTransaction = (async <T, P extends PrismaClient>(
...
 		result = await executeTransactionTry(prisma, arg, options)
-		if (result.result || !isRetryAllowed(result, optionsWithDefaults.DbDriver)) {
-			break
-		}
+		if (result.result) break
+
+		const retryAllowed = isRetryAllowed(result, optionsWithDefaults.DbDriver)
+		if (!retryAllowed) break
+
+		if (retryAllowed === 'increase-timeout') optionsWithDefaults.timeout *= 2 // TODO: implement smart timeout increase
```
CarlosGamero, line 68: This incremental behavior is only an example, if we agree this is a good feature we should define how the increase should look like

kibertoad, line 68: this sounds reasonable, but maybe we should cap it to some max value?

CarlosGamero: Sounds great! we can add a new parameter to define that max value with a default of 30s for example.

About how to increase the timeout, the `*=2` approach sounds good?

packages/app/prisma-utils/src/prismaTransaction.ts
```diff
@@ -104,11 +109,13 @@ const calculateRetryDelay = (
 	return Math.min(expDelay, maxDelayMs)
 }
 
-const isRetryAllowed = <T>(result: PrismaTransactionReturnType<T>, dbDriver: DbDriver): boolean => {
+type IsRetryAllowedResult = boolean | 'increase-timeout'
```
kjamrog, line 112: 🟢 I think returning something like:
```
{
  retryAllowed: boolean;
  increaseTimeout: boolean;
}
```
would be a bit more intuitive. But it can be just my preference, so no strong opinion about it. 
WDYT?

CarlosGamero: That sounds even better! I will apply your suggestion to the final version :D thanks 🙏

CarlosGamero:
Will create a new PR with the proper implementation

PR #201, opened 2024-07-02
Transaction auto increase timeout

CarlosGamero:
## Changes

Feature to auto-increase timeout in case it is the error on the transaction

## Checklist

### DIFF
--- a/packages/app/prisma-utils/src/prismaError.ts
+++ b/packages/app/prisma-utils/src/prismaError.ts
@@ -17,6 +17,9 @@ export const isPrismaClientKnownRequestError = (
 	typeof error.code === 'string' &&
 	error.code.startsWith('P')
 
+export const isPrismaTransactionClosedError = (error: PrismaClientKnownRequestError): boolean =>
+	error.code === PRISMA_TRANSACTION_ERROR && error.message.includes('Transaction already closed')
+
 /*
  * Prisma error code P2025 indicates that the operation failed because it depends on one or more
  * records that were required but not found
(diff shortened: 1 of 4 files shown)


packages/app/prisma-utils/src/prismaError.ts
```diff
@@ -17,6 +17,9 @@ export const isPrismaClientKnownRequestError = (
 	typeof error.code === 'string' &&
 	error.code.startsWith('P')
 
+export const isPrismaTransactionClosedError = (error: PrismaClientKnownRequestError): boolean =>
+	error.code === PRISMA_TRANSACTION_ERROR && error.message.includes('Transaction already closed')
```
CarlosGamero, line 21: There is no other way to recognize a timeout issue, in all my tests we always receive the same message (`Transaction already closed`).

We can't rely only on `PRISMA_TRANSACTION_ERROR` as, in the documentation they mention that it can be used for other purposes

packages/app/prisma-utils/src/prismaTransaction.ts
```diff
@@ -104,11 +111,17 @@ const calculateRetryDelay = (
 	return Math.min(expDelay, maxDelayMs)
 }
 
-const isRetryAllowed = <T>(result: PrismaTransactionReturnType<T>, dbDriver: DbDriver): boolean => {
+type isRetryAllowedResult = boolean | 'increase-timeout'
+
```
CarlosGamero, line 115: @kjamrog I implemented your suggestion at the beginning, but including the object for this only reason sounds to me like too much, but if you feel it is really better let me know and I will change it

packages/app/prisma-utils/src/types.ts
```diff
@@ -21,8 +24,16 @@ export type PrismaTransactionOptions = {
 }
 
 // Prisma $transaction with array does not support maxWait and timeout options
-export type PrismaTransactionBasicOptions = Omit<PrismaTransactionOptions, 'maxWait' | 'timeout'>
+export type PrismaTransactionBasicOptions = Omit<
+	PrismaTransactionOptions,
+	'maxWait' | 'timeout' | 'maxTimeout'
+>
```
CarlosGamero, line 30: Prisma transaction with array doesn't have timeout option, so excluding `maxTimeout` too

packages/app/prisma-utils/src/prismaTransaction.ts
kjamrog: Let's leave it as it is then, no worries!

kjamrog (approved):
Very nice 🚀

kibertoad (approved):

PR #203, opened 2024-07-03
Add base Biome config

kibertoad:
## Changes

This provides a strawman proposal to kickstart the definition of our biome rules, any suggestions to adjust the rules would be most welcome.

## Checklist


packages/dev/biome-config/configs/biome-base.json
```diff
@@ -0,0 +1,147 @@
...
+			"performance": {
+				"noAccumulatingSpread": "error",
+				"noBarrelFile": "error",
+				"noDelete": "error",
+				"noReExportAll": "error"
+			},
+			"style": {
+				"noNonNullAssertion": "off",
+				"noDefaultExport": "error",
+				"noParameterAssign": "error",
```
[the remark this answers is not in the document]
kibertoad: yes. personally I prefer explicit rule setup, but I'm open to remove default rules from the list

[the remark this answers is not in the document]
ondrejsevcik: I'm fine with it. I was just asking to make sure this is intentional.

My personal preference is to keep the defaults out. That way, I can look at the config and see only the subset that we decided to change. I suppose as Biome adds new rules to the recommended config, it will also get applied to our setup (even though it won't be explicitly specified here).

packages/dev/biome-config/configs/biome-base.json
```diff
@@ -0,0 +1,147 @@
+{
```
ondrejsevcik, line 1: 🟡 If possible, let's use `.jsonc` [file extension](https://biomejs.dev/internals/language-support/#jsonc-support). It should allow us to drop any comments if necessary. It can be handy in case we need to explain settings in this config.

packages/dev/biome-config/configs/biome-base.json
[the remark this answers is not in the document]
kibertoad: I don't mind this approach.
is there a list of default rules I could use to remove them from here?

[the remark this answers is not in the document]
ondrejsevcik: Here is a list of [recommended rules](https://biomejs.dev/linter/rules/#recommended-rules).

packages/dev/biome-config/configs/biome-base.json
```diff
@@ -0,0 +1,147 @@
...
+		}
+	},
+	"javascript": {
+		"globals": ["vitest"],
+		"formatter": {
+			"enabled": true,
+			"arrowParentheses": "always",
+			"bracketSpacing": true,
+			"indentWidth": 2,
+			"indentStyle": "space",
```
ondrejsevcik, line 23: 🟡 This will be hell to migrate. We're using tab in all of our FE projects. Should we stick to [default tab configuration](https://biomejs.dev/reference/configuration/#formatterindentstyle)?

kibertoad: sure

ondrejsevcik: Sure we can. Is BE using spaces by default (at Lokalise)?

kibertoad: yup, with the exception of Autopilot

packages/dev/biome-config/configs/biome-base.json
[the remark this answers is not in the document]
kibertoad: removed all of the default configuration

packages/dev/biome-config/configs/biome-base.json
kibertoad: done

CarlosGamero (approved):

packages/dev/biome-config/configs/biome-base.json
CarlosGamero: Should we also create an override for AP(backend)? or do we want to migrate to space?

kibertoad: last time I ran a poll on this topic, 100% of AP engineers who participated selected "I don't care" option, so it doesn't look like we need to maintain two separate sets of rules, we can just reformat it once and apply it within open PRs as well

CarlosGamero: Sounds great to me! 🙏 thanks for sharing Igor 🙇

PR #205, opened 2024-07-04
Accept any string as a correlationId

kibertoad:
## Changes

We do not always control the correlationId of a request that initiates the chain, so we should be flexible in what we accept there.

## Checklist

### DIFF
--- a/packages/app/events-common/src/baseEventSchemas.ts
+++ b/packages/app/events-common/src/baseEventSchemas.ts
@@ -5,7 +5,7 @@ export const BASE_EVENT_SCHEMA = z.object({
 	id: z.string().uuid().describe('event unique identifier'),
 	type: z.literal<string>('<replace.me>').describe('event type name'),
 	timestamp: z.string().datetime().describe('iso 8601 datetime'),
-	source: z.string().nonempty().describe('source service of the event'),
+	source: z.string().min(1).describe('source service of the event'),
 	payload: z.optional(z.object({})).describe('event payload based on type'),
 	metadata: z
 		.object({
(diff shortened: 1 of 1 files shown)


packages/app/events-common/src/baseEventSchemas.ts
```diff
@@ -5,19 +5,16 @@ export const BASE_EVENT_SCHEMA = z.object({
...
+			schemaVersion: z.string().min(1).describe('base event schema version'),
+			originalApp: z.string().min(1).describe('app/service initiated workflow'),
 		})
 		.describe('event metadata'),
-	correlationId: z
-		.string()
-		.uuid()
-		.describe('unique identifier passed to all events in workflow chain'),
-	version: z.string().nonempty().describe('event payload version'),
+	correlationId: z.string().describe('unique identifier passed to all events in workflow chain'),
```
ondrejsevcik, line 16: Is this required or could it be optional?

jmontoyaa (approved):

packages/app/events-common/src/baseEventSchemas.ts
kibertoad: I would say it is mandatory - when you are firing an event, you are expected to set it, and when you are receiving it, it should be set in all cases. Otherwise it is easy to forget to set correlationId and we would lose quite a bit of observability.

This is not our case, btw. I've tested what happens when you get undefined, and error is different from what you can see in BugSnag:

```json
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
```

PR #212, opened 2024-07-09
Add isomorphic utils

kibertoad:
## Changes

Moved a bit of node-core that could also be used in browser

## Checklist


.github/workflows/release.package.yml
```diff
@@ -57,9 +57,9 @@ jobs:
 
       - name: Git push
         run: COMMIT_MSG=$(npm run --silent package-version --workspace=${{ inputs.working_directory }}) && git commit -am "${{ inputs.package_name }} $COMMIT_MSG" && git push origin main
-
+01
```
kibertoad, line 60: cat why you type on my keyboard

.github/workflows/release.package.yml
```diff
@@ -57,9 +57,9 @@ jobs:
...
       - name: Release Package
         env:
           NODE_AUTH_TOKEN: ${{ secrets.npm_token }}
-        run: npm publish
+        run: npm publish --access public
```
kibertoad, line 64: without this flag the very first publish will fail

.github/workflows/release.package.yml
kibertoad: @dariacm not even joking

dariacm: The link is not working, if it's a cat picture I urge you to fix it for my mental health

kibertoad: didn't work from mobile, worked from PC.

Google: 0, Microsoft: 1

dariacm: Maybe there was a reason for that `01`, I would ask her

dariacm (approved):

.github/workflows/release.package.yml
kibertoad: she just bit me instead of replying, so I assume yes

.github/workflows/release.package.yml
CarlosGamero: Good catch! thanks Igor 🙇

CarlosGamero:
@kibertoad Sorry for the late review 😅, PR looks good to me but I have a follow-up question, are we planning to move "node-core utils" to this package? I have no problem with it, asking just to understand the purpose of this package

.github/workflows/release.package.yml
kibertoad: @CarlosGamero tbf, that's on npm. it could have picked up the public access setting from package.json, but noooooo

kibertoad:
@CarlosGamero only a subset of node-core that is relevant both on BE and FE. This PR is based on @ondrejsevcik 's request to have a place where we could put utils that FE would also benefit from.

we should probably move node-core to this repo, though

CarlosGamero:
@kibertoad I think that there are plenty of util methods that can be shared between FE and BE tbh, but yeah, we don't need to move all of them at once, we can do it step by step :D

CarlosGamero:
Ah, about moving node-core here, yeah, it sounds good, so we don't need to have so many repos 😅

kibertoad:
@CarlosGamero some of them we should revise and deprecate, e. g. I think there is native groupBy in TS now

PR #215, opened 2024-07-11
chore: allow comments within json files

ondrejsevcik:


packages/dev/biome-config/configs/biome-base.jsonc
```diff
@@ -8,6 +8,9 @@
       "enabled": true,
       "lineWidth": 100,
       "indentStyle": "space"
+    },
+    "parser": {
+      "allowComments": true
```
ondrejsevcik, line 13: Most of the tools in JS ecosystem allow comments within json files. We also have quite a few json files with those.

CarlosGamero (approved):

kibertoad (approved):

PR #219, opened 2024-07-11
http client 204 support

CarlosGamero:
## Changes

adding support for `204`, same as we did for FE client

## Checklist

### DIFF
--- a/packages/app/backend-http-client/README.md
+++ b/packages/app/backend-http-client/README.md
@@ -43,6 +43,7 @@ All _send_ methods accept a type parameter and the following arguments:
   - `clientOptions`;
   - `responseSchema`, used both for inferring the response type of the call, and also (if `validateResponse` is `true`) for validating the response structure;
   - `validateResponse`;
+  - `isEmptyResponseExpected`, used to specify if a 204 response should be treated as an error or not
 
   The following options are applied by default:
 
@@ -57,6 +58,7 @@ All _send_ methods accept a type parameter and the following arguments:
       retryOnTimeout: false,
   }
   ```
+  For `sendDelete()` `isEmptyResponseExpected` by default is set to `true`, for all other methods it is `false`.
 
 Additionally, `sendPost()`, `sendPut()`, `sendPutBinary()`, and `sendPatch()` also accept a `body` parameter.
 
--- a/packages/app/backend-http-client/README.md
+++ b/packages/app/backend-http-client/README.md
@@ -43,7 +43,7 @@ All _send_ methods accept a type parameter and the following arguments:
   - `clientOptions`;
   - `responseSchema`, used both for inferring the response type of the call, and also (if `validateResponse` is `true`) for validating the response structure;
   - `validateResponse`;
-  - `isEmptyResponseExpected`, used to specify if a 204 response should be treated as an error or not
+  - `isEmptyResponseExpected`, used to specify if a 204 response should be treated as an error or not. when `true` the response body type is adjusted to include potential `null`
 
   The following options are applied by default:
 
--- a/packages/app/backend-http-client/index.ts
+++ b/packages/app/backend-http-client/index.ts
@@ -2,9 +2,9 @@ export type {
   RequestOptions,
   HttpRequestContext,
   RequestResultDefinitiveEither,
-} from "./src/client/types";
+} from './src/client/types'
 
-export { JSON_HEADERS } from "./src/client/constants";
+export { JSON_HEADERS } from './src/client/constants'
 
 export {
   sendPut,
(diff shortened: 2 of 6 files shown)


packages/app/backend-http-client/index.ts
```diff
@@ -8,14 +16,9 @@ export {
...
   httpClient,
   buildClient,
-  type RequestOptions,
-  type Response,
-  type HttpRequestContext,
-  type ResponseSchema,
-  TEST_OPTIONS,
-  JSON_HEADERS,
-  NO_CONTENT_RESPONSE_SCHEMA,
-  UNKNOWN_RESPONSE_SCHEMA,
```
CarlosGamero, line 18: Removed:
- `Response` -> looks like it was not being used, as we are using RequestResultt from `undici-retry`
- `ResponseSchema` -> using `ZodSchema` instead
- `TEST_OPTIONS` -> was only used on internal tests
- `NO_CONTENT_RESPONSE_SCHEMA` -> was only used on internal tests

Please let me know if you disagree with one of this changes 🙏

packages/app/backend-http-client/src/client/constants.ts
```diff
@@ -0,0 +1,22 @@
+import type { MayOmit } from '@lokalise/node-core'
+import type { Client } from 'undici'
+
+import type { RequestOptions } from './types'
```
CarlosGamero, line 4: `httpClient.ts` is already quite big and difficult to read, so extracting constants here

packages/app/backend-http-client/src/client/types.ts
```diff
@@ -0,0 +1,44 @@
+import type { DefiniteEither } from '@lokalise/node-core'
+import type { Client } from 'undici'
+import type { RequestResult, RetryConfig } from 'undici-retry'
+import type { ZodSchema } from 'zod'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
CarlosGamero: Same for types

packages/app/backend-http-client/index.ts
kibertoad: TEST_OPTIONS and NO_CONTENT_RESPONSE_SCHEMA are exposed so that other services could use them for their tests as well, we actually use at least TEST_OPTIONS in e. g. node-service-template

CarlosGamero: Ok, will put `TEST_OPTIONS` back thanks for the clarification 🙏 

About `NO_CONTENT_RESPONSE_SCHEMA`, does it make sense to have it? `204` will now return `null` and there is no validation

kibertoad: but http client still requires you to pass some schema. and we can't make that requirement conditional, because in some cases we are only going to get 204 (e. g. DELETE), but in some cases we may get either 200 or 204 (GET)

CarlosGamero: You are right, but I think it is something that can be done by the user, but I have no strong opinion/problem with having it, so adding it back

kibertoad: it can be, but it's slightly more convenient not to have to reinvent it on every service :D

packages/app/backend-http-client/src/client/constants.ts
kibertoad: great idea!

packages/app/backend-http-client/index.ts
CarlosGamero: Yeah, sounds like a good reason, thanks for mentioning it Igor 🙏, both are back already

kibertoad: curious if removing Response will break something or not. let's try removing it and see if we miss it :D

kibertoad (approved):
great job!

kibertoad:
can you document new config option, btw? prolly we can copy the FE one

CarlosGamero:
Ups, yes, absolutely, I forgot about it, will document in a bit, thank you for the reminder 🙏

packages/app/backend-http-client/README.md
```diff
@@ -43,6 +43,7 @@ All _send_ methods accept a type parameter and the following arguments:
   - `clientOptions`;
   - `responseSchema`, used both for inferring the response type of the call, and also (if `validateResponse` is `true`) for validating the response structure;
   - `validateResponse`;
+  - `isEmptyResponseExpected`, used to specify if a 204 response should be treated as an error or not
```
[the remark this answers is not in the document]
CarlosGamero: Good catch! thank youuu 🙇

PR #224, opened 2024-07-12
fix: allow test folder to be nested

ondrejsevcik:
This change allows for `test` folder to be nested.

In FE, we have usually test folder nested within `src/test/whatever-file.ts`.

### DIFF
--- a/packages/app/backend-http-client/package.json
+++ b/packages/app/backend-http-client/package.json
@@ -11,11 +11,7 @@
     "url": "git://github.com/lokalise/backend-http-client.git"
   },
   "license": "Apache-2.0",
-  "files": [
-    "dist/**",
-    "LICENSE.md",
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

packages/app/backend-http-client/package.json
```diff
@@ -11,11 +11,7 @@
     "url": "git://github.com/lokalise/backend-http-client.git"
   },
   "license": "Apache-2.0",
-  "files": [
-    "dist/**",
-    "LICENSE.md",
-    "README.md"
-  ],
+  "files": ["dist/**", "LICENSE.md", "README.md"],
```
ondrejsevcik, line 14: I had to reformat this file. The CI Build was complaining about format of this file.

PR #225, opened 2024-07-12
MET-160 Support for Redis keyPrefix

dariacm:
## Changes

~~Sorry folks, this is going to be tough to review 😅~~

Platform suggested that we use Redis `keyPrefix` over `db` for isolation in Maestro service, so we tried to set up a Redis connection in our service but soon found out that BullMQ does not support connections with `keyPrefix` and have their own implementation for it (see [here](https://github.com/taskforcesh/bullmq/blob/aecac66b891045397d915fbc2cc1c3c92c5ae764/src/classes/redis-connection.ts#L87)). Therefore the goal of this PR is to allow passing a Redis configuration instead of a Redis connection to BullMQ and use `keyPrefix` if set only in BullMQ implementation, not Redis. While doing so, we also figured out that it's just easier to only support passing the configuration instead of supporting both config and connection and have to branch in multiple places based on the setup. In the end not much changes since BullMQ does not reuse connections but clones the configuration from it and opens a new one.

There will be some work to be done to update the library in Autopilot but the functionality should remain intact 🙇‍♀️ 

~~To make this easier for us but harder for reviewers, we also switched to Biome in the meantime.~~

## Checklist


kibertoad (approved):

CarlosGamero:
@dariacm should we maybe create and merge a PR only with the lint fixes (with the `skip-release` tag)?

dariacm:
@CarlosGamero you mean the Biome changes? Or something else? 🙇‍♀️

dariacm:
@CarlosGamero on it to facilitate review 🙏

packages/app/background-jobs-common/test/dependencyMocks.ts
```diff
@@ -33,24 +28,20 @@ export class DependencyMocks {
       transactionObservabilityManager: {
         start: vi.fn(),
         stop: vi.fn(),
-        // biome-ignore lint/suspicious/noExplicitAny: it's okay
```
dariacm, line 36: I honestly have no idea why it was complaining in https://github.com/lokalise/shared-ts-libs/pull/226 but not here

packages/app/background-jobs-common/package.json
```diff
@@ -18,7 +18,7 @@
   "types": "./dist/index.d.ts",
   "scripts": {
     "build": "rimraf dist && tsc",
-    "clean": "rimraf dist",
+    "clean": "rimraf dist ",
```
CarlosGamero, line 21: 🟢 Nit: unneeded space

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -121,20 +124,23 @@ export abstract class AbstractBackgroundJobProcessor<
...
     this.transactionObservabilityManager = dependencies.transactionObservabilityManager
     this.logger = dependencies.logger
     this.errorReporter = dependencies.errorReporter
     this._spy = config.isTest ? new BackgroundJobProcessorSpy() : undefined
+    this.runningPromises = []
   }
 
-  public static async getActiveQueueIds(redis: Redis): Promise<string[]> {
-    await redis.zremrangebyscore(
+  public static async getActiveQueueIds(redisConfig: RedisConfig): Promise<string[]> {
```
CarlosGamero, line 135: 🟡 This method is used from bull-board to retrieve the list of queues for the given connection, this method is called with an interval to refresh the on UI, wondering if we can still pass the Redis object instead of the config to not have to create a new connection with each call

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -85,11 +87,12 @@ export abstract class AbstractBackgroundJobProcessor<
 > {
   protected readonly logger: CommonLogger
 
-  private readonly redis: Redis
+  private redis?: Redis
```
CarlosGamero, line 90: 🟡 Please correct me if I am wrong, but I think we are not using this connection right? could we get rid of it?

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: Just as an idea, maybe instead removing it we can use it as `redisWithoutPrefix` we have in `start` method

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -121,20 +124,23 @@ export abstract class AbstractBackgroundJobProcessor<
   ) {
     this.config = config
```
CarlosGamero, line 125: 🟢 I think we always need to sanitize `redisConfig` (excluding `this.redis`, please check my prev comment) so maybe we can save it already sanitized to avoid having it in several places

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -455,13 +482,21 @@ export abstract class AbstractBackgroundJobProcessor<
...
+    const updateDataPromise = job.updateData({ metadata: job.data.metadata })
+    const clearLogsPromise = job.clearLogs()
+
+    const updateDataPromiseIndex = this.runningPromises.push(updateDataPromise) - 1
+    const clearLogsPromiseIndex = this.runningPromises.push(clearLogsPromise) - 1
+
+    await Promise.all([updateDataPromise, clearLogsPromise]).then(() => {
+      this.runningPromises.splice(updateDataPromiseIndex, 1)
+      this.runningPromises.splice(clearLogsPromiseIndex, 1)
+    })
```
CarlosGamero, line 499: Good idea! 🚀

CarlosGamero (commented):
Good job 🚀! thanks for doing it Daria 🙏, I left some commets but overall LGTM :D

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
dariacm: I added the possibility to pass either a connection or the configuration to fit both use cases 🙏

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
dariacm: You're right, good catch! ❤️

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
dariacm: Copyright @kibertoad 😉

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -455,14 +471,26 @@ export abstract class AbstractBackgroundJobProcessor<
    * @param job
    * @protected
    */
-  protected async purgeJobData(job: JobType): Promise<void> {
+  async purgeJobData(job: JobType): Promise<void> {
```
CarlosGamero, line 474: Are we exposing this method intentionally?

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
kibertoad: it doesn't catch all cases, though, I still want to investigate if it's possible to somehow force Redis connection to drain all ongoing operations, then we wouldn't need to manage it ourselves. but that'd be outside of this PR

PR #233, opened 2024-07-15
Adding biome.package.json

CarlosGamero:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/dev/biome-config/configs/biome-package.jsonc
+++ b/packages/dev/biome-config/configs/biome-package.jsonc
@@ -1,0 +1,11 @@
+{
+  "$schema": "../../../node_modules/@biomejs/biome/configuration_schema.json",
+  "extends": ["./biome-base.jsonc"],
+  "linter": {
+    "rules": {
+      "performance": {
+        "noBarrelFile": "off"
+      }
+    }
+  }
+}
--- a/packages/app/background-jobs-common/biome.json
+++ b/packages/app/background-jobs-common/biome.json
@@ -1,0 +1,7 @@
+{
+  "$schema": "../../../node_modules/@biomejs/biome/configuration_schema.json",
+  "extends": [
+    "../../../node_modules/@lokalise/biome-config/configs/biome-base.jsonc",
+    "../../../node_modules/@lokalise/biome-config/configs/biome-package.jsonc"
+  ]
+}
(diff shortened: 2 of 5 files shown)


packages/dev/biome-config/configs/biome-package.jsonc
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../../../node_modules/@biomejs/biome/configuration_schema.json",
+  "extends": ["./biome-base.jsonc"],
```
kibertoad, line 3: unfortunately, extending from the same package is not supported by biome, it needs to be combined on a consumer side

packages/dev/biome-config/configs/biome-package.jsonc
```diff
@@ -0,0 +1,11 @@
+{
```
kibertoad, line 1: can you mention new preset in the documentation?

packages/dev/biome-config/configs/biome-package.jsonc
CarlosGamero: I was experiencing it right now 😓 then I guess that I should remove this remove on on package do 
```
extends: [base, package]
```
right?

packages/dev/biome-config/configs/biome-package.jsonc
CarlosGamero: On it, thank you!

CarlosGamero: btw, is the following correct:
```Backend:

```json
{
	"$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
	"extends": ["./node_modules/@lokalise/biome-config/configs/biome-base.jsonc", "./node_modules/@lokalise/biome-config/configs/biome-esm.jsonc"]
}
```

I think we are not using esm config, right? asking just to understand it 🙏

kibertoad: newer services (Billing, Workflows) use ESM, Autopilot doesn't (benefits of a switch do not currently justify the effort needed)

packages/dev/biome-config/configs/biome-package.jsonc
kibertoad: yup!

kibertoad (approved):

CarlosGamero:
Background-jobs-common had lint issues, they are not important and I don't think we need to release a new version of it so marking PR as `skip-release` and I will release new `biome-config` version manually

PR #237, opened 2024-07-17
feat: add uniq utility

ondrejsevcik:


packages/app/universal-ts-utils/src/array/uniq.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, expect, it } from 'vitest'
+import { uniq } from './uniq'
+
+describe('uniq', () => {
+	it('returns a new array of mixed primitive value without duplicates', () => {
```
kibertoad, line 5: would be nice to test with: nulls, undefineds, NaN, objects

ondrejsevcik: Good point. I added a few more values into the inputs. Let me know if there are some others you would test for.

packages/app/universal-ts-utils/src/array/uniq.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, it } from 'vitest'
+import { uniq } from './uniq'
+
+describe('uniq', () => {
+  it('returns a new array of mixed primitive value without duplicates', () => {
+    const objectA = {}
+    const values = [1, 'a', Number.NaN, true, false, objectA, null, undefined]
+    const duplicateValues = [...values, ...values]
```
kibertoad, line 8: note that this way of creating duplicates ensure that they are equal not just by value but also by reference.

in order to make behaviour more explicit, it may be helpful to have another test, where duplicates have same value but different reference. I think that will cause some of deduplication fail, which is fine, but something that users of the method should be aware of.

kibertoad (approved):

packages/app/universal-ts-utils/src/array/uniq.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * A method to remove duplicates in an array of primitive values
+ */
+export const uniq = <T>(arr: T[]): T[] => Array.from(new Set(arr))
```
CarlosGamero, line 4: 🟢 Nit: No strong opinion but I would call it `unique` or `removeDuplicates` to make it clear

packages/app/universal-ts-utils/src/array/uniq.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    expect(uniq(duplicateValues)).toEqual([
+      1,
+      'a',
+      Number.NaN,
+      true,
+      false,
+      null,
+      undefined,
+      objectA,
+      objectB,
```
CarlosGamero, line 38: 🟡 Is this intended to work with objects? for different (in memory) objects but with the same content this method won't work, wondering if it can be confusing. 
If we are fine with it, all good, if not I would say we can try to specify which types are accepted by the method

packages/app/universal-ts-utils/src/array/uniq.spec.ts
CarlosGamero: Is this method intended to be user with objects? as Igor mentioned objects are compared by reference and wondering if this can be confusing.

If we don't want to use this method with objects maybe we can restrict the accepted types to avoid misunderstanding

packages/app/universal-ts-utils/src/array/uniq.ts
ondrejsevcik: Good suggestion. I was living under the impression that `uniq` is an English word. Turns out it's a linux command 😅 so `unique` it is.

packages/app/universal-ts-utils/src/array/uniq.spec.ts
ondrejsevcik: I added comment to clarify that the uniqueness is compared based on reference check.

packages/app/universal-ts-utils/src/array/unique.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * A method to remove duplicates from an array of primitive values.
+ * Elements are compared by reference using a Set.
```
kibertoad, line 3: is that true? isn't it done using native equality check, which for objects relies on a reference check, but for primitives does not?

ondrejsevcik: It's actually using SameValueZero algorithm. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#value_equality

kibertoad (approved):

PR #239, opened 2024-07-18
prisma-utils -> transaction isolation level type

CarlosGamero:
## Changes

Adding supported isolation level type so there is no space for typos when using `prismaTransaction` 

## Checklist


packages/app/prisma-utils/src/isolation_level/isolationLevel.ts
```diff
@@ -0,0 +1,8 @@
+type IsolationLevel =
+  | 'Serializable'
+  | 'ReadCommitted'
+  | 'ReadUncommitted'
+  | 'RepeatableRead'
+  | 'Snapshot'
+
+export type CockroachDbIsolationLevel = Extract<IsolationLevel, 'ReadCommitted' | 'Serializable'>
```
CarlosGamero, line 8: Prisma-generated code has its enum for this, so maybe we can avoid to expose this

packages/app/prisma-utils/src/types.ts
```diff
@@ -20,7 +21,11 @@ export type PrismaTransactionOptions = {
   // Prisma $transaction options
   maxWait?: number
   timeout?: number
-  isolationLevel?: string
+  /*
+    For not library only supports CockroachDB, when we add support for other databases we need to update this to
```
kibertoad, line 25:
```suggestion
    For now library only supports CockroachDB, when we add support for other databases we need to update this to
```

kibertoad (approved):

PR #249, opened 2024-07-25
Bugfix/bg job multiple starts

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
@@ -93,6 +93,9 @@ export abstract class AbstractBackgroundJobProcessor<
   private readonly _spy?: BackgroundJobProcessorSpy<JobPayload, JobReturn>
   private readonly runningPromises: Promise<unknown>[]
 
+  private isStarted = false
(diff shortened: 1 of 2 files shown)


packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -154,8 +157,24 @@ export abstract class AbstractBackgroundJobProcessor<
   }
 
   public async start(): Promise<void> {
-    if (this.queue) return // Skip if processor is already started
+    if (this.isStarted) return // if it is already started -> skip
+    if (this.startPromise) return await this.startPromise // if it is already starting -> wait (avoid concurrent start)
```
kibertoad, line 161: we can slightly reduce amount of code if we invert the condition, and do `this.startPromise = this.internalInit()` only if promise is not set, but then proceed with `await this.startPromise` either way

kibertoad (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: Mmm, let me apply your suggestion :D

PR #255, opened 2024-07-29
bg common refactor

CarlosGamero:
## Changes

`AbstractBackgroundJobProcessor` is becoming huge, within this PR we will try to simplify it to improve maintenance

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/constants.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/constants.ts
@@ -1,3 +1,7 @@
+import type { JobsOptions, WorkerOptions } from 'bullmq'
+
+import { daysToSeconds } from './utils'
+
 /**
  * How many days we retain completed jobs
  */
@@ -12,3 +16,29 @@ export const RETENTION_FAILED_JOBS_IN_DAYS = 7
  * How many days we retain queue ids
  */
 export const RETENTION_QUEUE_IDS_IN_DAYS = 14
+
+/**
+ * Default config
+ *    - Retry config: 3 retries with 30s of total amount of wait time between retries using
+ *            exponential strategy https://docs.bullmq.io/guide/retrying-failing-jobs#built-in-backoff-strategies
+ *    - Job retention: 50 last completed jobs, 7 days for failed jobs
+ */
+export const DEFAULT_JOB_CONFIG: JobsOptions = {
+  attempts: 3,
+  backoff: {
+    type: 'exponential',
+    delay: 5000,
+  },
+  removeOnComplete: { count: RETENTION_COMPLETED_JOBS_IN_AMOUNT },
+  removeOnFail: {
+    age: daysToSeconds(RETENTION_FAILED_JOBS_IN_DAYS),
+  },
+}
+
+export const QUEUE_IDS_KEY = 'background-jobs-common:background-job:queues'
+
+export const DEFAULT_WORKER_OPTIONS = {
+  concurrency: 10,
+  maxStalledCount: 3, // same as default attempts by default
+  ttl: 60,
+} as const satisfies Omit<WorkerOptions, 'connection'> & { ttl: number }
--- a/packages/app/background-jobs-common/src/background-job-processor/index.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/index.ts
@@ -1,6 +1,4 @@
 export * from './types'
-export * from './utils'
-export * from './constants'
 export * from './factories/CommonBullmqFactory'
 export * from './factories/AbstractBullmqFactory'
 export * from './logger/BackgroundJobProcessorLogger'
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
@@ -13,18 +13,19 @@ import pino from 'pino'
 import { merge } from 'ts-deepmerge'
 
 import {
-  RETENTION_COMPLETED_JOBS_IN_AMOUNT,
-  RETENTION_FAILED_JOBS_IN_DAYS,
+  DEFAULT_JOB_CONFIG,
+  DEFAULT_WORKER_OPTIONS,
+  QUEUE_IDS_KEY,
   RETENTION_QUEUE_IDS_IN_DAYS,
 } from '../constants'
 import type { AbstractBullmqFactory } from '../factories/AbstractBullmqFactory'
 import { BackgroundJobProcessorLogger } from '../logger/BackgroundJobProcessorLogger'
 import { BackgroundJobProcessorSpy } from '../spy/BackgroundJobProcessorSpy'
 import type { BackgroundJobProcessorSpyInterface } from '../spy/types'
-import type { BaseJobPayload, BullmqProcessor, SafeJob, SafeQueue } from '../types'
+import type { BaseJobPayload, BullmqProcessor, RequestContext, SafeJob, SafeQueue } from '../types'
 import {
   daysToMilliseconds,
-  daysToSeconds,
+  isRedisClient,
   isStalledJobError,
   isUnrecoverableJobError,
   resolveJobId,
@@ -37,38 +38,7 @@ import type {
   JobsPaginatedResponse,
 } from './types'
 
-export interface RequestContext {
-  logger: CommonLogger
-  reqId: string
-}
-
-/**
- * Default config
- *    - Retry config: 3 retries with 30s of total amount of wait time between retries using
- *            exponential strategy https://docs.bullmq.io/guide/retrying-failing-jobs#built-in-backoff-strategies
- *    - Job retention: 50 last completed jobs, 7 days for failed jobs
- */
-const DEFAULT_JOB_CONFIG: JobsOptions = {
-  attempts: 3,
-  backoff: {
-    type: 'exponential',
-    delay: 5000,
-  },
-  removeOnComplete: { count: RETENTION_COMPLETED_JOBS_IN_AMOUNT },
-  removeOnFail: {
-    age: daysToSeconds(RETENTION_FAILED_JOBS_IN_DAYS),
-  },
-}
-
-const QUEUE_IDS_KEY = 'background-jobs-common:background-job:queues'
-
 const queueIdsSet = new Set<string>()
-
-const DEFAULT_WORKER_OPTIONS = {
-  concurrency: 10,
-  maxStalledCount: 3, // same as default attempts by default
-  ttl: 60,
-} as const satisfies Omit<WorkerOptions, 'connection'> & { ttl: number }
 
 export abstract class AbstractBackgroundJobProcessor<
   JobPayload extends BaseJobPayload,
@@ -495,8 +465,4 @@ export abstract class AbstractBackgroundJobProcessor<
   protected onFailed(_job: JobType, _error: Error, _requestContext: RequestContext): Promise<void> {
     return Promise.resolve()
   }
-}
-
-function isRedisClient(redis: RedisConfig | Redis): redis is Redis {
-  return 'options' in redis
 }
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
@@ -308,7 +308,10 @@ export abstract class AbstractBackgroundJobProcessor<
     try {
       const transactionName = `bg_job:${this.config.ownerName}:${this.config.queueId}`
       this.transactionObservabilityManager.start(transactionName, jobId)
-      this.logJobStarted(job, job.requestContext)
+      job.requestContext.logger.info(
+        { origin: this.constructor.name, jobId },
+        `Started job ${job.name}`,
+      )
 
       const result = await this.process(job, job.requestContext)
       isSuccess = true
@@ -316,7 +319,14 @@ export abstract class AbstractBackgroundJobProcessor<
       await job.updateProgress(100)
       return result
     } finally {
-      this.logJobFinished(job, job.requestContext, isSuccess)
+      job.requestContext.logger.info(
+        {
+          origin: this.constructor.name,
+          jobId,
+          isSuccess,
+        },
+        `Finished job ${job.name}`,
+      )
       this.transactionObservabilityManager.stop(jobId)
     }
   }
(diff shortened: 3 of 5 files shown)


packages/app/background-jobs-common/src/background-job-processor/index.ts
```diff
@@ -1,6 +1,4 @@
 export * from './types'
-export * from './utils'
-export * from './constants'
```
CarlosGamero, line 3: Internal constants and utilities should not be exposed

packages/app/background-jobs-common/src/background-job-processor/constants.ts
```diff
@@ -12,3 +16,29 @@ export const RETENTION_FAILED_JOBS_IN_DAYS = 7
...
  */
 export const RETENTION_QUEUE_IDS_IN_DAYS = 14
+
+/**
+ * Default config
+ *    - Retry config: 3 retries with 30s of total amount of wait time between retries using
+ *            exponential strategy https://docs.bullmq.io/guide/retrying-failing-jobs#built-in-backoff-strategies
+ *    - Job retention: 50 last completed jobs, 7 days for failed jobs
+ */
+export const DEFAULT_JOB_CONFIG: JobsOptions = {
```
CarlosGamero, line 26: Constants and types extracted to make `AbstractBackgroundJobProcessor` smaller

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -429,24 +409,6 @@ export abstract class AbstractBackgroundJobProcessor<
...
-
-  logJobFinished(job: JobType, requestContext: RequestContext, isSuccess: boolean): void {
-    requestContext.logger.info(
-      {
-        isSuccess,
-      },
-      `Finished job ${job.name}`,
-    )
-  }
-
```
CarlosGamero, line 449: Why do we need this to be public, even protected sounds weird to me, class should have very basic logging for start and finish, if most complex logging is required we have hooks to allow to implement what is needed

packages/app/background-jobs-common/src/background-job-processor/types.ts
```diff
@@ -1,15 +1,18 @@
...
 
 export type JobFinalState = FinishedStatus
 export type BaseJobPayload = { metadata: { correlationId: string } }
 
 // "scripts" field is incompatible between free and pro versions, and it's not particularly important
 // biome-ignore lint/suspicious/noExplicitAny: it's okay
-export type SafeJob<T = any, R = any, N extends string = string> = Omit<Job<T, R, N>, 'scripts'> & {
-  requestContext?: RequestContext
-}
+export type SafeJob<T = any, R = any, N extends string = string> = Omit<Job<T, R, N>, 'scripts'>
```
CarlosGamero, line 15: Removing `requestContext` from `SafeJob` so it is not visible for child classes, handling it in a different way but keeping `requestContext` on the job object to avoid recreation if not needed

packages/app/background-jobs-common/src/background-job-processor/utils.ts
```diff
@@ -16,3 +20,27 @@ export const isStalledJobError = (error: Error): boolean =>
 export const sanitizeRedisConfig = (config: RedisConfig): RedisConfig => {
```
kibertoad, line 20: wonder if we can improve naming a bit, "sanitization" usually refers to removing unsafe values, and in this case this is a slightly different kind of transformation.

packages/app/background-jobs-common/src/background-job-processor/utils.ts
```diff
@@ -16,3 +20,27 @@ export const isStalledJobError = (error: Error): boolean =>
...
+  options?: JobOptionsType,
+): JobOptionsType => {
+  const preparedOptions: JobOptionsType = {
+    jobId: generateMonotonicUuid(),
+    ...DEFAULT_JOB_CONFIG,
+    ...(options ?? ({} as JobOptionsType)),
+  }
+
+  if (isTest && typeof preparedOptions.backoff === 'object') {
+    preparedOptions.backoff.delay = 1
```
kibertoad, line 37: might be helpful to add comment, explaining that this is done because zero delay is handled weirly in BullMQ for concurrent job.

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
```diff
@@ -0,0 +1,112 @@
...
+import type {
+  BackgroundJobProcessorConfig,
+  BackgroundJobProcessorDependencies,
+} from '../processors/types'
+import type { BaseJobPayload, RequestContext, SafeJob } from '../types'
+import { resolveJobId, sanitizeRedisConfig } from '../utils'
+
+const queueIdsSet = new Set<string>()
+
+export class BackgroundJobProcessorMonitor<
```
kibertoad, line 18: can you add JSDoc, explaining the idea behind the monitor?

kibertoad, line 18: Great job on this abstraction! It's a really well chosen cohesive unit of related functionality, and separating it does make code easier to understand.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
```diff
@@ -527,9 +520,6 @@ describe('AbstractBackgroundJobProcessor', () => {
           errorJson: expect.stringContaining(onFailedCall.error.message),
         },
       })
-      expect(
```
kibertoad, line 530: do we no longer find this assertion valuable? how do we know childloggers work as expected?

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
kibertoad: Main reason is to be able to configure what fields and in what structure get logged, so e. g. you could add dumping of entire payload under debug level. Having it as `protected` is sufficient for that purpose, though.

You are right that hooks should be a valid replacement for that, but having an example (or a reference) somewhere in documentation about them existing and how to use them would be helpful if that is our preferred way.

CarlosGamero: Yeah, the doc about hooks is missing 😓, if you agree with this change we should improve the overall doc, and in this particular case I think  we can create a `UPGRADING.md` to explain breaking changes and how to approach them

packages/app/background-jobs-common/src/background-job-processor/utils.ts
CarlosGamero: Do you have any particular name in mind? I don't have a strong opinion but for me `sanitize` for me sounds valid as we are removing from the config those fields that are not supported by BullMQ (will produce an error on init)

packages/app/background-jobs-common/src/background-job-processor/utils.ts
CarlosGamero: You are absolutely right! will add it :D

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
CarlosGamero: Good point, it can be a bit confusing so will add it :D

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
kibertoad: yeah, that sounds great!

packages/app/background-jobs-common/src/background-job-processor/utils.ts
kibertoad: maybe, if it's not confusing for you, then it's good enough, let's keep it

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
CarlosGamero: It is valuable, but we can't test it easily from `AbstractBackgroundJobProcessor` as `resolveExecutionLogger` is no longer visible at this level, I was thinking of adding those checks at `BackgroundJobProcessorMonitor` but it is not there yet (those tests are my next priority on this PR), does it sound good?

kibertoad: yeah, sounds good!

packages/app/background-jobs-common/src/background-job-processor/utils.ts
CarlosGamero: I am more than open to change it, not saying it is perfect but nothing better comes to my mind (what is normal for me, I am super bad with names 😓) so please if you have a better one in mind share it and I will update it :D  

Also, note that this is a library private method, it is not exposed

packages/app/background-jobs-common/src/background-job-processor/processors/FakeBackgroundJobProcessor.ts
```diff
@@ -41,22 +37,7 @@ export class FakeBackgroundJobProcessor<
...
+  protected override process(): Promise<void> {
     return Promise.resolve()
   }
-
-  /**
-   * @deprecated use job spy instead
-   */
-  public get processCalls(): JobData[] {
-    return this._processCalls
-  }
```
CarlosGamero, line 54: Using major version to remove this, long time since spies were implemented, we should be in a good place to remove it already

packages/app/background-jobs-common/UPGRADING.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Upgrading from `4.0.0` to `5.0.0`
+
+### Description of Breaking Changes
+
+1. **Property Removal**
+  - The deprecated `processCalls` property from `FakeBackgroundJobProcessor` has been removed.
+
+2. **Method Changes in `AbstractBackgroundJobProcessor`**
+  - The method `getActiveQueueIds` has been removed. A new constant method now serves this purpose.
```
kibertoad, line 11: `A new constant method now serves this purpose.` -> this sounds unclear, which constant method does this refer to?

packages/app/background-jobs-common/UPGRADING.md
```diff
@@ -0,0 +1,45 @@
...
+  - `logJobStarted` and `logJobFinished` methods are no longer public. The SDK now offers basic logging, and there are
```
kibertoad, line 12: can we mention these alternative ways?

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -215,6 +141,7 @@ export abstract class AbstractBackgroundJobProcessor<
   }
 
   private registerListeners() {
+    // TODO: extract hooks handling to a separate class
```
CarlosGamero, line 144: To not delay this more and not have an even bigger PR, as hooks refactor will not imply a breaking change, leaving it for a follow-up PR

packages/app/background-jobs-common/UPGRADING.md
```diff
@@ -0,0 +1,45 @@
...
+    alternative ways to extend logging if needed.
+  - The method `resolveExecutionLogger` can no longer be overridden. The SDK will now automatically create the
```
kibertoad, line 14: what if some library user has different logging needs and e. g. wants to have different fields on child logger or use different logger altogether?

packages/app/background-jobs-common/UPGRADING.md
```diff
@@ -0,0 +1,45 @@
...
+  - The following methods were not designed to be public and are no longer accessible:
+    - `daysToSeconds`
+    - `daysToMilliseconds`
+    - `isUnrecoverableJobError`
+    - `sanitizeRedisConfig`
+
+### Migration Steps
+
+#### For `FakeBackgroundJobProcessor`
+- If you were using the `processCalls` property, you should remove it and switch to using the spy feature which was
```
kibertoad, line 31: can you link to the corresponding documentation section?

packages/app/background-jobs-common/UPGRADING.md
kibertoad: ok, I see there are instructions further in the doc. maybe it's worth mentioning that for details check the migration steps?

kibertoad (approved):

packages/app/background-jobs-common/UPGRADING.md
CarlosGamero: I think this is covered on the "migration steps" section, right?

packages/app/background-jobs-common/UPGRADING.md
CarlosGamero: Forgot to paste the name, sorry 😓

CarlosGamero: Actually it is described but on migration steps, this part is like a summary of the breaking changes, do you think we should also mention it here?

kibertoad: mentioning name or mentioning that you should check the migration steps should suffice

packages/app/background-jobs-common/UPGRADING.md
kibertoad: kinda, but this is one change I'm slightly torn on, making it protected *might* be useful in some cases, but I don't have strong opinion on this, it's ok to leave it as is

CarlosGamero: I had the same concern at the beginning, but from what I saw we are not using it and I am sure if having it there "just in case" is a good idea so decided to remove it for now

CarlosGamero: If we see it is useful in the future we can put it back easily

packages/app/background-jobs-common/UPGRADING.md
CarlosGamero: Added a link to the migration steps section where it is described

packages/app/background-jobs-common/UPGRADING.md
kibertoad: well, we are not using it because default implementation is custom-tailored to Lokalise, I was more concerned about the users outside of it :D

packages/app/background-jobs-common/UPGRADING.md
CarlosGamero: Done 🙌 thank youuu

packages/app/background-jobs-common/UPGRADING.md
kibertoad: not a problem per se, but the kind of flexibility that I think can be useful. but as I've said, I'm also OK to keep it as is

PR #258, opened 2024-08-01
Use separate checks for separate healthchecks

kibertoad:
## Changes

Previously it will only register Gauges for the very first healthcheck.

## Checklist

### DIFF
--- a/packages/app/healthcheck-utils/src/HealthcheckResultsStore.ts
+++ b/packages/app/healthcheck-utils/src/HealthcheckResultsStore.ts
@@ -8,7 +8,7 @@ export type HealthcheckEntry = {
 }
 
 export type HealthcheckResultsStoreParams = {
-  healthcheckNumber: number // maximum amount of healthchecks that the system can have
+  maxHealthcheckNumber: number // maximum amount of healthchecks that the system can have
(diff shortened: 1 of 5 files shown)


packages/app/healthcheck-utils/src/HealthcheckResultsStore.ts
```diff
@@ -8,7 +8,7 @@ export type HealthcheckEntry = {
 }
 
 export type HealthcheckResultsStoreParams = {
-  healthcheckNumber: number // how many healthchecks does the system have
+  healthcheckNumber: number // maximum amount of healthchecks that the system can have
```
CarlosGamero, line 11: would make sense to change it to `maxHealthcheckNumber` or `maxHealthcheckSupported`, but it would be a breaking change, maybe we can make a not to do it on next major release

CarlosGamero (approved):

packages/app/healthcheck-utils/src/HealthcheckResultsStore.ts
kibertoad: it wasn't used yet, so maybe we can sneak that one under the radar :D

PR #262, opened 2024-08-06
NTC minor refactor

CarlosGamero:
## Changes

Doing some refactor to facilitate addition of new features in the library

## Checklist


packages/app/non-translatable-markup/src/methods/isTextTranslatable.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+  })
+
+  it('should return false if text contains only non-translatable content', () => {
+    const testCases = [
+      '\uE101Hello world!\uE102',
+      ' \uE101Hello world!\uE102 \uE101How are you\uE102 ',
+    ]
+
+    for (const testCase of testCases) {
```
ondrejsevcik, line 34: 🟢 You can use [each](https://vitest.dev/api/#test-each)  for the test cases.

packages/app/non-translatable-markup/vite.config.ts
```diff
@@ -11,6 +11,7 @@ export default defineConfig({
   dependencies: Object.keys(packageJson.peerDependencies),
   test: {
     coverage: {
+      exclude: ['src/**/index.ts'],
```
kibertoad, line 14: do we need all the internal barrel files?

packages/app/non-translatable-markup/src/methods/isTextTranslatable.spec.ts
CarlosGamero: Absolutely! will change it🙏

packages/app/non-translatable-markup/vite.config.ts
CarlosGamero: Going to merge as I would need these changes to continue with my task but if you have concerns, please feel free to leave them here and I will address them on follow-up

kibertoad: there are two index.ts files - the root one and one inside `methods`. Is second one useful?

CarlosGamero: Oh, I see your point, and at the beginning, I thought is was going to be bigger, but looking at the result ... it doesn't make too much sense ... will remove it on follow-up PR

PR #263, opened 2024-08-06
Add AbstractPeriodicJob to the bg job lib

kibertoad:
## Changes

Make the AbstractPeriodicJob reusable across libraries

## Checklist


packages/app/healthcheck-utils/src/AbstractPeriodicJob.ts
```diff
@@ -113,7 +113,7 @@ export abstract class AbstractPeriodicJob {
       shouldLogExecution: false,
       ...options,
       singleConsumerMode: {
-        enabled: true,
+        enabled: false,
```
kibertoad, line 116: disabled by default seems to be a less surprising setup for me

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
```diff
@@ -0,0 +1,301 @@
...
+import { resolveGlobalErrorLogObject } from '@lokalise/node-core'
+import type Redis from 'ioredis'
+import { stdSerializers } from 'pino'
+import type { LockOptions } from 'redis-semaphore'
+import { Mutex } from 'redis-semaphore'
+import type { LockLostCallback } from 'redis-semaphore'
+import type { ToadScheduler } from 'toad-scheduler'
+import { AsyncTask, SimpleIntervalJob } from 'toad-scheduler'
+
+const DEFAULT_JOB_INTERVAL = 60000
```
CarlosGamero, line 18: Does it mean that by default job will be executed once per minute? could we put a comment to clarify it 🙏

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
```diff
@@ -0,0 +1,301 @@
...
+const DEFAULT_EXCLUSIVE_LOCK_SUFFIX = 'EXCLUSIVE'
+
+export type BackgroundJobConfiguration = {
```
CarlosGamero, line 21: Nit: To make this file easier to read, could we extract types a separate file

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
```diff
@@ -0,0 +1,301 @@
...
+
+    /**
+     * Lock will be reset to this value after success, so that other node could potentially acquire the lock after it expires, but in order to prevent immediate acquire
+     */
+    lockTimeoutAfterSuccess?: number
+  }
+  /**
+   * If true, the job will log when it starts and finishes.
+   */
+  shouldLogExecution?: boolean
```
CarlosGamero, line 50: Shouldn't we always log start and finish?

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
```diff
@@ -0,0 +1,301 @@
...
+export type PeriodicJobDependencies = {
+  redis: Redis
+  logger: CommonLogger
+  transactionObservabilityManager: TransactionObservabilityManager
+  errorReporter: ErrorReporter
+  scheduler: ToadScheduler
+}
+
+export abstract class AbstractPeriodicJob {
+  public readonly options: Required<BackgroundJobConfiguration>
```
CarlosGamero, line 97: Should options be public?

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  public async dispose() {
+    this.scheduler.stopById(this.options.jobId)
+    await this.singleConsumerLock?.release()
+  }
+
+  protected abstract processInternal(executionUuid: string): Promise<unknown>
+
+  public async process(executorId: string) {
```
CarlosGamero, line 158: On `AbstractBackgroundJobProcessor` we use these names but in the other direction, should we follow the same approach? (`process` for the abstract method and `processInternal` for the private one)

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
kibertoad: after thinking about it some more, I've removed the default, as every job is going to be different, having arbitrary default is going to be more harmful than useful

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
kibertoad: not sure. note that there needs to be a public entry point that toad-scheduler task can invoke. would you call that `processInternal`?

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
CarlosGamero: I agree with that! thanks Igor 🙏

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
CarlosGamero: I see and it makes sense, but if I am not wrong it is `false` by default, I would expect to have it `true` but no strong opinion from my side

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
CarlosGamero: Oh I see, sorry for the misunderstanding, wondering if we can make the method private and `createTask`  receive the method to call, something like:
```
createTask(logger: CommonLogger, jobId: string, method: (execId: string) => Promise<void>): AsyncTask
```

But I see createTask is exposed, so maybe it is not a good idea 🤔

kibertoad: yeah, I would probably keep it as-is, not sure that tinkering with this part would decrease complexity significantly

CarlosGamero: Yeah, I am fine with it, my only point was not allowing to directly call that method or override it

CarlosGamero (approved):

PR #265, opened 2024-08-07
AP-4762 Symbols detection

CarlosGamero:
## Changes

isTextTranslatable consider symbols and numbers as non translatable

## Checklist


packages/app/non-translatable-markup/src/methods/index.ts
```diff
@@ -1,2 +0,0 @@
-export { isAttemptToEditNonTranslatableContent } from './isAttemptToEditNonTranslatableContent'
```
CarlosGamero, line 1: Removing this file as discussed on prev PR

packages/app/non-translatable-markup/src/methods/isAttemptToEditNonTranslatableContent.ts
```diff
@@ -9,9 +8,8 @@ export const isAttemptToEditNonTranslatableContent = (
   text: string,
   updatedText: string,
 ): boolean => {
-  if (!isTextTranslatable(updatedText)) {
-    return true
-  }
+  // early return if the text is the same
+  if (text === updatedText) return false
```
CarlosGamero, line 12: `isTextTranslatable` can't be used here now, but looking at the method I think that it is fine to stay without it and added a new condition which is a better early return

CarlosGamero:
Lint issue addressed on -> https://github.com/lokalise/shared-ts-libs/pull/266

packages/app/non-translatable-markup/src/methods/utils.ts
```diff
@@ -1,8 +1,28 @@
...
+ *      `\p{Emoji}` -> matches any emoji character
+ *  `u` -> enables unicode mode
+ *
+ * Note: `\p{S}` matches any kind of symbol character, and it covers all the symbols from `\p{Sm}`, `\p{Sc}`, `\p{Sk}`,
+ *  and `\p{So}` categories. However, we don't want to cover `\p{Sc}` (Currency Symbol), so we specify the relevant
+ *  subcategories separately.
+ *
+ * For more info see -> https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/regular_expressions/unicode_character_class_escape/index.md
+ */
+const symbolsPattern = /[\p{P}\p{Sm}\p{Sk}\p{So}\p{Emoji}']/u
```
kjamrog, line 26: If I see correctly, for checking symbols you split text by symbols and then check each part recursively? Won't it work if we use:
```
/[\p{P}\p{Sm}\p{Sk}\p{So}\p{Emoji}']+/u
```
And then check if whole string is matching regex?

packages/app/non-translatable-markup/src/methods/isTextTranslatable.ts
```diff
@@ -1,10 +1,25 @@
...
-  const parts = text.split(nonTranslatableTextRegexp).map((part) => part.trim())
+export const isTextTranslatable = (text: string): boolean =>
+  trimAndFilterEmpty(text.split(nonTranslatableTextRegexp)).some((piece) =>
+    isPieceTranslatable(piece),
+  )
 
-  return parts.some((part) => part !== '')
+const isPieceTranslatable = (piece: string): boolean => {
+  // numbers are not translatable
+  if (!Number.isNaN(+piece)) return false
```
kjamrog, line 15: Maybe for simplicity we could include numbers in regex pattern?

CarlosGamero: Good point! done, thank youuu

packages/app/non-translatable-markup/src/methods/utils.ts
CarlosGamero: Fixed, thank youuu

kjamrog (approved):

packages/app/non-translatable-markup/src/methods/isAttemptToEditNonTranslatableContent.ts
kibertoad: is the inversion from false to true intended here?

packages/app/non-translatable-markup/src/methods/isTextTranslatable.ts
```diff
@@ -1,10 +1,16 @@
...
+ * Returns true if the text contain at least one translatable piece of text.
+ * A translatable piece of text is a piece of text that is not surrounded by non-translatable tags,
+ * symbols or a number.
  */
-export const isTextTranslatable = (text: string): boolean => {
-  const parts = text.split(nonTranslatableTextRegexp).map((part) => part.trim())
+export const isTextTranslatable = (text: string): boolean =>
+  text
+    .split(nonTranslatableTextRegexp)
+    .map((p) => p.trim())
```
kibertoad, line 11: it's not clear what `p` stands for, can we make it more explicit?

packages/app/non-translatable-markup/src/methods/isAttemptToEditNonTranslatableContent.ts
CarlosGamero: Yep, is `text` and `updatedText` is the same it means that user is not trying to edit non translatable content

packages/app/non-translatable-markup/src/methods/isTextTranslatable.ts
CarlosGamero: Good point! already done, thank youuu

kibertoad (approved):

PR #270, opened 2024-08-09
id-utils biome

CarlosGamero:
## Checklist


packages/app/background-jobs-common/vitest.config.mts
```diff
@@ -5,7 +5,6 @@ import defineConfig from '@lokalise/package-vite-config/package'
 // @ts-ignore
 import packageJson from './package.json'
 
-/* eslint-disable import/no-default-export */
```
kibertoad, line 8: we don't need a biome override here?

packages/app/id-utils/package.json
```diff
@@ -1,53 +1,48 @@
...
+    "ulid-uuid-converter": "^1.0.4",
+    "ulidx": "^2.3.0",
+    "uuidv7": "^1.0.1"
+  },
+  "devDependencies": {
+    "@biomejs/biome": "^1.8.3",
+    "@lokalise/biome-config": "^1.3.0",
+    "@lokalise/package-vite-config": "latest",
+    "@vitest/coverage-v8": "^2.0.4",
+    "rimraf": "^5.0.7",
```
kibertoad, line 43: let's bump this to 6.x now that we are at it

packages/app/id-utils/package.json
```diff
@@ -1,53 +1,47 @@
...
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js"
+    }
+  },
+  "scripts": {
+    "build": "rimraf dist && tsc",
```
kibertoad, line 23: why remove vite builds? wouldn't that break compatibility with browser and ESM?

CarlosGamero: Mmm good point, will put it back, sorry

packages/app/id-utils/package.json
CarlosGamero: Done, thank youuu 🙏

kibertoad (approved):

packages/app/background-jobs-common/vitest.config.mts
CarlosGamero: Not ask me why, but is not needed on `vitest.config` override, maybe it is disabled by default on biome

PR #271, opened 2024-08-09
Add cron support

kibertoad:
## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
+++ b/packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
@@ -10,7 +10,7 @@ import type Redis from 'ioredis'
 import type { LockOptions } from 'redis-semaphore'
 import { Mutex } from 'redis-semaphore'
 import type { LockLostCallback } from 'redis-semaphore'
-import type { ToadScheduler } from 'toad-scheduler'
+import { CronJob, type ToadScheduler } from 'toad-scheduler'
 import { SimpleIntervalJob } from 'toad-scheduler'
 import type {
   BackgroundJobConfiguration,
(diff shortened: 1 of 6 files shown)


arthuracs (approved):

packages/app/healthcheck-utils/package.json
```diff
@@ -32,12 +32,14 @@
...
   },
   "peerDependencies": {
+    "@lokalise/background-jobs-common": ">=5.0.0",
     "ioredis": "^5.4.1",
     "prom-client": "^15.1.3",
     "toad-scheduler": "^3.0.1"
   },
   "devDependencies": {
     "@biomejs/biome": "^1.8.3",
+    "@lokalise/background-jobs-common": "^5.2.0",
```
CarlosGamero, line 42: As this is a `major` update, i think we should do this in a different PR to not trigger release

kibertoad: bah, good catch, this leaked from a different set of changes that I was doing

packages/app/background-jobs-common/src/periodic-jobs/AbstractPeriodicJob.ts
```diff
@@ -49,7 +50,9 @@ export abstract class AbstractPeriodicJob {
       singleConsumerMode: {
         enabled: false,
         exclusiveLockSuffix: options.singleConsumerMode?.exclusiveLockSuffix ?? 'EXCLUSIVE',
-        lockTimeout: options.intervalInMs * 2,
+        lockTimeout: options.schedule.intervalInMs
+          ? options.schedule.intervalInMs * 2
+          : DEFAULT_LOCK_TIMEOUT,
```
CarlosGamero, line 55: If we use cron option, `lockTimeout` will always be the default value, is it fine?

kibertoad: not super sure about it, but what would be a better alternative? you can still specify an explicit value if needed, not sure if we need a separate value for specifying a default

CarlosGamero: Mmm, not sure, either, we can go with it for now and if we see it can be a problem then we can add a new value for it

CarlosGamero (approved):

PR #274, opened 2024-08-12
Migrate to reusable AbstractPeriodicJob

kibertoad:
## Changes

As per title

## Checklist


CarlosGamero (approved):

packages/app/healthcheck-utils/src/HealthcheckRefreshJob.spec.ts
```diff
@@ -97,9 +97,6 @@ describe('HealthcheckRefreshJob', () => {
   beforeEach(() => {
     store.resetHealthcheckStores()
   })
-  afterEach(async () => {
-    await job.dispose()
```
kibertoad, line 101: we are not registering job anywhere, so it doesn't need to be disposed

PR #282, opened 2024-08-13
Prisma client builder

CarlosGamero:
## Changes

Adding PrismaClient builder method to be able to add a default config

## Checklist


packages/app/prisma-utils/src/types.ts
```diff
@@ -42,3 +47,16 @@ export type PrismaTransactionReturnType<T> = Either<
...
+// Prisma client builder types
+//----------------------------------------
+/**
+ * this is a temporal solution in the meantime Prisma includes ReadCommitted as a valid isolation level for CockroachDB
+ */
+export type PrismaClientBuilderOptions = Omit<Prisma.PrismaClientOptions, 'transactionOptions'> & {
+  dbDriver?: DbDriver // default: CockroachDb
+  transactionOptions?: Omit<Prisma.PrismaClientOptions['transactionOptions'], 'isolationLevel'> & {
+    isolationLevel?: CockroachDbIsolationLevel
+  }
```
CarlosGamero, line 61: Whenever the Prisma client supports `ReadCommitted` we will be able to remove this overwrite, also adding `dBDriver` there in case we want to have different default configs depending on it

packages/app/prisma-utils/src/prismaClientBuilder.ts
```diff
@@ -0,0 +1,16 @@
+import { PrismaClient } from '@prisma/client'
+import type { PrismaClientBuilderOptions } from './types'
+
+const defaultOptions: PrismaClientBuilderOptions = {
+  transactionOptions: { isolationLevel: 'ReadCommitted' },
+}
+
+export const prismaClientBuilder = (options: PrismaClientBuilderOptions = {}): PrismaClient => {
```
kibertoad, line 8: builder pattern assumes preparing configuration via fluent API and then invoking build(), having it as just a function might be confusing.
maybe we can call it a factory?

CarlosGamero: You are absolutely right! I will address the change, thanks Igor!

CarlosGamero: Change applied!

kibertoad (approved):
great job!

PR #283, opened 2024-08-13
Prisma utils build fix

CarlosGamero:
## Changes

Trying to fix build issue from #282 

## Checklist


packages/app/prisma-utils/package.json
```diff
@@ -20,7 +20,7 @@
         }
     },
     "scripts": {
-        "build": "rimraf dist && npm run db:update-client && vite build",
+        "build": "rimraf dist && npm run db:update-client && tsc",
```
CarlosGamero, line 23: I guess we don't need vite here, but please correct me if it is not true 🙏

kibertoad: yeah, I think we don't, it's BE only

kibertoad (approved):

PR #289, opened 2024-08-15
Implement child job throttling barrier

kibertoad:
## Changes

New barrier.

This is a breaking change due to a change in the generics order in AbstractBackgroundJobProcessor

## Checklist


packages/app/background-jobs-common/src/background-job-processor/barrier/ChildJobThrottlingBarrier.ts
```diff
@@ -0,0 +1,37 @@
...
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  childJobProcessor: AbstractBackgroundJobProcessor<any>
+}
+
+/**
+ * Limit amount of child jobs that can exist at a time.
+ * Note that for performance reasons it performs an optimistic check and can overflow in highly concurrent systems,
+ * so it is recommended to use lower values for max child jobs, to preserve a buffer for the overflow
+ */
+export function createChildJobThrottlingBarrier(
```
eduard-fa, line 20: 🟢 I'm a bit concerned that the term `child` might conflict with the [BullMQ flow's](https://docs.bullmq.io/guide/flows) terminology, potentially causing confusion. Is there a more suitable term we could use instead?
E.g. more abstract `createJobQueueSizeThrottlingBarrier `

eduard-fa (approved):

PR #293, opened 2024-08-16
get job count lazy init

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
@@ -531,7 +531,7 @@ describe('AbstractBackgroundJobProcessor', () => {
   })
 
   describe('getJobCount', () => {
-    it('works', async () => {
+    it('lazy init + job count works as expected', async () => {
       const processor = new FakeBackgroundJobProcessor<JobData>(
         deps,
         'queue1',
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
@@ -529,4 +529,28 @@ describe('AbstractBackgroundJobProcessor', () => {
       })
     })
   })
+
+  describe('getJobCount', () => {
+    it('works', async () => {
+      const processor = new FakeBackgroundJobProcessor<JobData>(
+        deps,
+        'queue1',
+        mocks.getRedisConfig(),
+      )
+      // lazy init
+      expect(await processor.getJobCount()).toBe(0)
+
+      const jobId = await processor.schedule({
+        id: 'test_id',
+        value: 'test',
+        metadata: { correlationId: 'correlation_id' },
+      })
+
+      expect(await processor.getJobCount()).toBe(1)
+      await processor.spy.waitForJobWithId(jobId, 'completed')
+      expect(await processor.getJobCount()).toBe(0)
+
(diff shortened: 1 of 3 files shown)


packages/app/prisma-utils/src/prismaClientFactory.spec.ts
```diff
@@ -17,7 +17,7 @@ describe('prismaClientFactory', () => {
       })
 
       expect(prisma).toBeDefined()
-      expect(await prisma.item1.findMany()).toHaveLength(0)
+      expect(await prisma.item1.findMany()).toBeDefined()
```
CarlosGamero, line 20: Just fixing flaky test, prisma transaction tests are affecting this one, as we don't care about results and we only want to check that the client works so we can avoid a DB clean-up here

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -109,7 +109,8 @@ export abstract class AbstractBackgroundJobProcessor<
     this.runningPromises = []
   }
 
-  public getJobCount(): Promise<number> {
+  public async getJobCount(): Promise<number> {
+    await this.startIfNotStarted()
```
kibertoad, line 113: can we do an explicit synchronous `this.isStarted` check first, to avoid introducing one more promise and one more tick on every call?

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
```diff
@@ -529,4 +529,28 @@ describe('AbstractBackgroundJobProcessor', () => {
       })
     })
   })
+
+  describe('getJobCount', () => {
+    it('works', async () => {
```
kibertoad, line 534: is this a test for lazy init specifically?

CarlosGamero: well, not only lazy init, it is also testing that counting works

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: `startIfNotStarted` is already doing it, do you think doing it also here makes sense?

kibertoad: I have just seen your updated version, sure, yeah, that is better

kibertoad: note that there is still overhead in using Promise.resolve over purely synchronous code, and it also relies on microtasks to work, but it is, indeed, faster than async functions, so it is fine to keep it like this

kibertoad (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: I completely agree with you and thank you so much for giving the extended info, I really like and appreciate your comments/explanation, it is really useful to learn ❤️

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
CarlosGamero: You are right, I did it quickly so didn't think too much on name ... sorry, will improve it now

PR #302, opened 2024-08-21
MET-188 Add support for repeatable jobs

dariacm:
## Changes

[Repeatable jobs](https://docs.bullmq.io/guide/jobs/repeatable) can be scheduled by passing the repeat settings. However, these settings are only accepted when using the [add()](https://github.com/taskforcesh/bullmq/blob/c60e6a3f45606c623ef6a2b16415d67eac9b0269/src/classes/queue.ts#L218) method and are ignored when using [addBulk()](https://github.com/taskforcesh/bullmq/blob/c60e6a3f45606c623ef6a2b16415d67eac9b0269/src/classes/queue.ts#L264) where [they are omitted](https://github.com/taskforcesh/bullmq/blob/c60e6a3f45606c623ef6a2b16415d67eac9b0269/src/interfaces/minimal-job.ts#L5). This PR uses `add()` instead of reusing `scheduleBulk()` in the `schedule()` method so that repeatable jobs can be scheduled and prevents this from happening in the future shall they introduce additional properties that are only processed in `add()`.

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
@@ -237,11 +237,25 @@ export abstract class AbstractBackgroundJobProcessor<
   }
 
   public async schedule(jobData: JobPayload, options?: JobOptionsType): Promise<string> {
-    const jobIds = await this.scheduleBulk([jobData], options)
-    return jobIds[0]
+    await this.startIfNotStarted()
+
+    const job = await this._queue?.add(this.config.queueId, jobData, prepareJobOptions(this.config.isTest, options))
+    if (!job?.id) {
+      throw new Error('Scheduled job ID is undefined')
+    }
+
+    if (this._spy) {
+      this._spy.addJob(job, 'scheduled')
+    }
+
+    return job.id
   }
 
   public async scheduleBulk(jobData: JobPayload[], options?: JobOptionsType): Promise<string[]> {
+    if (options?.repeat) {
+      throw new Error('scheduleBulk does not support repeatOptions. Please use schedule method instead')
+    }
+
     await this.startIfNotStarted()
 
     const jobs =
@@ -262,9 +276,7 @@ export abstract class AbstractBackgroundJobProcessor<
     }
 
     if (this._spy) {
-      for (const job of jobs) {
-        this._spy.addJob(job, 'scheduled')
-      }
+      this._spy.addJobs(jobs, 'scheduled')
     }
 
     return jobIds as string[]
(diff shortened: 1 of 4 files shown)


packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -237,11 +237,31 @@ export abstract class AbstractBackgroundJobProcessor<
...
+    return job.id
   }
 
   public async scheduleBulk(jobData: JobPayload[], options?: JobOptionsType): Promise<string[]> {
+    if (options?.repeat) {
+      throw new Error(
+        'scheduleBulk does not support repeatOptions. Please use schedule method instead',
+      )
+    }
+
```
CarlosGamero, line 264: If `repeat` does not work with the bulk method, instead of throwing an error should we just exclude it from the type?

kibertoad: wouldn't TS still accept jobs with unnecessary fields?

dariacm: +1 on what Igor said. That's basically what was happening but further when we were calling BullMQ, it was accepting `repeat` settings but not using them since they're excluded, and it's misleading because you don't get any error but your job is scheduled as a regular one, not a repeatable one 😕

CarlosGamero: TS will accept it, you are right, but if it is not part of API as a user I would expect it to not affect the method which is what is currently happening (bullMQ is ignoring it)

CarlosGamero: Actually, won't TS complain if we pass a not-allowed prop 🤔 
Screenshot 2024-08-21 at 15 19 08

In any case, it is still possible to pass it but as a developer, if it is not defined on API I won't expect it to work.

That being said if you think that the error is a better approach, I have no strong opinion, and feel free to go with this approach if it looks better for you 🙇

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
dariacm: I think you're right @CarlosGamero. I didn't notice the TS error yesterday, but it's definitely there, so that should be enough to discourage wrong usage imo. Let me update this

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
kibertoad: awesome, even better!

PR #303, opened 2024-08-23
Restoring UNKNOWN_RESPONSE_SCHEMA schema

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/backend-http-client/src/client/constants.ts
+++ b/packages/app/backend-http-client/src/client/constants.ts
@@ -29,4 +29,13 @@ export const TEST_OPTIONS: RequestOptions<unknown, any> = {
   responseSchema: z.unknown(),
 }
 
+/**
+ * Technically 204 will send an empty body, but undici-retry defaults to parsing unknown mimetype as text for
+ * compatibility reasons, so we should expect to get an empty string here
+ */
 export const NO_CONTENT_RESPONSE_SCHEMA = z.string().length(0)
+
+/**
+ * This schema is to be used when we don't really care about the response type and are prepared to accept any value
+ */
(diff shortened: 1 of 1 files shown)


packages/app/backend-http-client/src/client/constants.ts
```diff
@@ -29,4 +29,13 @@ export const TEST_OPTIONS: RequestOptions<unknown, any> = {
   responseSchema: z.unknown(),
 }
 
+/**
+ * Technically 204 will send an empty body, but undici-retry defaults to parsing unknown mimetype as text for
+ * compatibility reasons, so we should expect to get an empty string here
+ */
```
CarlosGamero, line 35: Restored also comment as I think it can be useful

packages/app/backend-http-client/src/client/constants.ts
```diff
@@ -29,4 +29,13 @@ export const TEST_OPTIONS: RequestOptions<unknown, any> = {
...
 export const NO_CONTENT_RESPONSE_SCHEMA = z.string().length(0)
+
+/**
+ * This schema is to be used when we don't really care about the response type and are prepared to accept any value
+ */
+export const UNKNOWN_RESPONSE_SCHEMA = z.unknown()
```
CarlosGamero, line 41: @kibertoad The on we had was `unknown`, I think it is better than `any` but please let me know if disagree

kibertoad, line 41: we also need to update exports, I think

CarlosGamero: Ups, sorry ... I thought we were exporting all, thank you 🙏

kibertoad (approved):

PR #304, opened 2024-08-23
Improving bg common logging

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
@@ -1,7 +1,8 @@
-import type {
-  CommonLogger,
-  RedisConfig,
-  TransactionObservabilityManager,
+import {
+  type CommonLogger,
+  type RedisConfig,
+  type TransactionObservabilityManager,
+  resolveGlobalErrorLogObject,
 } from '@lokalise/node-core'
 import Redis from 'ioredis'
 import { QUEUE_IDS_KEY } from '../constants'
@@ -89,18 +90,32 @@ export class BackgroundJobProcessorMonitor<
   public jobStart(job: JobType, requestContext: RequestContext): void {
     const transactionName = `bg_job:${this.ownerName}:${this.queueId}`
     this.transactionObservabilityManager.start(transactionName, resolveJobId(job))
-    requestContext.logger.info({ origin: this.processorName }, `Started job ${job.name}`)
+    requestContext.logger.info(this.buildLogParams(job), `Started job ${job.name}`)
+  }
+
+  public jobTryError(job: JobType, error: unknown, requestContext: RequestContext): void {
+    requestContext.logger.error(this.buildLogParams(job, error), `${job.name} try failed`)
   }
 
   public jobEnd(job: JobType, requestContext: RequestContext): void {
     requestContext.logger.info(
       {
-        origin: this.processorName,
+        ...this.buildLogParams(job),
         isSuccess: job.progress === 100,
       },
       `Finished job ${job.name}`,
     )
     this.transactionObservabilityManager.stop(resolveJobId(job))
+  }
+
+  private buildLogParams(job: JobType, error?: unknown) {
+    return {
+      origin: this.processorName,
+      jobId: resolveJobId(job),
+      jobName: job.name,
+      jobProgress: job.progress,
+      ...(error ? resolveGlobalErrorLogObject(error) : {}),
+    }
   }
 }
 
--- a/packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
@@ -93,7 +93,7 @@ export class BackgroundJobProcessorMonitor<
     requestContext.logger.info(this.buildLogParams(job), `Started job ${job.name}`)
   }
 
-  public jobTryError(job: JobType, error: unknown, requestContext: RequestContext): void {
+  public jobAttemptError(job: JobType, error: unknown, requestContext: RequestContext): void {
     requestContext.logger.error(this.buildLogParams(job, error), `${job.name} try failed`)
   }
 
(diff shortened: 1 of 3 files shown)


packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
```diff
@@ -89,19 +90,33 @@ export class BackgroundJobProcessorMonitor<
...
       },
       `Finished job ${job.name}`,
     )
     this.transactionObservabilityManager.stop(resolveJobId(job))
   }
+
+  private buildLogParams(job: JobType, error?: unknown) {
+    return {
+      origin: this.processorName,
+      jobId: resolveJobId(job),
```
kibertoad, line 114: isn't jobId coming from the childLogger already?

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
```diff
@@ -89,19 +90,33 @@ export class BackgroundJobProcessorMonitor<
...
         isSuccess: job.progress === 100,
       },
       `Finished job ${job.name}`,
     )
     this.transactionObservabilityManager.stop(resolveJobId(job))
   }
+
+  private buildLogParams(job: JobType, error?: unknown) {
+    return {
+      origin: this.processorName,
```
kibertoad, line 113: wonder if more of these fields should be moved to the childLogger building step

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
```diff
@@ -89,19 +90,33 @@ export class BackgroundJobProcessorMonitor<
   public jobStart(job: JobType, requestContext: RequestContext): void {
     const transactionName = `bg_job:${this.ownerName}:${this.queueId}`
     this.transactionObservabilityManager.start(transactionName, resolveJobId(job))
-    requestContext.logger.info({ origin: this.processorName }, `Started job ${job.name}`)
+    requestContext.logger.info(this.buildLogParams(job), `Started job ${job.name}`)
+  }
+
+  public jobTryError(job: JobType, error: unknown, requestContext: RequestContext): void {
```
kibertoad, line 96: name is slightly confusing, what does `jobTryError` mean?

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
CarlosGamero: you are completely right! I will revisit what is on child logger and remove from here, sorry 🙏

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
CarlosGamero: Not really, origin is where is the log generated, if the request context is passed to a service/use case, origin should not be there from my point of view, but please feel free to disagree

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
CarlosGamero: it is that the job throws an error on the process try, but it does not have to be a definitive error (the job can be retried depending on the config).

I agree that it is confusing, but I am super bad at naming, let me give it a thought but please if you have something in mind I am more than open to ideas 🙏

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
kibertoad: I don't mean the requestContext that we pass any deeper, just the logger that we instantiate for the job. I mean this part:

```ts
    const requestContext: RequestContext = {
      reqId: jobId,
      logger: new BackgroundJobProcessorLogger(
        this.logger.child({
          // add more stuff here, it won't propagate anywhere else
          'x-request-id': job.data.metadata.correlationId,
          jobId,
          jobName: job.name,
        }),
        job,
      ),
    }
```

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
kibertoad: Do you mean BullMQ "job attempt" or something else?

CarlosGamero: Yes, the job attempt, maybe `jobAttemtError` is meaningful enough?

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
CarlosGamero: Yeah, but we should always use `requestContext.logger`, right?
If, from the `processor.process` method, we call to a use-case or service where we need to log something we should be passing `requestContext` and in that case origin is not correct

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
kibertoad: sounds good!

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
kibertoad: ah, fair point.
but isn't that already a problem?

CarlosGamero: for other properties, it is ...  well I guess it's fine to put the job id or job name to see which was the "starter" of the process but we are using `origin`  to put the class from where the log is written

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
CarlosGamero: Done, thanks 🙏

kibertoad (approved):

PR #309, opened 2024-08-28
Adding isNonEmptyArray

CarlosGamero:
## Changes

Adding `isNonEmptyArray` method to facilitate working with `nonempty` constraint from Zod

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/src/array/isNonEmptyArray.ts
+++ b/packages/app/universal-ts-utils/src/array/isNonEmptyArray.ts
@@ -1,0 +1,12 @@
+/**
+ * Checks if the given array is non-empty.
+ *
+ * This function is a type guard that not only checks whether the array has at least one element,
+ * but also refines the type of the array to be a tuple indicating that the first element exists.
+ * This is useful for preventing operations on empty arrays and for gaining type-level assurances.
+ *
+ * @template T - The type of elements within the array.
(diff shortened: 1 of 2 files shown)


packages/app/universal-ts-utils/src/array/isNonEmptyArray.ts
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * This function is a type guard that not only checks whether the array has at least one element,
+ * but also refines the type of the array to be a tuple indicating that the first element exists.
+ * This is useful for preventing operations on empty arrays and for gaining type-level assurances.
+ *
+ * @template T - The type of elements within the array.
+ * @param {readonly T[]} array - The array to be checked, which should be read-only.
+ * @returns {array is [T, ...T[]]} - Returns true if the array is non-empty, false otherwise.
+ */
+export const isNonEmptyArray = <T>(array: readonly T[]): array is [T, ...T[]] => array.length > 0
```
kibertoad, line 12: is it correct to transform readonly array into non-readonly array here?

CarlosGamero: Mmm, good point! also, readonly does not make too much sense for this method (we are just checking length) removing it

kibertoad (approved):

PR #330, opened 2024-09-18
AP-4770 method for extracting translatable text pieces

kjamrog:
## Changes

Adds method for extracting translatable text parts, excluding NT parts and HTML tags. It will be used for words counting in Flow.

## Checklist


packages/app/non-translatable-markup/src/methods/extractTranslatableTextPieces.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // HTML tags - should be removed
+    { text: 'Hello!<> world', result: ['Hello!', 'world'] },
+    { text: '<div class="test">hello world</div>', result: ['hello world'] },
+    { text: 'hello world</tr>', result: ['hello world'] },
+    { text: 'Hello</br>world </br>', result: ['Hello', 'world'] },
+    { text: '</tr>', result: [] },
+    { text: '<tr>', result: [] },
+    { text: '<p></p>', result: [] },
+    { text: '<div class="test">\n\t&🚀</div>', result: ['&🚀'] }, // White characters are removed
```
CarlosGamero, line 25: 🟡 Do we consider `&🚀` translatable? I think that on the `isTextTranslatable` method we don't (please correct me if I am wrong), should we follow the same rules?

packages/app/non-translatable-markup/src/methods/extractTranslatableTextPieces.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    { text: '<tr', result: ['<tr'] }, // Unclosed tag
+
+    // Symbols and emojis
+    { text: '\n', result: [] },
+    { text: '🔥', result: ['🔥'] },
```
CarlosGamero, line 30: Same with this

packages/app/non-translatable-markup/src/methods/extractTranslatableTextPieces.spec.ts
kjamrog: In that method, we only extract parts of texts between NT and HTML tags, without looking what those parts contain. So no symbols are removed. I though that will be the most flexible as then the client that is using the method can decide what to do with them (e.g. whether we count them or not), I assumed that it can be confusing if we remove some parts of text that are now wrapped in tags. Does it make sense?

CarlosGamero: Yeah, it makes sense to me, but it is not aligned with what "translatable" means in the library, so I would maybe suggest renaming the method to avoid confusion but I don't have a good candidate in mind 🤔

kjamrog: For me, `extractTextBetweenTags` would be better, but we have such private method already and it does quite the opposite: extracts tags with text inside :D. How about renaming old one to `extractTextInTags` ?

So then, for 
```
\uE101{% if MyVariable %}\uE102 Hello
```
- `extractTextInTags` - will return `\uE101{% if MyVariable %}\uE102`
- `extractTextBetweenTags` - will return `Hello`

CarlosGamero: As the current `extractTextBetweenTags` is a private method we can rename it easily, and I love your idea, so please go ahead 🚀

CarlosGamero: Btw, please add good documentation to the method explaining that it ignores NTC regions and HTML like tags, maybe even with examples so it is clear what the method does

CarlosGamero: Btw, maybe we can try to improve a bit the method doc to make to clear what the method does with examples

kjamrog: Updated, let me know if it looks good now

CarlosGamero (approved):
Great job! 🚀

PR #331, opened 2024-09-18
Universal ts lib moving last array utils method from node core

CarlosGamero:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/src/array/areArraysEqual.ts
+++ b/packages/app/universal-ts-utils/src/array/areArraysEqual.ts
@@ -1,0 +1,34 @@
+/**
+ * Check if two arrays are equal.
+ * The arrays are considered equal if they have the same length, same order and the elements at each index by value are equal.
+ * Note: Equality is compared by value and not by reference.
+ */
+export function areArraysEqual(a: unknown[], b: unknown[]): boolean {
+  // Early return if we are comparing the same array
+  if (a === b) return true
+  if (a.length !== b.length) return false
+
+  return a.every((value, index) => areEquals(value, b[index]))
+}
+
+const areEquals = (item1: unknown, item2: unknown): boolean => {
+  if (item1 === item2) return true
+
+  // both should be defined at this point
+  // if both are nullable but with different values (null, undefined) they are not equal
+  if (!item1 || !item2) return false
+
+  // if both are arrays, compare them recursively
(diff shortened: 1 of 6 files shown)


packages/app/universal-ts-utils/src/array/callChunked.ts
```diff
@@ -0,0 +1,11 @@
+
+export async function callChunked<Item>(
+  chunkSize: number,
+  array: readonly Item[],
+  processFn: (arrayChunk: Item[]) => Promise<unknown>,
+): Promise<void> {
```
CarlosGamero, line 6: With this one, all array utils from node-core are already here, so we plan removing them from there

packages/app/universal-ts-utils/src/array/areStringArraysEqual.ts
```diff
@@ -1,4 +1,6 @@
 /**
+ * @deprecated Use `areArraysEqual` instead.
+ *
```
CarlosGamero, line 3: I thought that we can have something more common, and not only for string arrays

packages/app/universal-ts-utils/src/array/areArraysEqual.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Check if two arrays are equal.
+ * The arrays are considered equal if they have the same length, same order and the elements at each index by value are equal.
+ * Note: Equality is compared by value and not by reference.
+ */
+export function areArraysEqual(a: unknown[], b: unknown[]): boolean {
```
CarlosGamero, line 6: Thinking about having `areEquals` so it can be also used to compare objects, still thinking about it

kibertoad: from perf perspective separate functions are better, because isomorphic params (different structure across multiple calls) will cause v8 to deoptimize.

CarlosGamero: I am not sure, my idea was to have something like:
```
areEquals(a: unknown, b: unknown): boolean
```
it won't use isomorphic params but not sure if v8 will deoptimize it

CarlosGamero: As an example -> https://github.com/lokalise/shared-ts-libs/pull/331/commits/db19fb850f9dcd797289e38e4911f47557dad09a

CarlosGamero:
Will open a new PR with final version

PR #333, opened 2024-09-19
Adding index files to universal-ts-utils

CarlosGamero:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/src/index.ts
+++ b/packages/app/universal-ts-utils/src/index.ts
@@ -1,0 +1,1 @@
+export * from './array'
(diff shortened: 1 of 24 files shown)


Dosexe (approved):

packages/app/universal-ts-utils/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './array'
```
kibertoad, line 1: do we need this file.

CarlosGamero: Already removed, I think we can without it for now, if the main "index" begins to be too big we can consider splitting but for now I think we don't need it

kibertoad (approved):

PR #340, opened 2024-09-23
Introduce traversePaginatedSourceByHasMore()

freiondrej:
## Changes

Allow traversing a paginable resource with constant memory usage.

## Checklist

### DIFF
--- a/packages/app/api-common/src/paginationUtils.ts
+++ b/packages/app/api-common/src/paginationUtils.ts
@@ -201,3 +201,22 @@ export async function getPaginatedEntriesByHasMore<T extends Record<string, unkn
 
   return resultArray
 }
+
+export async function traversePaginatedSourceByHasMore<
+  T extends Record<string, unknown>,
+>(
+  pagination: OptionalPaginationParams,
+  apiCall: (params: OptionalPaginationParams) => Promise<PaginatedResponse<T>>,
+  processPage: (pageData: PaginatedResponse<T>['data']) => Promise<void>,
+): Promise<void> {
+  let hasMore: boolean | undefined
+  let currentCursor: string | undefined = pagination.after
+
+  do {
+    const pageResult = await apiCall({ ...pagination, after: currentCursor })
+    hasMore = pageResult.meta.hasMore
+    currentCursor = pageResult.meta.cursor
+
+    await processPage(pageResult.data)
+  } while (hasMore)
+}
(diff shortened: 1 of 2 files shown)


packages/app/api-common/src/paginationUtils.ts
```diff
@@ -201,3 +201,22 @@ export async function getPaginatedEntriesByHasMore<T extends Record<string, unkn
 
   return resultArray
 }
+
+export async function traversePaginatedSourceByHasMore<
```
freiondrej, line 205: As it stands, this is pretty much just a copy of `getPaginatedEntriesByHasMore`, with an addition of the processPage() call.

In the future, I can imagine something more parallelized, with apiCall being called in a promise pool as fast as the processPage() is able to process it (as I assume processPage will be the bottleneck), so that we wouldn't need to keep things in memory.

CarlosGamero, line 205: 🟢 Could we add JSDoc as in the other methods? just to explain clearly how the method works and how it is intended to be used

packages/app/api-common/src/paginationUtils.ts
```diff
@@ -201,3 +201,22 @@ export async function getPaginatedEntriesByHasMore<T extends Record<string, unkn
...
+  T extends Record<string, unknown>,
+>(
+  pagination: OptionalPaginationParams,
+  apiCall: (params: OptionalPaginationParams) => Promise<PaginatedResponse<T>>,
+  processPage: (pageData: PaginatedResponse<T>['data']) => Promise<void>,
+): Promise<void> {
```
CarlosGamero, line 211: 🟢 Could be interesting to be able to return the results of the processing step? something like
```
export async function traversePaginatedSourceByHasMore<
  T extends Record<string, unknown>,
  ReturnType = void,
>(
  pagination: OptionalPaginationParams,
  apiCall: (params: OptionalPaginationParams) => Promise<PaginatedResponse<T>>,
  processPage: (pageData: PaginatedResponse<T>['data']) => Promise<ReturnType> | ReturnType,
): Promise<ReturnType> {
```
An example usage could be to insert something on DB and return all new IDs. What do you think?

freiondrej-lokalise:
@CarlosGamero It has been some time since I first thought this could be useful and it turned out we didn't really need it. I'll close this now. Still, a big thank you for the review! 🙇

PR #348, opened 2024-10-02
Feature/universal ts utils new methods

CarlosGamero:
## Changes

Adding new utils methods:

- callChunked
- sort
- sortByField
- areEquals

## Checklist


packages/app/universal-ts-utils/src/types.ts
```diff
@@ -0,0 +1 @@
+export type NonObject = string | number | boolean | null | undefined | symbol | bigint
```
CarlosGamero, line 1: Not sure if this is a good idea, it will exposed for FE, is it fine?

CarlosGamero: Maybe it is better to define it internally per method, it will be duplicated but in this case I would it is fine (short piece of code)

kibertoad: No strong opinion, either option is OK for me, it's a small amount of code

CarlosGamero: Yeah ... I guess in the future we can come up with more complex shared types and we will have to decide but in this case I would prefer not exposing it

CarlosGamero: Or maybe we can create some kind of "private" folder and adjust `export` on package.json to ignore it 🤔 

But in any case, I think we can do it independently when needed and not for this small piece of code

packages/app/universal-ts-utils/src/array/sortByField.ts
```diff
@@ -0,0 +1,49 @@
...
+}[keyof T]
+
+/*
+ TODO: Extend this method more types of fields.
+ For now, it only supports fields of type string as it is the most common use case.
+ the objective is to use the compare method defined on `sort.ts` but don't want to expose it at this point.
+
+ So in a follow-up will try to come up with a solution to have "internal" not exposed utilities, and then extends this
+ method.
+*/
```
CarlosGamero, line 12: Will come up with a proposal for this in a separate PR

packages/app/universal-ts-utils/src/array/sortByField.ts
```diff
@@ -0,0 +1,49 @@
...
+ * ```
+ */
+export const sortByField = <T extends object, K extends KeysMatching<T, string>>(
+  array: T[],
+  field: K,
+  order: 'asc' | 'desc' = 'asc',
+): T[] =>
+  order === 'asc'
+    ? // @ts-expect-error
+      array.sort((a, b) => a[field].localeCompare(b[field]))
```
ondrejsevcik, line 47: 🟡 The function is using `sort` which modified the original array. It should probably switch to [toSorted](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted).

packages/app/universal-ts-utils/src/object/areEquals.ts
```diff
@@ -0,0 +1,46 @@
...
+ * areEquals(1, 1) // true
+ * areEquals([1, 2], [1, 2]) // true
+ * areEquals({ name: 'John' }, { name: 'John' }) // true
+ * areEquals(null, null) // true
+ * areEquals(undefined, null) // false
+ * areEquals([1, [2, 3]], [1, [2, 3]]) // true
+ * areEquals([{ id: 1 }], [{ id: 1 }]) // true
+ * ```
+ */
+export const areEquals = (a: unknown, b: unknown): boolean => {
```
ondrejsevcik, line 20: 🟢 Just my opinion.

This could be split into three separate functions - isDeepEqual, isShallowEqual and isStrictEqual and then compose those as needed (base on use case). Currently, this function does a bit of everything.

packages/app/universal-ts-utils/src/array/sortByField.ts
CarlosGamero: Good point, Ondrej! Thank you for bringing it up 🙌!

It seems that `toSorted` is currently unavailable in my TypeScript setup, as I encountered this error: `TS2550: Property toSorted does not exist on type string[]`. We might need to adjust our compiler options to enable this feature. Though using `@ts-ignore` makes it work, I'd prefer to avoid that workaround.

For now, I've modified the code to create a sorted copy of the array, which prevents changes to the original array. In a separate pull request, we can explore how to configure TypeScript to recognize `toSorted`, allowing us to use it without causing any breaking changes. What do you think?

packages/app/universal-ts-utils/src/array/sortByField.ts
```diff
@@ -0,0 +1,49 @@
...
+ *   { name: 'John', age: 25 }
+ * ];
+ *
+ * // Sort by 'name'
+ * const sortedByName = sortByField(data, 'name')
+ * // Output: [ { name: 'Alex', age: 30 }, { name: 'John', age: 25 }, { name: 'Zara', age: 22 } ]
+ * console.log(sortedByName)
+ * ```
+ */
+export const sortByField = <T extends object, K extends KeysMatching<T, string>>(
```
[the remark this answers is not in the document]
CarlosGamero: We utilize this feature in our services primarily for testing purposes. you are right, we can do it manually when needed, though having a shared method provides a cleaner and more readable solution.

We already have a similar method implemented in our AP codebase at `services/translations/backend-translation-storage/test/sortUtils.ts`. However, this method can be error-prone due to the lack of field name and type check at compilation time we have on this new version.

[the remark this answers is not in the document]
CarlosGamero: Also note that the case you posted is an easy one, but this method is inteded to be able to sort things like (0, false, true, 1, '3')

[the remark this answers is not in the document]
CarlosGamero: Btw, this method right now only supports strings, the support of what I mentioned before will come in a separate PR.
The method that we will use is the `compare` one from `sort.ts`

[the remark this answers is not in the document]
ondrejsevcik: > Maybe we can leave it outside and add it in the future if we see value on it, what do you think?

Yes! Let's add only code for use-cases we are using. We don't want to maintain code for "maybe use-case some day in the future" ;)

packages/app/universal-ts-utils/src/array/sort.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Sorts an array of primitive types in either ascending or descending order.
```
kibertoad, line 2: Would be nice to mention that we return a sorted copy, current description implies that we sort in place.

packages/app/universal-ts-utils/src/array/sortByField.ts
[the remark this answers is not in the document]
kibertoad: I agree, let's only add utility functions that we actually use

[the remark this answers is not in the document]
CarlosGamero: I see, so looks like we have an agreement 🙌 Thanks for raising this and helping me to make a decision 🙌.

I will update the code to support only the needed use cases (I think it is `string` and `number`)

packages/app/universal-ts-utils/src/array/sort.ts
CarlosGamero: Good point! I will update doc 🙏

packages/app/universal-ts-utils/src/array/sortByField.ts
ondrejsevcik: Yeah, it seems we would need to update TS for that and update tsconfig. Let's leave that for a separate PR then.

packages/app/universal-ts-utils/src/array/sort.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @template T The type of the elements in the array.
+ *
+ * @param {T[]} array - The array to be sorted. This should be an array of primitive values.
+ * @param {'asc' | 'desc'} [order='asc'] - The order in which to sort the array. Defaults to 'asc' if not specified.
+ *
+ * @returns {T[]} The sorted array in the specified order.
+ */
+export const sort = <T extends string | number | boolean>(
```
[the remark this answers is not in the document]
CarlosGamero: I am not really sure, to be honest, I am adding this method because we are using them, as far as I know, on tests and we don't have a big amount of data there so maybe we can postpone that perf improvement but not sure if you have in mind another cases

packages/app/universal-ts-utils/src/array/sortByField.ts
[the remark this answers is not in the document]
CarlosGamero: Done 🙏

packages/app/universal-ts-utils/src/object/areEquals.ts
CarlosGamero: So, what should we do with this? 😓🙏

packages/app/universal-ts-utils/src/array/sort.ts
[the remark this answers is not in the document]
kibertoad: ok, makes sense then!

packages/app/universal-ts-utils/src/object/areEquals.ts
ondrejsevcik: If it is doing deep equal, then I'm in favor of being more explicit and naming the function "isDeepEqual". It makes it much more obvious for the reader. And knowing whether the function is doing shallow/deep equal is essential.

CarlosGamero: Makes total sense to me, I am going to rename it and maybe we can implement the `isStrictEqual` in the future if we see it is needed 🙏

packages/app/universal-ts-utils/src/array/sortByField.ts
CarlosGamero: After all discussions looks like this is not going to be needed for now, so postponing it

PR #353, opened 2024-10-16
Updating to node-core 13

CarlosGamero:
## Changes

Updating to node-core 13

## Checklist

### DIFF
--- a/packages/app/backend-http-client/package.json
+++ b/packages/app/backend-http-client/package.json
@@ -32,7 +32,7 @@
         "postversion": "biome check --write package.json"
     },
     "dependencies": {
-        "@lokalise/node-core": "^13.0.1",
+        "@lokalise/node-core": "^13.0.0",
         "undici": "^6.19.8",
         "undici-retry": "^5.0.3",
         "zod": "^3.23.8"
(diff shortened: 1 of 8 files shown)


packages/app/backend-http-client/package.json
```diff
@@ -32,7 +32,7 @@
         "postversion": "biome check --write package.json"
     },
     "dependencies": {
-        "@lokalise/node-core": "^13.0.1",
+        "@lokalise/node-core": "^13.0.0",
```
CarlosGamero, line 35: With this, the library will support `13.X.X` instead of `13.0.X`

packages/app/error-utils/package.json
```diff
@@ -25,12 +25,12 @@
         "@bugsnag/js": "^8.0.0"
     },
     "peerDependencies": {
-        "@lokalise/node-core": ">=11.0.0"
+        "@lokalise/node-core": ">=13.0.0"
```
kibertoad, line 28: do we need to narrow it down?

CarlosGamero: Same as with fastify-extras 😅, will revert it

CarlosGamero: Change reverted

dariacm (approved):

kibertoad (approved):

PR #354, opened 2024-10-17
feat(AP-5236): Introduce metrics utils app package

Drodevbar:
## Changes

Introduces the `metrics-utils` package to make working with Prom metrics easier and more standardized.

## Checklist

### DIFF
--- a/packages/app/metrics-utils/README.md
+++ b/packages/app/metrics-utils/README.md
@@ -1,0 +1,49 @@
+# Metrics Utils
+
+This package contains set of utilities to work with Prometheus metrics.
+
+Table of contents:
+1. [AbstractCounterMetric](#AbstractCounterMetric)
+
+### AbstractCounterMetric
+
+A base class for single counter-like metrics.
+
+Usage:
+
+```typescript
+export class PizzaDeliveryCountMetric extends AbstractCounterMetric<
+	'status',
+	['delivered_to_customer', 'delivered_to_pickup_point', 'not_delivered']
+> {
+	constructor({ appMetrics }: Deps) {
+		super(
+			{
+				name: 'pizza_delivery_count',
+				helpDescription: 'Number of pizza deliveries per status',
+				label: 'status',
+				measurementKeys: ['delivered_to_customer', 'delivered_to_pickup_point', 'not_delivered'],
+			},
+			appMetrics,
+		)
+	}
+}
+
+const pizzaDeliveryCountMetric = new PizzaDeliveryCountMetric({ appMetrics })
+
+pizzaDeliveryCountMetric.registerMeasurement({
+    'delivered_to_customer': 1,
+    'delivered_to_pickup_point': 2,
+})
+```
+
+Where:
+1. `name` - metric name
+2. `helpDescription` - metric description
+3. `label` - metric label name
+4. `measurementKeys` - metric label values
+
+Creation of `PizzaDeliveryCountMetric` object initializes a `pizza_delivery_count` metric with `status` label and all the provided measurement keys
+(`delivered_to_customer`, `delivered_to_pickup_point`, `not_delivered`) as 0 values.
+
+`registerMeasurement` method allows to set the specific measurement values of the metric.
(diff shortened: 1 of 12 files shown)


packages/app/metrics-utils/package.json
```diff
@@ -0,0 +1,50 @@
...
+            "require": "./dist/index.cjs",
+            "import": "./dist/index.js"
+        }
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "scripts": {
+        "build": "rimraf dist && vite build",
+        "dev": "vite watch",
```
kibertoad, line 27: this is not needed

packages/app/metrics-utils/package.json
```diff
@@ -0,0 +1,50 @@
...
+        "@lokalise/fastify-extras": "^24.3.1"
+    },
+    "devDependencies": {
+        "@biomejs/biome": "^1.9.2",
+        "@lokalise/biome-config": "^1.5.0",
+        "@lokalise/package-vite-config": "3.1.0",
+        "@vitest/coverage-v8": "^2.1.1",
+        "rimraf": "^6.0.1",
+        "typescript": "5.6.2",
+        "vitest": "^2.0.3"
```
kibertoad, line 48: let's use latest vitest

packages/app/metrics-utils/src/AbstractCounterMetric.ts
```diff
@@ -0,0 +1,87 @@
...
+  helpDescription: string
+  label: TMetricLabel
+  measurementKeys: TMetricMeasurementKeys
+}
+
+export abstract class AbstractCounterMetric<
+  TMetricLabel extends string,
+  TMetricMeasurementKeys extends string[],
+> {
+  private readonly counterMetric: Counter<TMetricLabel> | null
```
kibertoad, line 19: can you add comment explaining it's null when metrics are disabled on app level?

packages/app/metrics-utils/package.json
```diff
@@ -0,0 +1,50 @@
+{
+    "name": "@lokalise/metrics-utils",
```
kibertoad, line 2: can you also add readme, explaining how to use and what it does?

packages/app/metrics-utils/package.json
Drodevbar: Adjusted to the format of other packages across this repo ✅

packages/app/metrics-utils/package.json
Drodevbar: Updated `vitest` and `vite` to latest stable versions ✅

packages/app/metrics-utils/src/AbstractCounterMetric.ts
Drodevbar: Comment added ✅

packages/app/metrics-utils/package.json
Drodevbar: I enhanced the readme with usage instructions. Could you please see if it's clearer now from your point of view?

packages/app/metrics-utils/README.md
```diff
@@ -0,0 +1,49 @@
+# Metrics Utils
+
+This package contains set of utilities to work with Prometheus metrics.
+
+Table of contents:
+1. [AbstractCounterMetric](#AbstractCounterMetric)
+
+### AbstractCounterMetric
+
+A base class for single counter-like metrics.
```
kibertoad, line 10: can you add short explanation why you may prefer this over raw Prometheus metrics? e. g. safe multiple inits

Drodevbar: Sure, added an extra explanation there

kibertoad (approved):
Thank you, this is awesome!

Drodevbar:
@kibertoad It looks like the addition of a new package somehow conflicts with deps of another one (`context-fastify-plugins`) - https://github.com/lokalise/shared-ts-libs/actions/runs/11382381650/job/31665741626 I'll investigate. Sharing with you now in case you have some bright ideas how to solve it.

kibertoad:
@Drodevbar probably not strictly enough pinned deps are messing things up. I can check in an hour or so

Drodevbar:
@kibertoad I see it's fastify-metrics which requires fastify bo be >= 5 (as part of `metrics-utils`), while `context-fastify-plugins` only works with fastify v4. Do you think we should update fastify-plugins to work with v5 of fastify or use package aliases, workspaces, etc.?

PR #373, opened 2024-10-29
AP-5778 introduce prisma metrics plugin

Dosexe:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/prisma-utils/src/plugins/MetricsCollector.ts
+++ b/packages/app/prisma-utils/src/plugins/MetricsCollector.ts
@@ -30,7 +30,7 @@ function getMetrics(
     metrics.counters[metric.key] = new prometheus.Counter({
       name: metric.key,
       help: metric.description,
-      labelNames: ['prisma', 'connection-pool'] as const,
+      labelNames: ['prisma', 'connection_pool'] as const,
     })
     metrics.names.push(metric.key)
   }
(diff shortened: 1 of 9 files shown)


packages/app/prisma-utils/src/plugins/MetricsCollector.ts
```diff
@@ -0,0 +1,170 @@
...
+  counters: Record<string, prometheus.MetricObject>
+  gauges: Record<string, prometheus.MetricObject>
+  histograms: Record<string, prometheus.MetricObject>
+}
+
+export type MetricCollectorOptions = {
+  metricsPrefix: string
+}
+
+function registerMetrics(_prefix: string, jsonMetrics: Metrics): PrometheusMetricsDefinitions {
```
Dosexe, line 23: _prefix is provided but not used, since all already prefix with `prisma_`, but I would leave the value in case we want to use it in future

packages/app/prisma-utils/src/plugins/MetricsCollector.ts
```diff
@@ -0,0 +1,170 @@
...
+    counters: {},
+    gauges: {},
+    histograms: {},
+    names: [],
+  }
+  for (const metric of jsonMetrics.counters) {
+    metrics.counters[metric.key] = new prometheus.Counter({
+      name: metric.key,
+      help: metric.description,
+      labelNames: ['prisma', 'connection_pool'] as const,
```
Dosexe, line 34: prisma has its own labels, we can use them instead + add our prisma label, not all metrics are related to connection pool

CarlosGamero (approved):

PR #390, opened 2024-11-25
Adjust code to support newer bullmq

kibertoad:
## Changes

There were some changes around generics in newer BullMQ, this makes our code compatible with it.

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
@@ -17,7 +17,7 @@ import { DEFAULT_QUEUE_OPTIONS, DEFAULT_WORKER_OPTIONS } from '../constants'
 import type { AbstractBullmqFactory } from '../factories/AbstractBullmqFactory'
 import { BackgroundJobProcessorSpy } from '../spy/BackgroundJobProcessorSpy'
 import type { BackgroundJobProcessorSpyInterface } from '../spy/types'
-import type { BaseJobPayload, BullmqProcessor, RequestContext, SafeJob, SafeQueue } from '../types'
+import type { BaseJobPayload, BullmqProcessor, RequestContext, SafeJob } from '../types'
 import {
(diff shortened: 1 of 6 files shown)


packages/app/background-jobs-common/src/background-job-processor/types.ts
```diff
@@ -20,28 +20,6 @@ export type SafeJob<T = any, R = any, N extends string = string> = Omit<
   waitUntilFinished(queueEvents: any, ttl?: number): Promise<R>
 }
 
-export type SafeQueue<
```
kibertoad, line 23: this is no longer necessary, BullMQ changes finally make types between pro and free versions more compatible out-of-the-box

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -264,8 +264,10 @@ export abstract class AbstractBackgroundJobProcessor<
     const jobs =
       (await this._queue?.addBulk(
         jobData.map((data) => ({
-          name: this.config.queueId,
-          data,
+          // biome-ignore lint/suspicious/noExplicitAny: bull expects ExtractNameType<JobPayload>, but ExtractNameType is not exposed
+          name: this.config.queueId as any,
```
kibertoad, line 268: I hope there is a better solution for this, I would be grateful if someone could look deeper into how we can make TS happy here

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: I will try to have a look later 🙏 But if you can't find a better solution I doubt I will be able to do it 😅

kibertoad: You are much better with TypeScript type wizardry than I am, so there is definitely hope :D

PR #393, opened 2024-11-26
Vault env var sync indicates that 403 might be due to invalid path.

kamilwylegala:
## Changes

Minor change, explained in error log that 403 might be reported due to invalid secrets path.

## Checklist

### DIFF
--- a/packages/app/script-utils/src/vault/vault.ts
+++ b/packages/app/script-utils/src/vault/vault.ts
@@ -100,7 +100,9 @@ export const vaultGetVars = (service: string): Record<string, string> => {
     return responseJson.data.data
   } catch (e: unknown) {
     if (isExecError(e)) {
-      globalLogger.error(`Vault ${service} error downloading env vars -> ${e.stderr.toString()}`)
+      globalLogger.error(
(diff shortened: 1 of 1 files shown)


packages/app/script-utils/src/vault/vault.ts
```diff
@@ -100,7 +100,9 @@ export const vaultGetVars = (service: string): Record<string, string> => {
     return responseJson.data.data
   } catch (e: unknown) {
     if (isExecError(e)) {
-      globalLogger.error(`Vault ${service} error downloading env vars -> ${e.stderr.toString()}`)
+      globalLogger.error(
+        `Vault ${service} error downloading env vars -> ${e.stderr.toString()}. Make sure you're using correct path to secrets. Vault reports 403 if the path is incorrect.`,
```
kibertoad, line 104: can't we do simple "file exists" check to throw an error if file does not exist?

kamilwylegala: But this is response from vault cli, it reports 403 always if secrets path doesn't exist. Did you mean file exist check for `.env` file?

kibertoad (approved):

packages/app/script-utils/src/vault/vault.ts
kibertoad: I was confused about what kind of paths we talk about, all good :D

PR #401, opened 2024-12-05
Migrate to undici 7

kibertoad:
## Changes

This uses newer version of undici-retry that depends on undici v7

## Checklist


packages/app/backend-http-client/src/client/httpClient.ts
```diff
@@ -93,7 +93,7 @@ export async function sendDelete<
       reset: options.disableKeepAlive ?? false,
       bodyTimeout: Object.hasOwn(options, 'timeout') ? options.timeout : DEFAULT_OPTIONS.timeout,
       headersTimeout: Object.hasOwn(options, 'timeout') ? options.timeout : DEFAULT_OPTIONS.timeout,
-      throwOnError: false,
+      throwOnError: undefined,
```
kibertoad, line 96: undici no longer supports this as a parameter and expects https://undici.nodejs.org/#/docs/api/Dispatcher?id=responseerror to be used instead.

Since we always consumed this parameter directly instead of relying on undici handling (in order to provide sufficient amount of error details), we just ensure that we no longer pass-through this parameter.

CarlosGamero (approved):
🚀

PR #416, opened 2024-12-19
Trim text util

CarlosGamero:
## Changes

Adding utility to trim a text but contemplating HTML space and in addition can return the trimmed part in case it is needed

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/src/node.ts
+++ b/packages/app/universal-ts-utils/src/node.ts
@@ -23,6 +23,9 @@ export * from './public/object/isEmpty.js'
 export * from './public/object/pick.js'
 export * from './public/object/transformToKebabCase.js'
 
+// string
+export * from './public/string/trimText.js'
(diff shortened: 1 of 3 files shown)


packages/app/universal-ts-utils/src/node.ts
```diff
@@ -23,6 +23,9 @@ export * from './public/object/isEmpty.js'
 export * from './public/object/pick.js'
 export * from './public/object/transformToKebabCase.js'
 
+// string
```
kibertoad, line 26: can you also add readme entry for it?

CarlosGamero: Discussed on Slack, we are not documenting methods right now but we should do it so I will address it in a separate task -> https://lokalise.atlassian.net/browse/DXP-335

kibertoad (approved):

PR #426, opened 2024-12-23
Remove dotenv

kibertoad:
## Changes

This removes the unnecessary dependency.

## Checklist


packages/app/prisma-utils/test/envSetupHook.ts
```diff
@@ -1 +1 @@
-process.loadEnvFile('./.env.test')
+process.loadEnvFile('./.env')
```
kibertoad, line 1: generally it's not recommended to commit your .env file, but we already were using it both for tests and for dev, so technically it's not an .env.test

And if it's called `.env`, prisma is able to pick it up directly (no, there is no config param, unfortunately).

We can also rename .env to .env.default and add the copy command instead.

CarlosGamero (approved):

PR #443, opened 2025-01-27
improve exports config for universal ts utils

CatchMe2:
## Changes

This PR improves the exports config in universal-ts-utils package.json.
After the changes, adding explicit extensions when importing an util will be forbidden, and the import will be automatically resolved to the js file.
Thanks to this change IDE will no longer list .dts or source map files in imports autocomplete
image

This is a MAJOR change, as every `*.js` import will have to be replaced with an extensionless one
```
import { mapNonEmptyArray } from '@lokalise/universal-ts-utils/array/mapNonEmptyArray.js'
into
import { mapNonEmptyArray } from '@lokalise/universal-ts-utils/array/mapNonEmptyArray'
```

For more info, please take a look into: https://hirok.io/posts/package-json-exports#subpaths-resolution-advanced

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/package.json
+++ b/packages/app/universal-ts-utils/package.json
@@ -39,15 +39,8 @@
         "vitest": "^2.1.3"
     },
     "exports": {
-        "./node": {
-            "types": "./dist/node.d.ts",
-            "import": "./dist/node.js"
-        },
-        "./*": {
-            "types": "./dist/public/*.d.ts",
-            "import": "./dist/public/*",
-            "require": "./dist/public/*"
-        },
+        "./node": "./dist/node.js",
+        "./*": "./dist/public/*.js",
         "./package.json": "./package.json"
     }
 }
--- a/packages/app/universal-ts-utils/README.md
+++ b/packages/app/universal-ts-utils/README.md
@@ -23,7 +23,7 @@ entry point.
 
 **Frontend:**
 ```typescript
-import { chunk } from '@lokalise/universal-ts-utils/array/chunk.js';
+import { chunk } from '@lokalise/universal-ts-utils/array/chunk';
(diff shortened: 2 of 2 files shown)


packages/app/universal-ts-utils/package.json
```diff
@@ -39,15 +39,8 @@
...
-            "types": "./dist/node.d.ts",
-            "import": "./dist/node.js"
-        },
-        "./*": {
-            "types": "./dist/public/*.d.ts",
-            "import": "./dist/public/*",
-            "require": "./dist/public/*"
-        },
+        "./node": "./dist/node.js",
+        "./*": "./dist/public/*.js",
```
kibertoad, line 43: how does it know where to look for types this way?

CatchMe2: If types are not specified explicitly, they are picked automatically based on the provided source file path.
Please take a look at https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html#editing-the-packagejson
image
The same behavior for the `main` field is applied to `exports`.

In general, there is no point to explicitly define `types` unless they are not colocated with the source file.

kibertoad: we don't use either "main" or index.d.ts in the root. are you sure this resolution will work?

CatchMe2: I've tested it in polyglot and it works.
When importing a file with a `.js` extension, the error is reported during `tsc`.
image

After removing the extension, all works as expected (no issues during `tsc`)

kibertoad (approved):

kibertoad:
@CatchMe2 tried it out locally, and it's not working for me. let's huddle tomorrow

PR #444, opened 2025-01-27
Use correct extension for components

kibertoad:
## Changes

This makes TSX files resolve to JSX and not JS

## Checklist

### DIFF
--- a/packages/dev/biome-config/package.json
+++ b/packages/dev/biome-config/package.json
@@ -2,9 +2,7 @@
     "name": "@lokalise/biome-config",
     "version": "1.6.0",
     "license": "Apache-2.0",
-    "files": [
-        "configs/*"
-    ],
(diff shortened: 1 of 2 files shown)


CatchMe2 (approved):

packages/dev/biome-config/package.json
```diff
@@ -16,9 +14,7 @@
         "url": "git://github.com/lokalise/shared-ts-libs.git"
     },
     "exports": {
-        "./*": [
-            "./configs/*.jsonc"
-        ]
+        "./*": ["./configs/*.jsonc"]
```
CatchMe2, line 17: `exports` value can be a plain string instead of an array of strings

kibertoad: fair, but it's more flexible this way

PR #445, opened 2025-01-29
Implement API contracts

kibertoad:
## Changes

See https://github.com/lokalise/frontend-http-client/pull/40 for the original discussion

## Checklist


packages/app/universal-ts-utils/README.md
```diff
@@ -382,6 +382,36 @@ const a = isStandardizedError({ code: 'code', message: 'test' }) // True
...
+
+Usage examples:
+
+```ts
+    import {buildGetRoute, buildDeleteRoute, buildDeleteRoute} from '@lokalise/universal-ts-utils'
+    const getContract = buildGetRoute({
+    responseBodySchema: BODY_SCHEMA,
+    requestPathParamsSchema: PATH_PARAMS_SCHEMA,
+    pathResolver: (pathParams) => `/users/${pathParams.userId}`,
+})
```
CarlosGamero, line 399:
```suggestion
import {buildGetRoute, buildDeleteRoute, buildDeleteRoute} from '@lokalise/universal-ts-utils'
const getContract = buildGetRoute({
    responseBodySchema: BODY_SCHEMA,
    requestPathParamsSchema: PATH_PARAMS_SCHEMA,
    pathResolver: (pathParams) => `/users/${pathParams.userId}`,
})
```

packages/app/universal-ts-utils/src/public/api-contracts/apiContracts.ts
```diff
@@ -0,0 +1,219 @@
...
+  isNonJSONResponseExpected?: IsNonJSONResponseExpected
+  isEmptyResponseExpected?: IsEmptyResponseExpected
+  responseBodySchema: ResponseBodySchema
+  requestPathParamsSchema?: PathParamsSchema
+  requestQuerySchema?: RequestQuerySchema
+  requestHeaderSchema?: RequestHeaderSchema
+  pathResolver: RoutePathResolver<InferSchemaOutput<PathParamsSchema>>
+}
+
+export type ChangeRouteDefinition<
```
CarlosGamero, line 29: 🟡 I am not sure about this change route sometimes, we use a POST request as a GET to be able to pass a large amount of "filters".

kibertoad: How would you call it? technically POST, PUT and PATCH are all supposed to be mutation verbs, although yeah, it doesn't always work that way, we use actions pattern to "create" actions

CarlosGamero: I don't have any good idea tbh 😓, the only thing that comes to my mind is splitting and having a method/definition per HTTP method.

I don't have a strong opinion on this, if you think that "change" is a good name I am also fine with it

CarlosGamero (approved):

packages/app/universal-ts-utils/src/public/api-contracts/apiContracts.ts
kibertoad: discussed offline, went for PayloadRoute

szymonchudy (approved):

PR #448, opened 2025-01-31
feature: introduce new param to improve discoverability of autorun opt

dariacm:
## Changes

There are different scenarios where we would like to schedule or unschedule a BullMQ repeatable job through a CLI command. While it is definitely possible to set up the app to do so, it might happen that:

- Other jobs might be processed by the CLI instance while it is up, rather than the production application;
- In case of scheduling a job, it might be executed by the CLI instance rather than the production application.

The solution I am proposing leverages the worker option `autorun` in BullMQ, which allows to choose when to manually start the worker:
image
In case of CLI commands, we can start the queue so that jobs can be scheduled and/or unscheduled, but never start the worker, so that jobs will not be picked up by this instance for processing.

To improve discoverability of this option, I am introducing an ad-hoc parameter in the constructor to use it.

Additional context: https://github.com/lokalise/maestro/pull/526

## Checklist


CarlosGamero (approved):
🚀

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
```diff
@@ -159,6 +159,34 @@ describe('AbstractBackgroundJobProcessor', () => {
...
       await processor.dispose()
     })
+
+    it('processors starts queue but not worker if workerAutoRunEnabled is true', async () => {
+      const processor = new FakeBackgroundJobProcessor<JobData>(
+        deps,
+        randomUUID(),
+        mocks.getRedisConfig(),
+        true,
+        false,
```
kibertoad, line 169: not for this PR, but I wonder if we should switch to a object param, it has lots of arguments already, and two booleans in a row are confusing and error-prone

kibertoad (approved):

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
dariacm: You mean for the fake processor?

kibertoad: yeah, hence it's a pretty low priority, as it's not a prod code

dariacm: https://lokalise.atlassian.net/browse/DXP-342 

I can do it but I created a ticket just in case I lose it in the pile of things to do 😄

PR #456, opened 2025-02-07
Feature/queue manager pro version support

CarlosGamero:
## Checklist


packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
```diff
@@ -1,46 +1,77 @@
...
   ) {
-    this.queueIds = new Set<string>()
-    this.jobRegistry = jobRegistry
-
-    for (const queue of queues) {
-      this.queueIds.add(queue.queueId)
-      this.queueMap[queue.queueId] = queue
-    }
+    this.factory = factory
+    this.queueRegistry = new QueueRegistry(queues)
```
CarlosGamero, line 74: Merging `QueueConfiguration` and `JobDefinition`, they both need to be in sync in order to have a valid setup, so it makes sense to merge them and only have to define a single object.

As a consequence of the prev, we no longer need a job registry and with an internal helper class (QueueRegistry) is enough.
Please let me know if you disagree 🙏

kibertoad: oh wow, that is cool! and type inferrence still works fine?

CarlosGamero: Not really, I am on it 😅

CarlosGamero: it seems to be working now 🚀

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.spec.ts
```diff
@@ -166,7 +167,7 @@ describe('QueueManager', () => {
       await expect(
         queueManager.schedule(
           'queue1',
-          // @ts-expect-error Should expect mandatory fields from queue1 schema
+          // @ts-ignore Should expect mandatory fields from queue1 schema
```
kibertoad, line 170: why change from ts-expect-error?

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.spec.ts
```diff
@@ -227,7 +228,7 @@ describe('QueueManager', () => {
 
       await expect(
         queueManager.scheduleBulk('queue1', [
-          // @ts-expect-error Should expect mandatory fields from queue1 schema
+          // @ts-ignore Should expect mandatory fields from queue1 schema
```
kibertoad, line 231: ts-expect-error is better here, as if there is no error, type engorcement is not working as expected

CarlosGamero: Yeah, I am not very happy with that, the problem is that for some reason IDE is not detecting the type issue and it is marking ts-expect-error as an error 🤦, but if you remove it and execute lint command the type problem is there.

I replaced it with ts-ignore to make IDE and linter happy, but if you know a better way to achieve it I am more than happy to change it

kibertoad: is this a problem only with test files, or it manifests in runtime code as well?

CarlosGamero: Only with test files in the IDE

kibertoad: I would say it's fine then, and we should keep it as ts-expect-error.

Btw, what if we make a default tsconfig cover everything, and use a separate one for building the lib? (basically inverse the configs). Wonder if that will help the ide

CarlosGamero: Ok! I will change it, and good idea, will try to test it in a separate branch

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.spec.ts
CarlosGamero: Discussed on next comment https://github.com/lokalise/shared-ts-libs/pull/456#discussion_r1947550194

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.spec.ts
CarlosGamero: Adjusted :D

CarlosGamero:
@kibertoad If it is fine, I will address the change we discussed about having the queue available on the processor with proper typing in a follow-up PR, just to no make this one even bigger 😓

PR #457, opened 2025-02-09
Feature/bg split queue on processor

CarlosGamero:
## Changes

Please describe

## Checklist


packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
```diff
@@ -24,7 +27,6 @@ export type BackgroundJobProcessorConfigNew<
   redisConfig: RedisConfig
   barrier?: BarrierCallback<JobPayload, ExecutionContext, JobReturn, JobType>
   lazyInitEnabled?: boolean
-  workerAutoRunEnabled?: boolean
```
CarlosGamero, line 27: I think this is no longer needed, we will schedule using the queue manager and not the processor

packages/app/background-jobs-common/src/background-job-processor/barrier/JobQueueSizeThrottlingBarrier.spec.ts
```diff
@@ -18,8 +18,7 @@ describe('JobQueueSizeThrottlingBarrier', () => {
   beforeEach(async () => {
     mocks = new DependencyMocks()
     deps = mocks.create()
-    mocks.startRedis()
-    await mocks.clear()
+    await mocks.clearRedis()
```
kibertoad, line 21: this name is very misleading, we use a real redis and not a mock there. wonder if we should either rename mocks to testDependencies, or move the redis out of it

packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
kibertoad: yeah, I agree

packages/app/background-jobs-common/src/background-job-processor/barrier/JobQueueSizeThrottlingBarrier.spec.ts
CarlosGamero: I absolutely agree with you, I will take it with the TS config improvement we discussed

packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
CarlosGamero: I also removed lazyInitEnabled, it was not used

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    // Then
+    const UUID_REGEX =
+      /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/
+    expect(UUID_REGEX.test(jobId)).toBe(true)
+
+    const job = await simpleProcessor.spy.waitForJobWithId(jobId, 'completed')
+    expect(job.data).toMatchObject(jobData)
+
+    const resolvedJob = await queueManager.getQueue('queue1').getJob(job.id!)
```
kibertoad, line 91: is something wrong with the type inference here? isn't `id` a mandatory field?

CarlosGamero: Not sure why it was not complaining before, the job.id property is string | undefined on BullMq
```
export declare class Job<DataType = any, ReturnType = any, NameType extends string = string> implements MinimalJob<DataType, ReturnType, NameType> {
    protected queue: MinimalQueue;
    /**
     * The name of the Job
     */
    name: NameType;
    /**
     * The payload for this job.
     */
    data: DataType;
    /**
     * The options object for this job.
     */
    opts: JobsOptions;
    id?: string;
```

kibertoad: gotcha, thanks!

kibertoad (approved):
incredible job as always, Carlos!

PR #466, opened 2025-02-11
Unified start() method signature to match existing approaches

ForterLokalise:
## Changes

We've changed the signature of start() method of QueueManager by adding an optional list of queues to start. Because of that, when you awilix-manager tries to initialise QueueManager and passes Dependencies as a default first argument, everything breaks. So we need a method without arguments for default initialization.

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
@@ -85,7 +85,15 @@ export class QueueManager<
     return this._queues[queueId]
   }
 
-  public async start(queueIdsToStart?: string[]): Promise<void> {
+  public async start(): Promise<void> {
+    if (this.isStarted) return // if it is already started -> skip
+
+    if (!this.startPromise) this.startPromise = this.internalStart()
+    await this.startPromise
+    this.startPromise = undefined
+  }
+
+  public async startSpecifiedQueues(queueIdsToStart?: string[]): Promise<void> {
     if (this.isStarted) return // if it is already started -> skip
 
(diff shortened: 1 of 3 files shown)


packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
```diff
@@ -85,7 +85,15 @@ export class QueueManager<
...
-  public async start(queueIdsToStart?: string[]): Promise<void> {
+  public async start(): Promise<void> {
+    if (this.isStarted) return // if it is already started -> skip
+
+    if (!this.startPromise) this.startPromise = this.internalStart()
+    await this.startPromise
+    this.startPromise = undefined
+  }
+
+  public async startSpecifiedQueues(queueIdsToStart?: string[]): Promise<void> {
```
CarlosGamero, line 96: 🟡 if I understand the PR, we will use the start method cromo Awilix asyncStart right? 

How would we start a single queue on tests? (I mean, the jobsEnabled param)

Wondering if we could use a single method, and if jobsEnabled is true, start all queues

kibertoad: yeah, I share the concern. we need to figure out a way to preserve granular execution in our setup. we may need some changes on awilix-manager side

ForterLokalise: The issue was resolved from awilix-manager side: https://github.com/kibertoad/awilix-manager/pull/44, so I'm closing this. Thanks folks! 🙏🏻

PR #467, opened 2025-02-12
fix: Update QueueManager.ts start args

andrewi-wd:
## Changes

Replace undefined arg with a boolean for QueueManager to start

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
@@ -85,6 +85,10 @@ export class QueueManager<
     return this._queues[queueId]
   }
 
+  /**
+   * Start the queues
+   * @param enabled default true - if true, start all queues, if false, do nothing, if array, start only the queues in the array (array of queue IDs expected)
+   */
   public async start(enabled: string[] | boolean = true): Promise<void> {
     if (this.isStarted) return // if it is already started -> skip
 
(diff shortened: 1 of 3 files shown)


packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
```diff
@@ -85,10 +85,12 @@ export class QueueManager<
     return this._queues[queueId]
   }
 
-  public async start(queueIdsToStart?: string[]): Promise<void> {
+  public async start(enabled: string[] | boolean = true): Promise<void> {
```
CarlosGamero, line 88: Could we add some tests to ensure we don't break the behaviour in the future? 🙏

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
CarlosGamero: Oh, sorry, tests are already there ... my bad

kibertoad, line 88: do we need to update anything in readme?

andrewi-wd: Added a test for a case when `false` value is provided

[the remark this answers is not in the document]
andrewi-wd: Added https://github.com/lokalise/shared-ts-libs/pull/467/commits/c8e615787c41c128bafa68434cd8f4b040693950

andrewi-wd: Current readme does not dive deep and has no details on different `start()` options. But all code examples are up to date, using no arg which defaults to `enabled = true` 👍

kibertoad (approved):

PR #483, opened 2025-02-19
Support status code responses

kibertoad:
## Changes

Can be used for generating better OpenAPI docs

## Checklist


packages/app/universal-ts-utils/src/public/api-contracts/apiContracts.ts
```diff
@@ -21,26 +86,27 @@ export type CommonRouteDefinition<
 > = {
   isNonJSONResponseExpected?: IsNonJSONResponseExpected
   isEmptyResponseExpected?: IsEmptyResponseExpected
-  responseBodySchema: ResponseBodySchema
+  successResponseBodySchema: ResponseBodySchema
```
CarlosGamero, line 89: Mmm, not sure about this param. Maybe what I am going to say is a crazy idea so please feel free to disagree.

I guess we are using `successResponseBodySchema` to parse response on HTTP clients, but can't we use `[200, 300)` range for it? and maybe use the status code to find the right schema in `responseSchemasByStatusCode`

kibertoad: discussed offline, this will break type inference for HTTP clients. we'll investigate in a separate PR if some black TS wizardry could still work here

PR #486, opened 2025-02-19
Trying to fix issue by using default exports

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/fastify-api-contracts/package.json
+++ b/packages/app/fastify-api-contracts/package.json
@@ -12,13 +12,6 @@
         "type": "git",
         "url": "git://github.com/lokalise/shared-ts-libs.git"
     },
-    "exports": {
-        ".": {
-            "types": "./dist/index.d.ts",
-            "require": "./dist/index.cjs",
-            "import": "./dist/index.js"
-        }
-    },
     "publishConfig": {
         "access": "public"
     },


packages/app/fastify-api-contracts/package.json
```diff
@@ -12,13 +12,6 @@
         "type": "git",
         "url": "git://github.com/lokalise/shared-ts-libs.git"
     },
-    "exports": {
-        ".": {
-            "types": "./dist/index.d.ts",
-            "require": "./dist/index.cjs",
-            "import": "./dist/index.js"
-        }
-    },
```
CarlosGamero, line 21: Screenshot 2025-02-19 at 17 26 21

This is the issue we are trying to fix.

I tested locally by linking lib to project and removing explicit `exports` seems to fix the issue. However, I have no idea why it can cause the problems or if after releasing the issue will still be there but worth trying, but in any case trusting on default ones is a good approach I guess what do you think?

kibertoad: let's try it!

kibertoad: how does it know where to get compiled sources from, though?

kibertoad (approved):

packages/app/fastify-api-contracts/package.json
CarlosGamero: I think it is applying default values, I guess index is exported by default

PR #491, opened 2025-02-20
Contract metadata mapper

CarlosGamero:
## Checklist


packages/app/fastify-api-contracts/src/fastifyApiContracts.ts
```diff
@@ -132,8 +87,10 @@ export function buildFastifyNoPayloadRoute<
     RequestQuerySchema,
     RequestHeaderSchema
   >,
+  apiContractToRouteMapper: ApiContractMetadataToRouteMapper = () => ({}),
```
kibertoad, line 90: I would suggest calling it `contractMetadataToRouteMapper` to make it less ambiguous

packages/app/fastify-api-contracts/src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    z.Schema | undefined,
+    z.Schema | undefined,
+    z.Schema | undefined,
+    z.Schema | undefined,
+    boolean,
+    boolean
+  >,
+>(
+  metadata: ApiContract['metadata'],
+) => Omit<RouteType, 'handler' | 'method' | 'schema' | 'url'>
```
kibertoad, line 71: do we want to make it more generic and allow override any route fields, not just the config?

packages/app/fastify-api-contracts/src/fastifyApiContracts.ts
CarlosGamero: Good catch! thank you 🙏

packages/app/fastify-api-contracts/src/types.ts
CarlosGamero: My idea was to allow users to use metadata to override any route field except the ones that are being filled already with the contract, just to be more flexible, but I have no strong opinion, we can start by `config` (using `Pick`) and extend in the future if needed. Actually, it sounds like a safer approach, What do you think?

CarlosGamero: Discussed offline, for visibility, we are going to support all hooks + config + body limit.

Already adjusted, thank you Igor 🙏

kibertoad (approved):

PR #493, opened 2025-02-21
Type inference improvement

kjamrog:
## Changes

This change allows to properly infer response type in case preprocessing on response is used. To achieve that we need to use different type inference for input and output values

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/src/public/api-contracts/apiContracts.ts
+++ b/packages/app/universal-ts-utils/src/public/api-contracts/apiContracts.ts
@@ -34,7 +34,7 @@ export type CommonRouteDefinition<
   requestPathParamsSchema?: PathParamsSchema
   requestQuerySchema?: RequestQuerySchema
   requestHeaderSchema?: RequestHeaderSchema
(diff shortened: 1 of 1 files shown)


packages/app/frontend-http-client/tsconfig.json
```diff
@@ -1,15 +1,15 @@
...
+        "module": "ESNext",
         "target": "ES2022",
         "lib": ["ES2022", "dom"],
         "sourceMap": false,
         "declaration": true,
         "declarationMap": false,
         "types": ["node", "vitest/globals"],
         "strict": true,
-        "moduleResolution": "node16",
+        "moduleResolution": "Bundler",
```
CarlosGamero, line 12: Is this needed? I think it is a breaking change, and not sure if we are fine with introducing it rn

kibertoad: does it affect the output of the module itself, not just its internal behaviour?

kjamrog: It's not needed, leftover from debugging why linting wasn't working properly (the issue was different)

CarlosGamero (approved):

PR #500, opened 2025-02-24
feat: support headers

kibertoad:
## Changes

Support passing headers or header factory

## Checklist

### DIFF
--- a/packages/app/frontend-http-client/package.json
+++ b/packages/app/frontend-http-client/package.json
@@ -46,14 +46,14 @@
         "fast-querystring": "^1.1.2"
     },
     "peerDependencies": {
-        "@lokalise/universal-ts-utils": "^4.2.1",
+        "@lokalise/universal-ts-utils": "^4.2.2",
         "wretch": "^2.8.0",
         "zod": "^3.22.0"
     },
     "devDependencies": {
         "@biomejs/biome": "^1.8.3",
         "@lokalise/biome-config": "^1.0.0",
-        "@lokalise/universal-ts-utils": "^4.2.1",
+        "@lokalise/universal-ts-utils": "^4.2.2",
         "@types/node": "^22.0.0",
         "@vitest/coverage-v8": "^3.0.5",
         "jest-fail-on-console": "^3.1.2",
--- a/packages/app/frontend-http-client/src/client.spec.ts
+++ b/packages/app/frontend-http-client/src/client.spec.ts
@@ -19,6 +19,14 @@ import {
   sendPost,
   sendPut,
 } from './client.js'
+import {constants as httpConstants} from "http2";
+
+export const JSON_HEADERS = {
+  'Content-Type': 'application/json',
+}
+const HEADERS_SCHEMA = z.object({
+  authorization: z.string(),
+}).strip()
 
 describe('frontend-http-client', () => {
   const mockServer = getLocal()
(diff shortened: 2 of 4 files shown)


packages/app/frontend-http-client/src/client.ts
```diff
@@ -40,7 +51,8 @@ function sendResourceChange<
     ResponseBody,
     IsNonJSONResponseExpected,
     IsEmptyResponseExpected,
-    RequestQuerySchema
+    RequestQuerySchema,
+    RequestHeaderSchema
```
ntrpilot-lokalise, line 55: 🟢 No trailing commas? 😢

packages/app/frontend-http-client/src/client.ts
```diff
@@ -135,44 +154,51 @@ export function sendGet<
...
-    if (bodyParseResult.error === 'NOT_JSON') {
-      if (params.isNonJSONResponseExpected) {
-        return response
-      }
-      return Promise.reject(
-        buildWretchError(
-          `Request to ${params.path} has returned an unexpected non-JSON response.`,
-          response,
-        ),
+  return wretch
```
ntrpilot-lokalise, line 159: 🟡  Can you not reuse the method from above? This seems to me like unneeded duplication

packages/app/frontend-http-client/src/client.ts
```diff
@@ -135,44 +154,51 @@ export function sendGet<
...
+        params.path,
+        params.responseBodySchema,
+        params.isEmptyResponseExpected,
       )
-    }
 
-    if (bodyParseResult.error === 'EMPTY_RESPONSE') {
-      if (params.isEmptyResponseExpected) {
-        return null
+      if (bodyParseResult.error === 'NOT_JSON') {
```
ntrpilot-lokalise, line 170: 🟡 This is also duplicated - can it be a utility function?

CarlosGamero (approved):

packages/app/frontend-http-client/src/client.ts
```diff
@@ -270,46 +306,53 @@ export function sendDelete<
...
-    if (bodyParseResult.error === 'NOT_JSON') {
-      if (params.isNonJSONResponseExpected === false) {
-        return Promise.reject(
-          buildWretchError(
-            `Request to ${params.path} has returned an unexpected non-JSON response.`,
-            response,
-          ),
-        )
-      }
+  return wretch
```
ntrpilot-lokalise, line 311: 🟡 Again - this seems to duplicate the code above..?

packages/app/frontend-http-client/src/client.spec.ts
```diff
@@ -196,6 +205,90 @@ describe('frontend-http-client', () => {
       })
     })
 
+    it('returns deserialized response for GET with headers', async () => {
```
ntrpilot-lokalise, line 208: 🟢  It's not super obvious what is actually being tested here - maybe you can make it a little more obvious

ntrpilot-lokalise (approved):
I have some comments, but nothing that I would consider a blocker - thus the approval 😄

packages/app/frontend-http-client/src/client.ts
kibertoad: generics do not support trailing commas

packages/app/frontend-http-client/package.json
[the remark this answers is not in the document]
kibertoad: I think it wasn't forcing correct dependency otherwise locally, but generally peerDependency should be fine. I will remove it, and if nothing breaks, keep that way

packages/app/frontend-http-client/src/client.ts
```diff
@@ -25,13 +26,23 @@ import { parseQueryParams } from './utils/queryUtils.js'
...
 
-function sendResourceChange<
+function resolveHeaders(
+  headers:
+    | Record<string, string>
+    | (() => Record<string, string>)
+    | (() => Promise<Record<string, string>>),
+): Record<string, string> | Promise<Record<string, string>> {
+  return (typeof headers === 'function' ? headers() : headers) ?? {}
+}
```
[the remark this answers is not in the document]
kibertoad: addressed it in a somewhat different way, I don't want to mix types of input and output here

packages/app/frontend-http-client/src/client.ts
kibertoad: that was a great catch, and extracting it actually helped fix few bugs along the way!

packages/app/frontend-http-client/src/client.ts
kibertoad: done

packages/app/frontend-http-client/src/client.ts
kibertoad: not quite, if you take a look, there are subtle differences, and making it too conditional will complicate code significantly. I've extracted the part that was easy to reuse now.

packages/app/frontend-http-client/src/client.ts
ntrpilot-lokalise: OH yeah.. Missed that this was a list of generics 😄

packages/app/frontend-http-client/src/client.ts
```diff
@@ -7,34 +7,75 @@ import type {
...
-): Record<string, string> | Promise<Record<string, string>> {
+function resolveHeaders(headers: HeadersSource): HeadersObject | Promise<HeadersObject> {
   return (typeof headers === 'function' ? headers() : headers) ?? {}
 }
 
+function handleBodyParseResult<RequestBodySchema extends z.ZodSchema>(
+  bodyParseResult: Either<
+    'NOT_JSON' | 'EMPTY_RESPONSE' | ZodError<RequestBodySchema>,
+    z.output<RequestBodySchema>
+  >,
```
ntrpilot-lokalise, line 40: 🟢  This feels like a type that should be extracted - correct me if I am wrong, but this should be enforced as the return type of the bodyParser. This works though, just an idea

packages/app/frontend-http-client/src/client.ts
```diff
@@ -7,34 +7,75 @@ import type {
...
+      )
+    }
+
+    return null
+  }
+
+  if (bodyParseResult.error) {
+    return Promise.reject(bodyParseResult.error)
+  }
+}
```
ntrpilot-lokalise, line 77: 🔴 This method had no default return.. I am sure we have/had a lint rule that makes this an error - Perhaps you can return `Promise.resolve(bodyParseResult.result)` so that it can be used unconditionally below.. (Monoids FTW :all-the-things:)

packages/app/frontend-http-client/src/types.ts
```diff
@@ -1,6 +1,8 @@
 import type { Wretch, WretchResponse } from 'wretch'
 import type { ZodSchema, z } from 'zod'
 
+export type HeadersObject = Record<string, string>
+export type HeadersSource = HeadersObject | (() => HeadersObject) | (() => Promise<HeadersObject>)
```
ntrpilot-lokalise, line 5: 🥳

packages/app/frontend-http-client/src/client.ts
kibertoad: you are right. it was possible to actually simplify the code in question

packages/app/frontend-http-client/src/client.ts
kibertoad: done

ntrpilot-lokalise (approved):

PR #503, opened 2025-02-27
adjust NT tags regex

Dosexe:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/non-translatable-markup/src/methods/removeNonTranslatableTags.ts
+++ b/packages/app/non-translatable-markup/src/methods/removeNonTranslatableTags.ts
@@ -4,4 +4,4 @@ import { nonTranslatableTagsRegexpG } from './utils'
  * Removes any non-translatable tag.
  */
 export const removeNonTranslatableTags = (text: string): string =>
-  text.replace(nonTranslatableTagsRegexpG, '')
+  text.replace(nonTranslatableTagsRegexpG, '$1')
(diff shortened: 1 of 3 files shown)


packages/app/non-translatable-markup/src/methods/removeNonTranslatableTags.ts
```diff
@@ -4,4 +4,4 @@ import { nonTranslatableTagsRegexpG } from './utils'
  * Removes any non-translatable tag.
  */
 export const removeNonTranslatableTags = (text: string): string =>
-  text.replace(nonTranslatableTagsRegexpG, '')
+  text.replace(nonTranslatableTagsRegexpG, '$1')
```
Dosexe, line 7: we do not want to remove single tags or empty tags, but rather only if something present in between, singular tags are considered to be valid text, and empty tags is non valid case at all.

packages/app/non-translatable-markup/src/methods/utils.ts
```diff
@@ -4,7 +4,7 @@ const nonTranslatableTextPattern = `${NON_TRANSLATABLE_START_TAG}.*?${NON_TRANSL
 export const nonTranslatableTextRegexp = new RegExp(nonTranslatableTextPattern)
 export const nonTranslatableTextRegexpG = new RegExp(nonTranslatableTextPattern, 'g')
```
CarlosGamero, line 5: Do we also need to modify `nonTranslatableTextPattern` to contemplate new use case?

Dosexe: From looking where it used, we do not I believe, the regex catches changes on anything in between the tags, if by any means there is nothing in between the tags, you will never be able to change it (also it is not a valid content at all).

eduard-fa (approved):
LGTM
Let's wait for @CarlosGamero review.

Drodevbar (approved):

CarlosGamero (approved):

PR #505, opened 2025-02-27
bg processor shared spy instance

CarlosGamero:
## Changes

With this PR, we are sharing the spy instance between the processor and queue manager, which helps to have better visibility on the real state of the job.

Also, we are removing `isTest` and `redisConfig` from the processor config, it will now be taken from the queue manager. This helps to simplify config and avoid configuration issues (eg, having queue and processor with different redis instances)

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
@@ -41,7 +41,8 @@ export class QueueManager<
     QueueConfiguration<QueueOptionsType>['queueId'],
     BackgroundJobProcessorSpy<
       JobPayloadForQueue<Queues, QueueConfiguration<QueueOptionsType>['queueId']>,
-      undefined
+      // biome-ignore lint/suspicious/noExplicitAny: At this point we don't know the return type of the spy
+      any
     >
   > = {}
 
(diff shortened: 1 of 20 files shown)


packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
```diff
@@ -31,17 +31,18 @@ export class QueueManager<
   QueueOptionsType extends QueueOptions = QueueOptions,
   JobOptionsType extends JobsOptions = JobsOptions,
 > {
+  public readonly config: QueueManagerConfig
+
```
CarlosGamero, line 35: Using this on processor, so we avoid misconfigurations with `redisConfig` and `isTest` params

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -134,9 +134,17 @@ export abstract class AbstractBackgroundJobProcessorNew<
     this.queueManager = dependencies.queueManager
     this.errorReporter = dependencies.errorReporter
 
-    this._spy = config.isTest ? new BackgroundJobProcessorSpy() : undefined
+    this._spy = this.queueManager.config.isTest
+      ? (this.queueManager.getSpy<QueueId, JobReturn>(
+          this.config.queueId,
+        ) as BackgroundJobProcessorSpy<JobPayloadForQueue<Queues, QueueId>, JobReturn>)
+      : undefined
```
CarlosGamero, line 141: By sharing the spy instance, we have job state consistency between queue manager and processor

PR #506, opened 2025-03-03
Browser compatible cursor encoding

kjamrog:
## Changes

In the browser, we cannot use Buffer for cursor encoding. Therefore, within this change, `btoa` and `atob` functions will be used when Buffer is not available

## Checklist


kibertoad (commented):
Can you document that these methods support both browser and node envs?

szymonchudy:
@kibertoad we should be fine https://caniuse.com/?search=atob

kibertoad:
@szymonchudy Not saying it wouldn't work, just saying it would be nice to document this method is cross-env-compatible :-D

kibertoad (approved):

PR #512, opened 2025-03-04
package-vite-config vitest 3

CarlosGamero:
## Checklist

### DIFF
--- a/packages/dev/package-vite-config/package.json
+++ b/packages/dev/package-vite-config/package.json
@@ -33,7 +33,7 @@
     "peerDependencies": {
         "vite": "^5.4.2",
         "vite-plugin-dts": "^4.2.1",
-        "vitest": "^2.0.3"
+        "vitest": "^3.0.7"
     },
     "devDependencies": {
         "@biomejs/biome": "^1.9.4",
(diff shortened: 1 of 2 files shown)


dariacm (approved):

packages/dev/package-vite-config/package.json
```diff
@@ -33,7 +33,7 @@
     "peerDependencies": {
         "vite": "^5.4.2",
         "vite-plugin-dts": "^4.2.1",
-        "vitest": "^2.0.3"
+        "vitest": "^3.0.7"
```
kibertoad, line 36: no changes in the config itself needed? I thought it had some different expectations around vitest-related entries

kibertoad (approved):

packages/dev/package-vite-config/package.json
CarlosGamero: I checked [docs](https://vitest.dev/guide/migration#vitest-3), and I think the changes are not affecting this package, but please let me know if I missed something.

Btw, to add a bit more context about why I am migrating to vitest 3, yesterday I tried to migrate to vite 6, and it looks like vitest 3 is needed; there is a version conflict with `vite-node` if we try to use vite 6 with vitest 2. I didn't spend much time investigating it or trying to find a workaround as in any case, migrate to Vitest 3 is beneficial

kibertoad (approved):

PR #523, opened 2025-03-07
chore: add sharable tsconfig package

CatchMe2:
## Changes

This PR adds a sharable tsconfig package. As an example, this package is used for `non-translatable-markup`.

## Checklist

### DIFF
--- a/packages/dev/tsconfig/README.md
+++ b/packages/dev/tsconfig/README.md
@@ -1,0 +1,139 @@
+# @lokalise/tsconfig
+
+Shared TypeScript configuration for Lokalise projects.
+
+## Getting Started
+
+### Requirements:
+- TypeScript `^5.8.0`
+- ESM codebase
+
+### Installation:
+```bash
+npm install --save-dev @lokalise/tsconfig
+```
+
+TypeScript configuration depends on whether you're using `tsc` to build your code or a `bundler` (Vite, esbuild, SWC, etc.).
+
+## **Using a Bundler**
+
+Create a `tsconfig.json` for type-checking:
+
+```jsonc
+{
+  "extends": "@lokalise/tsconfig/bundler",
+  "extends": "@lokalise/tsconfig/bundler-dom", // When your code runs in the DOM
+  // Whitelist files that should be used for type-checking:
+  "include": ["src/**/*", "test/**/*", "vitest.config.ts"]
+}
+```
+**Note:** Choose only one `"extends"` option for `tsconfig.json`.
+
+## **Using `tsc` (TypeScript Compiler)**
+
+Create a `tsconfig.json` for type-checking:
+
+```jsonc
+{
+  "extends": "@lokalise/tsconfig/tsc",
+  "extends": "@lokalise/tsconfig/tsc-dom", // When your code runs in the DOM
+  // Whitelist files that should be used for type-checking:
+  "include": ["src/**/*", "test/**/*", "vitest.config.ts"]
+}
+```
+**Note:** Choose only one `"extends"` option for `tsconfig.json`.
+
+Create a `tsconfig.build.json` for building the code:
+
+```jsonc
+{
+  "extends": ["./tsconfig.json", "@lokalise/tsconfig/build-app"], // For an app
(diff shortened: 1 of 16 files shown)


packages/dev/tsconfig/configs/base.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "target": "ES2022",
+    "skipLibCheck": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    "moduleDetection": "force",
```
kibertoad, line 8: why use force and not auto?

CatchMe2: This option forces TypeScript to consider all files as modules. This helps to avoid '[cannot redeclare block-scoped variable](https://www.totaltypescript.com/cannot-redeclare-block-scoped-variable)' errors.

packages/dev/tsconfig/configs/base.json
```diff
@@ -0,0 +1,20 @@
...
+    "verbatimModuleSyntax": true,
+    "noEmit": true,
```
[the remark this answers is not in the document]
CatchMe2: Yeah, the idea is that `tsconfig.json` is always used for type-checking. Apart from that if `tsc` is used for build, it should use a dedicated `tsconfig.build.json` which should extend one of `@lokalise/tsconfig/build-xxx` that will set `"noEmit": false`.

packages/dev/tsconfig/README.md
```diff
@@ -0,0 +1,139 @@
...
+TypeScript configuration depends on whether you're using `tsc` to build your code or a `bundler` (Vite, esbuild, SWC, etc.).
+
+## **Using a Bundler**
+
+Create a `tsconfig.json` for type-checking:
+
+```jsonc
+{
+  "extends": "@lokalise/tsconfig/bundler",
+  "extends": "@lokalise/tsconfig/bundler-dom", // When your code runs in the DOM
```
kibertoad, line 25: is it possible to have two separate extends? if the idea is to use one of them, I'd comment one out

packages/dev/tsconfig/README.md
```diff
@@ -0,0 +1,139 @@
...
+  // Whitelist files that should be used for type-checking:
+  "include": ["src/**/*", "test/**/*", "vitest.config.ts"]
+}
+```
+**Note:** Choose only one `"extends"` option for `tsconfig.json`.
+
+## **Using `tsc` (TypeScript Compiler)**
```
kibertoad, line 32: we can make it more explicit by mentioning that this is typically used for backend applications and libraries

packages/dev/tsconfig/README.md
```diff
@@ -0,0 +1,139 @@
...
+```json
+{
+  "extends": "@lokalise/tsconfig/tsc",
+  "compilerOptions": {
+    "types": ["vitest/globals"]
+  }
+}
+```
+
+## **Options That Can Be Disabled for Migration**
```
kibertoad, line 86:
```suggestion
## **Options That Are Enabled By Default**
```

packages/dev/tsconfig/README.md
CatchMe2: Yeah, I was wondering how to document different configs correctly. I've added the note
```
**Note:** Choose only one `"extends"` option for **`tsconfig.json`.
```
But mby having a dedicated code-block for every option instead of having multiple `extends` in a single example tsconfig would be more readable? WDYT?

packages/dev/tsconfig/README.md
CatchMe2: The idea for this section is to list options that might make migrating to @lokalise/tsconfig more difficult. However, we suggest disabling them initially if you encounter many TypeScript errors during migration.

There are also other options that are `Enabled By Default`.

kibertoad: "Can Be Disabled To Ease Up The Adoption" maybe?.. it's not clear what migration stands for in this case

kibertoad (approved):

CarlosGamero (approved):

PR #526, opened 2025-03-11
feat: Support API contracts for payload requests in backend-http-client

kibertoad:
## Changes

Will add non-payload requests in the follow-up PR.

## Checklist

### DIFF
--- a/packages/app/backend-http-client/src/client/apiContractTypes.ts
+++ b/packages/app/backend-http-client/src/client/apiContractTypes.ts
@@ -1,0 +1,196 @@
+import type { RequestResult } from 'undici-retry'
+import type { ZodSchema, z } from 'zod'
+
+export type HeadersObject = Record<string, string>
+export type HeadersSource = HeadersObject | (() => HeadersObject) | (() => Promise<HeadersObject>)
+type FreeformRecord = Record<string, unknown>
+
+export type CommonRequestParams<
+  ResponseBody,
+  IsNonJSONResponseExpected extends boolean,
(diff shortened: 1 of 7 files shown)


packages/app/backend-http-client/src/client/apiContractTypes.ts
```diff
@@ -0,0 +1,196 @@
+import type { RequestResult } from 'undici-retry'
+import type { ZodSchema, z } from 'zod'
+
+export type HeadersObject = Record<string, string>
+export type HeadersSource = HeadersObject | (() => HeadersObject) | (() => Promise<HeadersObject>)
+type FreeformRecord = Record<string, unknown>
```
CarlosGamero, line 6: 🟢 nit: We have FreeformRecord in `universal-ts-utils`

packages/app/backend-http-client/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
     "compilerOptions": {
         "outDir": "dist",
-        "module": "commonjs",
+        "module": "ESNext",
+        "moduleResolution": "bundler",
```
CarlosGamero, line 5: 🟡 should we update biome rules to add the file extension on imports?

CarlosGamero (approved):

packages/app/backend-http-client/tsconfig.json
kibertoad: module is still commonjs, we probably can convert to ESM later, but better outside of this PR

PR #533, opened 2025-03-12
Chore/api common ts config

CarlosGamero:
## Changes

Please describe

## Checklist


packages/app/api-common/src/apiSchemas.ts
```diff
@@ -1,7 +1,7 @@
 import z from 'zod'
-import type { RefinementCtx } from 'zod/lib/types'
+import type { RefinementCtx } from 'zod'
 
-import { decodeCursor } from './cursorCodec'
+import { decodeCursor } from './cursorCodec.js'
```
CarlosGamero, line 4: This lib was declared as `module` already but all imports were missing extension, which makes me think that maybe it was not really working 🤔, so adding `major` label just in case

kibertoad (approved):

PR #534, opened 2025-03-12
Fix id-utils exports

CarlosGamero:
## Checklist


packages/app/id-utils/package.json
```diff
@@ -4,20 +4,15 @@
     "type": "module",
     "files": ["dist", "README.md", "LICENSE.md"],
     "license": "Apache-2.0",
-    "main": "./dist/index.cjs",
-    "module": "./dist/index.js",
-    "types": "./dist/index.d.ts",
+    "main": "./dist/index.js",
```
CarlosGamero, line 7: My bad, this was a major release ... 😓 it was marked as a module but it was not true, we need to unpublish 2.2.1, if you agree I will do it

packages/app/id-utils/package.json
```diff
@@ -1,23 +1,18 @@
 {
     "name": "@lokalise/id-utils",
-    "version": "2.2.1",
+    "version": "3.0.0",
```
CarlosGamero, line 3: We need to manually release this as some commonjs libs in the repo are using this

packages/app/id-utils/package.json
kibertoad: ok!

kibertoad (approved):

packages/app/id-utils/package.json
CarlosGamero:
```
npm error code E405
npm error 405 Method Not Allowed - PUT https://registry.npmjs.org/@lokalise%2fid-utils/-rev/17-0bf2181f0929940362fcd3041ede5a78 - You can no longer unpublish this package.
npm error Failed criteria: 
npm error has dependent packages in the registry
```
😢

PR #538, opened 2025-03-12
Chore/error utils common tsconfig

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/error-utils/src/bugsnag.spec.ts
+++ b/packages/app/error-utils/src/bugsnag.spec.ts
@@ -73,9 +73,9 @@ describe('bugsnag', () => {
       expect(notifySpy).toHaveBeenCalled()
 
       const callback = notifySpy.mock.calls[0]![1]
-      let context = {}
+      let context: unknown = {}
       const event = {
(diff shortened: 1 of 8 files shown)


CarlosGamero:
Prisma utils depends on error utils, so need to migrate prisma one before continuing with this

kibertoad (approved):

packages/app/error-utils/src/bugsnag.spec.ts
```diff
@@ -71,15 +72,15 @@ describe('bugsnag', () => {
       // Then
       expect(notifySpy).toHaveBeenCalled()
 
-      const callback = notifySpy.mock.calls[0][1]
+      const callback = notifySpy.mock.calls[0]![1]
       let context = {}
       const event = {
-        addMetadata: (key, obj) => {
+        addMetadata: (key: any, obj: any) => {
```
CatchMe2, line 78: Can we use `unknown` instead?

CarlosGamero: Yeah, it was not important as they are just tests, but you are right!

PR #541, opened 2025-03-13
Fixing index issue for be-http-client

CarlosGamero:
## Checklist


packages/app/backend-http-client/src/index.ts
```diff
@@ -2,16 +2,16 @@ export type {
   RequestOptions,
   HttpRequestContext,
   RequestResultDefinitiveEither,
-} from './src/client/types.js'
+} from './client/types.js'
```
CarlosGamero, line 5: index was outside of `src`, so the build was broken 😓. Sorry, I did not notice this when migrating to common tsconfig

kibertoad (approved):

PR #545, opened 2025-03-13
Chore: events-common shared tsconfig

CarlosGamero:
## Checklist


packages/app/events-common/package.json
```diff
@@ -4,26 +4,21 @@
...
-        ".": {
-            "types": "./dist/index.d.ts",
-            "require": "./dist/index.cjs",
-            "import": "./dist/index.js"
-        }
+        ".": "./dist/index.js",
+        "./package.json": "./package.json"
     },
     "scripts": {
         "build": "rimraf dist && vite build",
```
CarlosGamero, line 18: I think we can use tsc, it is a BE only dep, what do you think?

kibertoad: sure, sounds reasonable

kibertoad (approved):

PR #546, opened 2025-03-13
Chore: FE http client common tsconfig

CarlosGamero:
## Changes

Please describe

## Checklist


packages/app/frontend-http-client/package.json
```diff
@@ -53,13 +43,17 @@
...
         "@lokalise/biome-config": "^1.0.0",
         "@lokalise/universal-ts-utils": "^4.2.2",
+        "@lokalise/package-vite-config": "latest",
+        "@lokalise/tsconfig": "^1.0.2",
         "@types/node": "^22.0.0",
         "@vitest/coverage-v8": "^3.0.5",
         "jest-fail-on-console": "^3.1.2",
         "mockttp": "^3.13.0",
         "rimraf": "^6.0.0",
         "tsup": "8.4.0",
```
kibertoad, line 53: probably can be dropped

CatchMe2:
@CarlosGamero Is there an option to use `tsc` for the build process instead of `vite`?

CarlosGamero:
@CatchMe2 This is a FE library and Vite can be advantageous due to its optimisation capabilities for FE. While choosing one or the other depends on the specific use case, and tsc might be sufficient, opting for Vite can provide a robust solution and help avoid potential migrations later on

CatchMe2:
Since it is a library that will be used by FE, it will be optimized during the FE build anyway. What's the point of publishing it as an already optimized one?

On the other hand, why don't we want to be consistent and optimize every lib? I don't see a reason for optimizing libs for FE and not doing so for BE. Optimized lib for the BE means faster `npm install`, faster startup, and faster test execution.

IMO we should either optimize every lib or none.

CarlosGamero:
@CatchMe2 you convinced me, let's use `tsc` in all our libs by default, thanks for the advice 🙏 also thanks to @kibertoad for helping to decide and sharing https://cmdcolin.github.io/posts/2022-05-27-youmaynotneedabundler

CatchMe2 (approved):

PR #549, opened 2025-03-13
Chore/supported languages common tsconfig

CarlosGamero:
## Checklist


packages/app/supported-languages/rollup.config.js
```diff
@@ -1,34 +0,0 @@
...
-import typescript from '@rollup/plugin-typescript'
-
-const require = createRequire(import.meta.url)
-const packageJson = require('./package.json')
-
-// eslint-disable-next-line import/no-default-export
-export default {
-	input: 'src/index.ts',
-	output: [
-		{
```
CarlosGamero, line 17: I am not really user if we need `rollup` here, but if we can achieve the same with vite, I would vote for using it to reduce number of different tools we use

kibertoad: I agree!

CarlosGamero: Please review the last output vs the new to double check we are not missing something FE needs 🙏

kibertoad (approved):

PR #555, opened 2025-03-18
chore: enable ts extension usage

CatchMe2:
## Changes

This PR enables `.ts` extension usage that will be rewritten into `.js` during the build process.

## Checklist

### DIFF
--- a/packages/dev/tsconfig/configs/base.json
+++ b/packages/dev/tsconfig/configs/base.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
   "compilerOptions": {
+    "lib": ["ES2022"],
     "target": "ES2022",
     "skipLibCheck": true,
     "esModuleInterop": true,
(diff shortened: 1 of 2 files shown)


CarlosGamero (approved):

CarlosGamero:
kargo:merge

drdaemos:
JFYI this makes every config relying on this to be incompatible with `noEmit: false`.

Perhaps, changes like that could be better as major version change

kibertoad:
@drdaemos which parameter, specifically? allowImportingTsExtensions?

kibertoad:
@drdaemos yeah, def should be checked on CI.

but also we should move this param from base to typecheck config

kibertoad:
@drdaemos wait, but why wasn't rewrite param set to true? was it overriden somewhere?

drdaemos:
It was never set at all, “noEmit” was overwritten to “false”.

Speaking frankly I don’t know all the details of the decisions but “polyglot-ui” doesn’t use .ts imports yet. 

Maybe we should just set the “rewrite…” option to true, maybe even as a default here

PR #556, opened 2025-03-18
AP-6426 added `sendPostWithProgress()` util function

oskarski:
## Changes

I've added the function that allows sending `POST` request and track its progress. This will be used for tracking file upload and has been implemented as a result of the conversation [here](https://github.com/lokalise/autopilot/pull/5354#discussion_r1971286655)

## Checklist

### DIFF
--- a/packages/app/frontend-http-client/src/client.ts
+++ b/packages/app/frontend-http-client/src/client.ts
@@ -6,7 +6,7 @@ import type {
   PayloadRouteDefinition,
 } from '@lokalise/universal-ts-utils/api-contracts/apiContracts'
 import type { WretchResponse } from 'wretch'
-import { z } from 'zod'
+import { type ZodSchema, z } from 'zod'
 import type {
   DeleteParams,
   FreeDeleteParams,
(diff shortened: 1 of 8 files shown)


oskarski:
@kibertoad @laurismikals could I ask you for the review?

laurismikals (approved):
Looks good 💪

oskarski:
@szymonchudy could you also take a look?

packages/app/frontend-http-client/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export {
   sendPost,
+  sendPostWithProgress,
```
kibertoad, line 3: can you also add documentation for the method to the readme file?

oskarski: Done

packages/app/frontend-http-client/src/client.ts
```diff
@@ -209,6 +214,55 @@ export function sendPost<
   return sendResourceChange(wretch, 'post', params)
 }
 
+export async function sendPostWithProgress<ResponseBody>({
```
kibertoad, line 217: Does this only work with forms? If so, I wonder if it should be reflected in the function name

oskarski: Hmm, technically it should work with other data as well, however we were using it only with `FormData`. I can adjust the type from `data: FormData` to `data: Document | XMLHttpRequestBodyInit` witch matches the type from the argument of `XMLHttpRequest.send()` method. Would that be enough?

kibertoad: sure, that would be great! could you also add tests for these cases?

szymonchudy (approved):
🚀 That's amazing we will have that in the library. It would be cool to test it in Nextgen, e.g., during project creation

packages/app/frontend-http-client/src/client.ts
oskarski: I've ended up adding just `XMLHttpRequestBodyInit` which covers our use cases. Also I've added the tests for different data types

PR #560, opened 2025-03-22
redis-semaphore 5.6.2

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/background-jobs-common/tsconfig.json
+++ b/packages/app/background-jobs-common/tsconfig.json
@@ -2,10 +2,6 @@
     "extends": "@lokalise/tsconfig/tsc",
     "include": ["src/**/*", "test/**/*", "vitest.config.ts"],
     "compilerOptions": {
-        "types": ["vitest/globals"],
-        // The following rules are due to an issue with `redis-semaphore` https://github.com/swarthy/redis-semaphore/pull/230#issuecomment-2717755204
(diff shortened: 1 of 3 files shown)


kibertoad (approved):

packages/app/background-jobs-common/tsconfig.json
```diff
@@ -2,10 +2,6 @@
     "extends": "@lokalise/tsconfig/tsc",
     "include": ["src/**/*", "test/**/*", "vitest.config.ts"],
     "compilerOptions": {
-        "types": ["vitest/globals"],
-        // The following rules are due to an issue with `redis-semaphore` https://github.com/swarthy/redis-semaphore/pull/230#issuecomment-2717755204
-        // should be removed once the issue is resolved
-        "verbatimModuleSyntax": false,
-        "isolatedModules": true
```
kibertoad, line 9: do we have isolatedModules enabled in our shared config? I think it's generally preferable option to enable.

CarlosGamero: it is not enabled, I agree with but I am not sure why it is disabled maybe worth checking with Mateusz and if it makes sense let's enable it on the package level

kibertoad: sure, let's do that!

PR #581, opened 2025-04-04
feat: Separate list of languages and locales supported by Lokalise

kjamrog:
## Changes

In Lokalise, we do not support all languages and locales that are included in supported-languages library. Therefore, this PR adds separate list of supported ones

## Checklist


packages/app/supported-languages/src/index.ts
```diff
@@ -52,6 +53,10 @@ export const getAllRegions = () => {
   return regions
 }
 
+export const getLokaliseSupportedLanguagesAndLocales = () => {
```
kibertoad, line 56: why wrap in a function and not just expose the constant?

kibertoad (approved):

PR #588, opened 2025-04-10
Chore: revising vitest config

CarlosGamero:
## Changes

I noticed that we have different commands available for testing on each package, I think that having `test` and `test:ci` available is helpful. 

Also, the vitest config is different on each package, in some cases it is fine but we should try to follow some settings as much as possible to reduce the cognitive load of switching from one to the other. 
In the near future, we might want to have a package with a default config for Vitest, I feel it can be useful to avoid boilerplate definitions, this can be considered as a previous step.

## Checklist


packages/dev/eslint-config-biome-companion-frontend/README.md
```diff
@@ -1,75 +0,0 @@
-# eslint-config-biome-companion-frontend
```
CarlosGamero, line 1: Is this still valid? the package only contains a readme

kibertoad: it is intentional, you are supposed to copy-paste config from it

kibertoad:
@CarlosGamero can you add description, explaining the goals of the PR?

packages/app/api-common/vitest.config.ts
```diff
@@ -10,7 +10,7 @@ export default defineConfig({
     coverage: {
       provider: 'v8',
       include: ['src/**/*.ts'],
-      exclude: ['src/index.ts', 'src/**/*.spec.ts', 'src/**/*.test.ts'],
```
kibertoad, line 13: why not exclude test files?

kibertoad (approved):

PR #595, opened 2025-04-19
Allow Get Route Contract to have empty response.

leonaves:
We want to use a 204 with no content on a Get endpoint here:

https://github.com/lokalise/autopilot/pull/5664/files#diff-b568a28d846a8c953d129e18f5ed874d5f9a442d970c4b85f419f06e879f0278R498

We believe this is a semantically correct use of this status code, it will be provided when there is no content and we don't expect there to be any content. This is distinct from `data: []` which could be empty for another reason, like the import hasn't begun, or there was some unknown issue. See:

image

However our API contracts don't allow get routes to be empty, and give a type error trying to build a fastify route like this, so I'm proposing loosening this type. I'm open to this being challenged, but personally I don't see why we should restrict GET requests to never being empty if the HTTP spec allows it and it makes sense for certain use-cases.

## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/api-contracts/src/apiContracts.ts
+++ b/packages/app/api-contracts/src/apiContracts.ts
@@ -81,8 +81,8 @@ export type GetRouteDefinition<
   PathParamsSchema extends z.Schema<PathParams> | undefined = undefined,
   RequestQuerySchema extends z.Schema | undefined = undefined,
   RequestHeaderSchema extends z.Schema | undefined = undefined,
-  IsNonJSONResponseExpected extends boolean = false,
-  IsEmptyResponseExpected extends boolean = false,
+  IsNonJSONResponseExpected extends boolean = boolean,
+  IsEmptyResponseExpected extends boolean = boolean,
 > = CommonRouteDefinition<
   PathParams,
   SuccessResponseBodySchema,


packages/app/api-contracts/src/apiContracts.ts
```diff
@@ -81,8 +81,8 @@ export type GetRouteDefinition<
   PathParamsSchema extends z.Schema<PathParams> | undefined = undefined,
   RequestQuerySchema extends z.Schema | undefined = undefined,
   RequestHeaderSchema extends z.Schema | undefined = undefined,
-  IsNonJSONResponseExpected extends boolean = false,
-  IsEmptyResponseExpected extends boolean = false,
+  IsNonJSONResponseExpected extends boolean = boolean,
+  IsEmptyResponseExpected extends boolean = boolean,
```
kibertoad, line 85: that just adjusts the default, isn't it already perfectly possible to set isEmptyResponseExpected to true for the contract?

leonaves: It's possible to create a contract with is set to true yes, but not possible to pass that contract to `buildFastifyNoPayloadRouteHandler`:

image

Because it only accepts the default variation of a `GetRouteDefinition`:

https://github.com/lokalise/shared-ts-libs/blob/7a1e9de84655ac0fe3eed0b0f6571134667f7b7a/packages/app/fastify-api-contracts/src/fastifyApiContracts.ts#L30-L36

Maybe there's a better way to address this!

kibertoad: shouldn't it be fixed on fastify-api-contract side then?

leonaves: Hmm, that could work too, what would you suggest, something like this?:

```typescript
export function buildFastifyNoPayloadRouteHandler<
  ResponseBodySchema extends OptionalZodSchema = undefined,
  PathParams extends OptionalZodSchema = undefined,
  RequestQuerySchema extends OptionalZodSchema = undefined,
  RequestHeaderSchema extends OptionalZodSchema = undefined,
  IsNonJSONResponseExpected extends boolean = boolean,
  IsEmptyResponseExpected extends boolean = boolean,
>(
  _apiContract:
    | GetRouteDefinition<
        InferredOptionalSchema<PathParams>,
        ResponseBodySchema,
        PathParams,
        RequestQuerySchema,
        RequestHeaderSchema,
        IsNonJSONResponseExpected,
        IsEmptyResponseExpected,
      >
    | DeleteRouteDefinition<
        InferredOptionalSchema<PathParams>,
        ResponseBodySchema,
        PathParams,
        RequestQuerySchema,
        RequestHeaderSchema
      >,
  handler: FastifyNoPayloadHandlerFn<
    InferredOptionalSchema<ResponseBodySchema>,
    InferredOptionalSchema<PathParams>,
    InferredOptionalSchema<RequestQuerySchema>,
    InferredOptionalSchema<RequestHeaderSchema>
  >,
): FastifyNoPayloadHandlerFn<
  InferredOptionalSchema<ResponseBodySchema>,
  InferredOptionalSchema<PathParams>,
  InferredOptionalSchema<RequestQuerySchema>,
  InferredOptionalSchema<RequestHeaderSchema>
> {
  return handler
}
```

I guess my thinking was, the change in the PR allows a more relaxed typing of a `GetRouteDefinition` wherever it's used, not just in `buildFastifyNoPayloadRouteHandler`, which might cover more places where you could want to pass in a `GetRouteDefinition` with a `true` `IsEmptyResponseExpected`. After, all if the `buildGetRoute` factory allows the creation of such a contract, why do other methods that might want to accept a `GetRouteDefinition` as an argument need to explicitly loosen the typing?

kibertoad: @leonaves I think that the error you are seeing lies elsewhere. See https://github.com/lokalise/shared-ts-libs/pull/596 -> it works with empty responses just fine.

Could you please share the branch and the reproduction for the problem that you are seeing?

kibertoad: oh, wait, I think I can reproduce this with the `buildFastifyNoPayloadRouteHandler`. lemme dig deeper

PR #600, opened 2025-04-22
Feature: EXP-325 AWS config 2

CarlosGamero:
## Changes

- Adding utilities to resolve MQT/sns publisher and consumer options
- Adding release action
- Filling readme

## Checklist


packages/app/aws-config/src/message-queue-toolkit/MessageQueueToolkitSnsOptionsResolver.ts
```diff
@@ -0,0 +1,225 @@
...
+      createCommand: CreateTopicCommandInput
+    }
+
+const MESSAGE_TYPE_FIELD = 'type'
+const DLQ_SUFFIX = '-dlq'
+const DLQ_MAX_RECEIVE_COUNT = 5
+const DLQ_MESSAGE_RETENTION_PERIOD = 7 * 24 * 60 * 60 // 7 days in seconds
+const VISIBILITY_TIMEOUT = 60 // 1 minutes
+const HEARTBEAT_INTERVAL = 20 // 20 seconds
+const MAX_RETRY_DURATION = 2 * 24 * 60 * 60 // 2 days in seconds
```
CarlosGamero, line 39: These are the values we are using in Autopilot, and they are working fine.

Ideally, we use the same values in all our services, but if we need to override it, we can add new param options for that (not allowed for now, and will add only if we see it is needed)

packages/app/aws-config/src/message-queue-toolkit/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+ * pattern: <system_name>-<(flow|model)_name>-<(service|module)_name>
+ *
+ * Regex explanation:
+ * System name: [a-z]+(_[a-z]+)* -> One or more lowercase letters, optionally separated by underscores
+ * - -> Hyphen
+ * Flow or model name: [a-z]+(_[a-z]+)* -> One or more lowercase letters, optionally separated by underscores
+ * - -> Hyphen
+ * service or module name: [a-z]+(_[a-z]+)* -> One or more lowercase letters, optionally separated by underscores
+ */
+export const QUEUE_NAME_REGEX = /^[a-z]+(_[a-z]+)*-[a-z]+(_[a-z]+)*-[a-z]+(_[a-z]+)*$/
```
CarlosGamero, line 28: This regex is checking our naming conventions for queues and topics.

The idea is to enable `validateNamePatterns` in test mode (or in non-prod mode) to make sure all our resources are following guidelines

packages/app/aws-config/README.md
```diff
@@ -1,2 +1,136 @@
...
+  },
+};
+```
+
+#### Internal vs. External Topics
+
+- **Internal Topics** (default)
+  - You own and manage the SNS topic.
+  - `TopicConfig` must include `owner`, `service`, and optionally `externalAppsWithSubscribePermissions`.
+  - At runtime, the `MessageQueueToolkitSnsOptionsResolver` will issue a **CreateTopic** command (with name prefixing, tags,
```
kjamrog, line 89: 🟢 This description suggests that `MessageQueueToolkitSnsOptionsResolver` already creates a topic, while in fact it just resolves the config to contain topic creation parameters. And then it's up to the application on how it's used right?

packages/app/aws-config/README.md
```diff
@@ -1,2 +1,136 @@
...
+
+Read AWS configuration from environment variables:
+
+```ts
+import { getAwsConfig } from '@lokalise/aws-config';
+
+const awsConfig = getAwsConfig();
+```
+
+Set the following environment variables:
```
kibertoad, line 18: would be great to have more detailed explanation for env variables

packages/app/aws-config/README.md
```diff
@@ -1,2 +1,136 @@
...
+- `AWS_REGION` (required)
+- `AWS_KMS_KEY_ID` (optional)
+- `AWS_ALLOWED_SOURCE_OWNER` (optional)
+- `AWS_ENDPOINT` (optional)
+- `AWS_RESOURCE_PREFIX` (optional)
+- `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` (optional; or use default providers)
+
+### Resource Prefix
+
+Apply the configured resource prefix:
```
kibertoad, line 28: let's add more details on what this means and when it is supposed to used (e. g. to support multi-tenancy)

packages/app/aws-config/src/message-queue-toolkit/types.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export type MessageQueueToolkitSnsOptionsResolverConfig = Pick<
+  AwsTagsParams,
+  'appEnv' | 'system' | 'project'
+> & {
+  /** Enable validation of topic and queue names */
+  validateNamePatterns?: boolean
+}
+
+type BaseParams = {
```
kibertoad, line 16: os this is intentionally private?

packages/app/aws-config/src/message-queue-toolkit/prehandlers/createRequestContextPreHandler.ts
```diff
@@ -0,0 +1,17 @@
+import type { RequestContext } from '@lokalise/fastify-extras'
+import type { CommonLogger } from '@lokalise/node-core'
+import type { ConsumerBaseMessageType, Prehandler } from '@message-queue-toolkit/core'
+
+export type RequestContextPreHandlerOutput = { requestContext: RequestContext }
+
+export const createRequestContextPreHandler =
```
kibertoad, line 7: we probably should document that in readme as well

packages/app/aws-config/src/message-queue-toolkit/prehandlers/createRequestContextPreHandler.ts
```diff
@@ -0,0 +1,17 @@
...
+  (
+    logger: CommonLogger,
+  ): Prehandler<ConsumerBaseMessageType, unknown, RequestContextPreHandlerOutput> =>
+  (event, _context, outputs, next) => {
+    outputs.requestContext = {
+      reqId: event.metadata.correlationId,
+      logger: logger.child({ 'x-request-id': event.metadata.correlationId }),
```
kibertoad, line 14: I'd suggest making this parametrized, with default value of `'x-request-id'`

packages/app/aws-config/src/message-queue-toolkit/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './types.ts'
```
kibertoad, line 1: do we need this barrel file?

packages/app/aws-config/src/message-queue-toolkit/MessageQueueToolkitSnsOptionsResolver.ts
```diff
@@ -0,0 +1,225 @@
...
+        ? {
+            // allows to retry failed messages immediately
+            terminateVisibilityTimeout: true,
+            batchSize: params.batchSize,
+          }
+        : {
+            heartbeatInterval: HEARTBEAT_INTERVAL,
+            batchSize: params.batchSize,
+          },
+      deadLetterQueue: params.isTest
```
kibertoad, line 159: we should list all opinionated defaults and conventions coming from our wrapper in the readme

packages/app/aws-config/README.md
CarlosGamero: Correct! I adjusted slightly the description to make it clear

packages/app/aws-config/README.md
CarlosGamero: :goku_on_it

packages/app/aws-config/src/message-queue-toolkit/types.ts
CarlosGamero: Yes, it contains shared params between publisher and consumer, those are the types we should be using.

packages/app/aws-config/README.md
CarlosGamero: Done :D

packages/app/aws-config/README.md
CarlosGamero: Improved! please let me know if it not clear

packages/app/aws-config/src/message-queue-toolkit/prehandlers/createRequestContextPreHandler.ts
CarlosGamero: Added :D

packages/app/aws-config/src/message-queue-toolkit/prehandlers/createRequestContextPreHandler.ts
CarlosGamero: Mmm, this is a very opinionated library already, and I think our convention is to use `x-request-id` (correct me if I'm wrong)

If so, I would say that not allowing it to be overridden is even beneficial not to leave space for misconfiguration, but please feel free to disagree 🙏

packages/app/aws-config/README.md
```diff
@@ -1,2 +1,136 @@
...
+- **External Topics** (`isExternal: true`)
+  - The SNS topic is pre‑existing and managed outside your application.
+  - `TopicConfig` includes `topicName`, `isExternal: true`, and your `queues`, but **must omit** `owner`, `service`, 
+   and `externalAppsWithSubscribePermissions`.
+  - At runtime, the resolver will only locate (prefix) the existing topic by name and subscribe your 
+   queues—**no topic creation or tagging** is attempted.
+
+Under the hood, the TypeScript union enforces this shape.
+
+### Message Queue Toolkit SNS Resolver
```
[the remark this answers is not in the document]
CarlosGamero: This is a nice point! We should integrate it into the template once the package is released. Do you think we could link the template to this readme as "Full usage example"? I feel this is a better option to be honest

packages/app/aws-config/src/message-queue-toolkit/index.ts
kibertoad: no strong opinion, but personally I'm not a fan of barrel files, one for public exports seems enough for me

packages/app/aws-config/README.md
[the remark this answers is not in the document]
kibertoad: node-service-template is meant for public usage, though, so adding lokalise-specific config there wouldn't make much sense.

We could use the more lightweight version that we will move to MQT eventually

packages/app/aws-config/src/message-queue-toolkit/MessageQueueToolkitSnsOptionsResolver.ts
CarlosGamero: Great catch! already done :D

packages/app/aws-config/src/message-queue-toolkit/index.ts
CarlosGamero: I generally like having a per "module" index files in big libs to avoid having a big common index, for me, they use to be more difficult to maintain than smaller ones separated by concerns/module.

But, in this case, it does not make too much sense as it is a small library so going to remove it

CarlosGamero: Done

packages/app/aws-config/README.md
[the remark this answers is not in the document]
kibertoad: sounds good!

kibertoad (approved):

PR #608, opened 2025-04-25
Removing @lokalise/events-common

CarlosGamero:
## Changes

We moved our events schemas to MQT, and all our services are already migrated, hence this library is no longer needed. Once this PR is merged, we will unpublish lib from npm

## Checklist


kibertoad:
@CarlosGamero npm doesn't allow unpublishing packages, at best you can deprecate them

CarlosGamero:
@kibertoad You are right! The command is `unpublish`, but it is actually deprecating all versions for the package 😅

PR #617, opened 2025-05-07
bull queues grouping

CarlosGamero:
## Changes

Adding new queue config param enabling out-of-the-box Bull dashboard queue feature

Screenshot 2025-05-08 at 17 20 18

## Checklist


packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
```diff
@@ -51,8 +51,10 @@ export type BackgroundJobProcessorConfig<
 > = {
   queueId: string
   isTest: boolean
-  // Name of a webservice or a module running the bg job. Used for logging/observability
+  /** Name of a webservice or a module running the bg job. Used for logging/observability */
   ownerName: string
+  /** Used to compose the queue name and allow bull dashboard grouping feature */
+  bullDashboardGrouping?: string[]
```
CarlosGamero, line 57: The idea is to have something like:
```
{
 queueId: 'my-queue'
 bullDashboardGrouping: ['my-service', 'my-module']
}

// Result queue id -> my-service.my-module.my-queue (delimiter TBD, using dot as an example)
```

Only adding the new param to the old processor, the new one should not have it, as it should be handled by the queue manager.

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
```diff
@@ -95,7 +95,7 @@ export class QueueManager<
 
     if (enabled === false) return // if it is disabled -> skip
 
-    if (!this.startPromise) this.startPromise = this.internalStart(enabled)
+    this.startPromise ??= this.internalStart(enabled)
```
kibertoad, line 98: What's the benefit over a simple if?

packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
kibertoad: What's the rationale behind free-fprm grouping? Why not enforce particular structure by having specific fields for aervice and module, to encourage consistency?

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
CarlosGamero: I think there is no benefit, it is coming from a Sonar warning and just applied the suggestion

kibertoad: perdonally I find this syntax more confusing.

@dariacm any opinion on this?

dariacm: Oh wow... I have never seen this syntax 😄 

I also find it confusing, I would favor readability over shorter syntax if there is no other benefit from using it 🙇‍♀️

packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
kibertoad: in order to avoid breaking changes, we can add opt-in flag for enabling grouping prefixing.

Personally I would lean more into being opinionated here, and support extra metadata fields that could be used for prefixing if present, but more flexible solution is OK as well. Maybe we can provide some common group resolvers, that take service metadata and convert them into particularly ordered string array?

packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts
kibertoad: maybe let's disable this Sonarqube suggestion?

dariacm: We can disable it for projects, but I think this is coming from @CarlosGamero local env, right? There is no project associated to this repo

CarlosGamero: Yeah, it’s coming from my local env, I do no have preference so I will revert this change 🙏

packages/app/background-jobs-common/src/background-job-processor/processors/types.ts
CarlosGamero: I do not have a strong opinion; my thinking about a generic solution is to allow grouping by service/team/module, or all of them. It is up to the user to decide it (even though in our service abstraction, we force specific fields and order). 

> Maybe we can provide some common group resolvers, that take service metadata and convert them into particularly ordered string array?

Mmm, this is an interesting suggestion. Could you please provide an example about what you have in mind? maybe a quick huddle might be more effective in this case 😅

CarlosGamero: Discussed, for visibility: We will use this API but also add a common grouping resolver which will take care of the sorting of the params so we avoid having to replicate it on every service

packages/app/background-jobs-common/src/background-job-processor/public-utils/constants.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Delimiter used to group queues in BullMQ dashboard
+ */
+export const QUEUE_GROUP_DELIMITER = '.'
```
CarlosGamero, line 4: My initial idea was to use `:`, but queue IDS cannot contain it, it is a forbidden character, so I decided to use `.`, please let me know if you disagree

packages/app/background-jobs-common/src/background-job-processor/managers/QueueRegistry.ts
```diff
@@ -1,29 +1,92 @@
...
-  private readonly supportedQueuesMap: Record<string, Queues[number]> = {}
   public readonly queueIds: Set<string>
 
-  constructor(supportedQueues: Queues) {
-    this.queueIds = new Set<string>()
+  private readonly factory: BullmqQueueFactory<QueueType, QueueOptionsType>
+  private readonly redisConfig: RedisConfig
+
+  private readonly queuesConfig: Record<string, Queues[number]> = {}
+  private queues: Record<QueueConfiguration<QueueOptionsType>['queueId'], QueueType> = {}
```
CarlosGamero, line 22: Moving queue handling to the registry, makes sense to have it here

packages/app/background-jobs-common/README.md
```diff
@@ -84,6 +84,8 @@ To set up a queue configuration, you need to define a list of objects containing
 
 - **`queueId`**: The unique identifier for the queue.
 - **`queueOptions`**: Options for the queue. Refer to the [BullMQ documentation](https://docs.bullmq.io/guide/queues) for more details.
+- **`bullDashboardGrouping`**: Optional array of strings for grouping queues in [bull-dashboard](https://github.com/felixmosh/bull-board),
```
kibertoad, line 87: linking to https://github.com/felixmosh/bull-board/pull/867 might be helpful

kibertoad: wonder if it makes sense to explain that array groups are applied left to right (first string in the array is the top group, and we go down from there in the tree)

packages/app/background-jobs-common/src/index.ts
```diff
@@ -5,8 +5,6 @@ export * from './background-job-processor/index.ts'
 export type { BarrierCallback, BarrierResult } from './background-job-processor/barrier/barrier.ts'
 export { AbstractPeriodicJob } from './periodic-jobs/AbstractPeriodicJob.ts'
```
kibertoad, line 7: we don't expose the common resolver

packages/app/background-jobs-common/README.md
CarlosGamero: Great catch!!

CarlosGamero: Great catch! my bad, thank you 🙏

packages/app/background-jobs-common/README.md
```diff
@@ -84,6 +84,8 @@ To set up a queue configuration, you need to define a list of objects containing
 
 - **`queueId`**: The unique identifier for the queue.
 - **`queueOptions`**: Options for the queue. Refer to the [BullMQ documentation](https://docs.bullmq.io/guide/queues) for more details.
```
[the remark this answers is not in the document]
CarlosGamero: Absolutely! will add it

packages/app/background-jobs-common/src/index.ts
CarlosGamero: Great catch! my bad, thank you 🙏

packages/app/background-jobs-common/README.md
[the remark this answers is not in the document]
CarlosGamero: Done, please let me know if it not clear enough 😓

packages/app/background-jobs-common/README.md
CarlosGamero: Good point! already added :D

CarlosGamero: Ops, sorry didn’t see the second part of the message😓 will do it tomorrow

CarlosGamero: readme updated, please let me know if it is not clear 🙏

kibertoad (approved):
thank you, this is perfect!

PR #618, opened 2025-05-07
Feat: faster vault utils

leonaves:
## Changes

So the title is faster vault utils, but I guess really it's "enabling faster vault utils". I'm exporting some lower level utils, as logging in for every `vaultGetVars` is massively slowing startup time for autopilot when using `synchronizeEnvFileWithVault` in a loop, so really it's best to pull it out. Made a minor improvement to not `brew list` when checking vault is installed, and, the main thing that's gonna help downstream is created a new async version of vaultGetVars which is gonna help us parallelise the above mentioned loop in our startup scrtips.

## Checklist

### DIFF
--- a/packages/app/script-utils/src/vault/syncEnvWithVault.ts
+++ b/packages/app/script-utils/src/vault/syncEnvWithVault.ts
@@ -13,7 +13,7 @@ import { vaultGetVars, vaultLogin } from './vault.ts'
  * @param envContents String contents of the .env file
  * @param parsedEnv Record of .env file key-values (parsed using `dotenv.parse`)
  */
-const upsertEnvValue = (
+export const upsertEnvValue = (
   key: string,
   value: string,
   envContents: string,
@@ -43,9 +43,10 @@ const upsertEnvValue = (
  *
  * @param envVars Record of variable key-values
  * @param file Path to the .env file
+ * @param createEmpty Whether to create empty file even if no variables are passed
  */
-export const updateEnvFile = (envVars: Record<string, string>, file: string) => {
-  if (Object.entries(envVars).length === 0) {
+export const updateEnvFile = (envVars: Record<string, string>, file: string, createEmpty = false) => {
+  if (Object.entries(envVars).length === 0 && !createEmpty) {
     globalLogger.info(`Skipping env file ${file}`)
     return
   }
--- a/packages/app/script-utils/src/index.ts
+++ b/packages/app/script-utils/src/index.ts
@@ -1,2 +1,3 @@
 /* c8 ignore next */
 export { synchronizeEnvFileWithVault } from './vault/syncEnvWithVault.ts'
+export { vaultLogin, vaultGetVars, vaultGetVarsAsync } from './vault/vault.ts'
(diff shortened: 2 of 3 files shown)


packages/app/script-utils/src/vault/syncEnvWithVault.ts
```diff
@@ -43,9 +43,10 @@ const upsertEnvValue = (
  *
  * @param envVars Record of variable key-values
  * @param file Path to the .env file
+ * @param createEmpty Whether to create empty file even if no variables are passed
  */
-export const updateEnvFile = (envVars: Record<string, string>, file: string) => {
-  if (Object.entries(envVars).length === 0) {
+export const updateEnvFile = (envVars: Record<string, string>, file: string, createEmpty = false) => {
```
leonaves, line 48: So this function is just replicated wholesale in the autopilot scripts apart from this one addition which seems silly. it's not a breaking change as without the new param it has the same behaviour, but now we can use this one instead of just a copied and pasted version.

packages/app/script-utils/src/vault/syncEnvWithVault.ts
```diff
@@ -13,7 +13,7 @@ import { vaultGetVars, vaultLogin } from './vault.ts'
  * @param envContents String contents of the .env file
  * @param parsedEnv Record of .env file key-values (parsed using `dotenv.parse`)
  */
-const upsertEnvValue = (
+export const upsertEnvValue = (
```
leonaves, line 16: This function is currently just copied and pasted into the autopilot setup scripts word-for-word so I thought better to expose it and have a single source of truth.

packages/app/script-utils/src/vault/syncEnvWithVault.ts
```diff
@@ -43,9 +43,14 @@ const upsertEnvValue = (
...
  * @param file Path to the .env file
+ * @param createEmpty Whether to create empty file even if no variables are passed
  */
-export const updateEnvFile = (envVars: Record<string, string>, file: string) => {
-  if (Object.entries(envVars).length === 0) {
+export const updateEnvFile = (
+  envVars: Record<string, string>,
+  file: string,
+  createEmpty = false,
+) => {
```
leonaves, line 52: So this function is just replicated wholesale in the autopilot scripts apart from this one addition which seems silly. it's not a breaking change as without the new param it has the same behaviour, but now we can use this one instead of just a copied and pasted version.

VSevostianov (approved):
Not sure if my approve allows anything, but I like what I see :)

leonaves:
Any ideas on why this test is failing would be greatly appreciated, can't see how it's to do with my changes.

CarlosGamero:
@leonaves could you please update AP? 🙏 also, please share which test is flaky, I will try to have a look :D

leonaves:
I'm updating AP as we speak :) PR incoming very soon. 

The test suite that was flaky was `frontend-http-client/src/client.spec.ts`

My guess is there's some side effects in the tests that mean when they run together they can interfere with each other because no individual test fails when you run it alone but if you run them all sometimes one does, and it's not always the same one.

PR #622, opened 2025-05-09
Adding resolved queue id getter to old processor

CarlosGamero:
## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
@@ -170,6 +170,10 @@ export abstract class AbstractBackgroundJobProcessor<
     return this._spy
   }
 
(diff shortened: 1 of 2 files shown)


packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -170,6 +170,10 @@ export abstract class AbstractBackgroundJobProcessor<
     return this._spy
   }
 
+  public get resolvedQueueId(): string {
+    return resolveQueueId(this.config)
+  }
+
```
CarlosGamero, line 176: @kibertoad in theory this should be needed, but we have a legacy flow producer and I need a way to get the resolved queue id to not hardcode it.

I am adding it to old processo only, we should not use flow produce if it is not supported by our library and once we add support, this should be handled internally

kibertoad (approved):

PR #623, opened 2025-05-12
EXP-438 sqs queue names and prefix length validation

kjamrog:
## Changes

- Adds validation for:
  - maximum SQS queue name length 
  - maximum SNS topic name length 
  - maximum AWS resource prefix length

## Checklist

### DIFF
--- a/packages/app/aws-config/README.md
+++ b/packages/app/aws-config/README.md
@@ -25,7 +25,7 @@ Set the following environment variables:
 - `AWS_ENDPOINT` (optional): Custom endpoint URL for AWS services, commonly used for local testing with tools like 
  LocalStack (e.g., `http://localhost:4566`).
 - `AWS_RESOURCE_PREFIX` (optional): Prefix applied to all AWS resource names to avoid naming collisions or support 
- multi-tenancy. See **Resource Prefix** below for details. Maximum allowed prefix length is 12 characters.
+ multi-tenancy. See **Resource Prefix** below for details. Maximum allowed prefix length is 10 characters.
 - `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` (optional): AWS credentials for programmatic access. If unset, the AWS 
  SDK's default credential provider chain (environment variables, shared credentials file, EC2 instance metadata, etc.)
  is used.
--- a/packages/app/aws-config/README.md
+++ b/packages/app/aws-config/README.md
@@ -25,7 +25,7 @@ Set the following environment variables:
 - `AWS_ENDPOINT` (optional): Custom endpoint URL for AWS services, commonly used for local testing with tools like 
  LocalStack (e.g., `http://localhost:4566`).
 - `AWS_RESOURCE_PREFIX` (optional): Prefix applied to all AWS resource names to avoid naming collisions or support 
- multi-tenancy. See **Resource Prefix** below for details.
+ multi-tenancy. See **Resource Prefix** below for details. Maximum allowed prefix length is 12 characters.
 - `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` (optional): AWS credentials for programmatic access. If unset, the AWS 
  SDK's default credential provider chain (environment variables, shared credentials file, EC2 instance metadata, etc.)
  is used.
--- a/packages/app/aws-config/src/applyAwsResourcePrefix.ts
+++ b/packages/app/aws-config/src/applyAwsResourcePrefix.ts
@@ -7,5 +7,5 @@ import type { AwsConfig } from './awsConfig.ts'
  */
 export const applyAwsResourcePrefix = (resourceName: string, awsConfig: AwsConfig): string => {
   if (!awsConfig.resourcePrefix) return resourceName
-  return `${awsConfig.resourcePrefix}_${resourceName}`
+  return `${awsConfig.resourcePrefix}${resourceName}`
 }
--- a/packages/app/aws-config/src/applyAwsResourcePrefix.ts
+++ b/packages/app/aws-config/src/applyAwsResourcePrefix.ts
@@ -7,5 +7,5 @@ import type { AwsConfig } from './awsConfig.ts'
  */
 export const applyAwsResourcePrefix = (resourceName: string, awsConfig: AwsConfig): string => {
   if (!awsConfig.resourcePrefix) return resourceName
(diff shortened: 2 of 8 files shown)


packages/app/aws-config/src/message-queue-toolkit/MessageQueueToolkitSnsResolver.spec.ts
```diff
@@ -163,7 +200,7 @@ describe('MessageQueueToolkitSnsOptionsResolver', () => {
       it('should work using all properties', () => {
         const result = resolver.resolvePublisherBuildOptions({
           topicName,
-          awsConfig: buildAwsConfig({ resourcePrefix: 'preffix' }),
+          awsConfig: buildAwsConfig({ resourcePrefix: 'preffix_' }),
```
kibertoad, line 203: is double ff intentional?

kibertoad (approved):

packages/app/aws-config/src/applyAwsResourcePrefix.ts
```diff
@@ -7,5 +7,5 @@ import type { AwsConfig } from './awsConfig.ts'
  */
 export const applyAwsResourcePrefix = (resourceName: string, awsConfig: AwsConfig): string => {
   if (!awsConfig.resourcePrefix) return resourceName
-  return `${awsConfig.resourcePrefix}_${resourceName}`
+  return `${awsConfig.resourcePrefix}${resourceName}`
```
kjamrog, line 10: Removing `_` delimiter to use the same approach as Expert

packages/app/aws-config/src/message-queue-toolkit/MessageQueueToolkitSnsResolver.spec.ts
kjamrog: Probably a typo, I'll adjust it

kjamrog: updated ✅

packages/app/aws-config/src/applyAwsResourcePrefix.ts
kibertoad: @CarlosGamero does this change make sense to you?

kjamrog: Just for context: right now other apps (Expert) do not use delimiter and therefore to have matching names we need to use different env values: `{prefix}_` and `{prefix}`. 
While I would prefer to have delimiter, it's not that easy to add it on other side, therefore I'm adjusting the approach here. Let me know if you need more info

packages/app/aws-config/README.md
```diff
@@ -25,7 +25,7 @@ Set the following environment variables:
 - `AWS_ENDPOINT` (optional): Custom endpoint URL for AWS services, commonly used for local testing with tools like 
  LocalStack (e.g., `http://localhost:4566`).
 - `AWS_RESOURCE_PREFIX` (optional): Prefix applied to all AWS resource names to avoid naming collisions or support 
- multi-tenancy. See **Resource Prefix** below for details.
+ multi-tenancy. See **Resource Prefix** below for details. Maximum allowed prefix length is 12 characters.
```
CarlosGamero, line 28: Why 12? it sounds like too much to me

packages/app/aws-config/src/applyAwsResourcePrefix.ts
CarlosGamero: I feel the approach we use in the library is the right one. The objective of `_` is to have a clear separator between the queue name and the prefix to have some consistency in AWS.

Actually, if you have a better separator in mind, feel free to share, and let's discuss changing it, but IMO it is beneficial to ensure consistency

packages/app/aws-config/README.md
kjamrog: That's a maximum that will fit into AWS limits (considering 64 as max queue name length). Do you think we would benefit from making it smaller? Or you would prefer to make queue name limit bigger instead?

packages/app/aws-config/src/message-queue-toolkit/MessageQueueToolkitSnsOptionsResolver.ts
```diff
@@ -39,6 +39,8 @@ const VISIBILITY_TIMEOUT = 60 // 1 minutes
 const HEARTBEAT_INTERVAL = 20 // 20 seconds
 const MAX_RETRY_DURATION = 2 * 24 * 60 * 60 // 2 days in seconds
 
+const MAX_QUEUE_NAME_LENGTH = 64
+
```
CarlosGamero, line 43: Could you please add a comment explaining the reason for this 🙏

packages/app/aws-config/src/applyAwsResourcePrefix.ts
kibertoad: yeah, that was my feeling as well. I'd change Expert approach instead

kjamrog: I fully agree, I based on the judgement from Expert side that it's quite complex to do. So, you would both vote for challenging that?

packages/app/aws-config/README.md
CarlosGamero: I feel that something like `PREXXXXX` (8 chars) should be enough, maybe we can round it up to 10?

CarlosGamero:
@kjamrog I think topic names have some constraints as queues, could we add similar checks for them? 🙏

packages/app/aws-config/README.md
kjamrog: `Pre-11111-` is 10 characters, so I think that's the minimum (if we'll add delimiter automatically then I can make it 9)

kjamrog:
@CarlosGamero as per AWS docs: 
> Topic names are limited to 256 characters

So I think we are safe here, but I will add the checks anyways for consistency

packages/app/aws-config/src/applyAwsResourcePrefix.ts
kjamrog: Delimiter restored

packages/app/aws-config/README.md
kjamrog: Updated to 10 ✅

kibertoad (approved):

PR #630, opened 2025-05-19
AP-7179 allow passing custom callback to `MswHelper`

oskarski:
## Changes

Please describe

## Checklist

### DIFF
--- a/packages/app/universal-testing-utils/README.md
+++ b/packages/app/universal-testing-utils/README.md
@@ -98,22 +98,23 @@ describe('MswHelper', () => {
         })
     })
 
+    // use this approach when you need to implement custom logic within mocked endpoint,
+    // e. g. call your own mock
     describe("mockValidResponseWithImplementation", () => {
         it("mocks POST request with custom implementation", async () => {
             const apiMock = vi.fn();
 
-            // you don't need specify any path params, they automatically are set to *
-            mswHelper.mockValidResponseWithImplementation(
-                postContractWithPathParams,
-                server,
-                {
-                    handleRequest: async (requestInfo) => {
-                        apiMock(await requestInfo.request.json());
+            mswHelper.mockValidResponseWithImplementation(postContractWithPathParams, server, {
+                // setting this to :userId makes the params accessible by name within the callback
+                pathParams: { userId: ':userId' },
+                handleRequest: async (requestInfo) => {
+                    apiMock(await requestInfo.request.json())
 
-                        return { id: `id-${requestInfo.params.userId}` };
-                    },
+                    return {
+                        id: `id-${requestInfo.params.userId}`,
+                    }
                 },
-            );
+            })
 
             const response = await sendByPayloadRoute(
                 wretchClient,
--- a/packages/app/universal-testing-utils/README.md
+++ b/packages/app/universal-testing-utils/README.md
@@ -98,6 +98,45 @@ describe('MswHelper', () => {
         })
     })
 
+    describe("mockValidResponseWithImplementation", () => {
+        it("mocks POST request with custom implementation", async () => {
+            const apiMock = vi.fn();
+
+            // you don't need specify any path params, they automatically are set to *
+            mswHelper.mockValidResponseWithImplementation(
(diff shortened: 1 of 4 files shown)


packages/app/universal-testing-utils/src/MswHelper.ts
```diff
@@ -21,6 +21,20 @@ export type MockParamsNoPath<ResponseBody> = {
   responseBody: ResponseBody
 } & CommonMockParams
 
+// We do not control what data is used to call the endpoint, so it is safe to assume its unknown
+type HandleRequest<ResponseBody> = (dto: unknown) => ResponseBody
+
+export type MockWithImplementationParams<PathParams, ResponseBody> = {
```
kibertoad, line 27: Can this one extend MockWithImplementationParamsNoPath?

packages/app/universal-testing-utils/src/MswHelper.ts
```diff
@@ -21,6 +21,20 @@ export type MockParamsNoPath<ResponseBody> = {
...
+  pathParams: PathParams
+  handleRequest: HandleRequest<ResponseBody>
```
kibertoad, line 29: would adding this as an optional parameter to existing methods be too clunky?

oskarski: I think it would, especially when it comes to the types, cause we would need to add a discriminated union which makes the types less readable

packages/app/universal-testing-utils/src/MswHelper.ts
oskarski: Done

packages/app/universal-testing-utils/src/MswHelper.ts
```diff
@@ -105,6 +117,50 @@ export class MswHelper {
     )
   }
 
+  mockValidResponseWithImplementation<
```
kibertoad, line 120: can you add new method to the readme?

packages/app/universal-testing-utils/src/MswHelper.ts
```diff
@@ -21,6 +21,18 @@ export type MockParamsNoPath<ResponseBody> = {
   responseBody: ResponseBody
 } & CommonMockParams
 
+export type MockWithImplementationParamsNoPath<ResponseBody> = {
+  // We do not control what data is used to call the endpoint, so it is safe to assume its unknown
+  handleRequest: (dto: unknown) => ResponseBody
```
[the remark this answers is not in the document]
oskarski: Done

packages/app/universal-testing-utils/README.md
```diff
@@ -98,6 +98,45 @@ describe('MswHelper', () => {
...
+    describe("mockValidResponseWithImplementation", () => {
+        it("mocks POST request with custom implementation", async () => {
+            const apiMock = vi.fn();
+
+            // you don't need specify any path params, they automatically are set to *
+            mswHelper.mockValidResponseWithImplementation(
+                postContractWithPathParams,
+                server,
+                {
+                    handleRequest: (dto) => {
```
kibertoad, line 110: maybe let's call it body or requestBody, as it may be more self-explanatory?

I wonder if we should provide req, headers and path params as well (or maybe just pass an entire request)

oskarski: Hmm, let's expose the whole request - this should give all the possibilities to the library users

packages/app/universal-testing-utils/src/MswHelper.ts
```diff
@@ -21,6 +22,16 @@ export type MockParamsNoPath<ResponseBody> = {
   responseBody: ResponseBody
 } & CommonMockParams
 
+export type MockWithImplementationParamsNoPath<ResponseBody> = {
```
kibertoad, line 25: shouldn't there also be a generic for the request structure? (we should be able to infer it from the API contract passed to the method)

I can help with adding that, if you want

oskarski: Yes, please 🙂

PR #632, opened 2025-05-19
chore: Async package release

andrewi-wd:
## Changes

This pull request removes individual GitHub Actions workflows for packaging and releasing multiple NPM packages, adds a single GitHub workflow where package dependencies can be set. This allows to contribute to multiple packages in a single PR, while keeps the release jobs async wherever possible.

Job example:
Screenshot 2025-05-19 at 14 32 44

Pipeline example:
Screenshot 2025-05-19 at 14 48 02

## Checklist


.github/workflows/release.packages.yml
```diff
@@ -0,0 +1,472 @@
...
+
+
+  ### Dev packages
+
+  biome-config:
+    name: Release biome-config
+    if: |
+      always() &&
+      !contains(needs.*.result, 'failure') &&
+      !contains(needs.*.result, 'cancelled') &&
```
andrewi-wd, line 43: These conditions are always required to be able to check for skipped jobs.

IE: if your new PR has changes only in 1 package, which has dependency on `biome-config`, a release of `biome-config` could be skipped or should be released first.

.github/workflows/release.packages.yml
```diff
@@ -0,0 +1,472 @@
+name: release.packages.yml
```
CarlosGamero, line 1: Should it maybe be `release.yml`? sorry but I was confused by the existing `release.package.yml` and maybe makes sense to differenciate them in naming a bit more

.github/workflows/release.packages.yml
```diff
@@ -0,0 +1,472 @@
...
+      contains(needs.changed-files-job.outputs.all_changed_files, 'packages/dev/eslint-config/')
+    needs:
+      - changed-files-job
+      - prettier-config
+    uses: lokalise/shared-ts-libs/.github/workflows/release.package.yml@main
+    with:
+      working_directory: 'packages/dev/eslint-config'
+      package_name: '@lokalise/eslint-config'
+    secrets:
+      npm_token: ${{ secrets.NPM_TOKEN }}
```
CarlosGamero, line 85: Could we move the if condition to `release.package.yml`? I see it is pretty similar in all cases just chaning path (`working_directory`)

CarlosGamero: Maybe we can also move `needs`, but not sure if it is possible

CarlosGamero:
@andrewi-wd This is a great improvement and initiative, thank you! 🙏

I have a couple of questions about how this will work: 😅

- I'm not too familiar with how the `needs` work. Could this approach introduce circular dependency issues? For example: `Package A -> Package B -> Package A`.
- With this change, we'll be able to work on multiple packages within the same PR and release them together. However, since we can only use a single semver tag (patch, minor, or major) per PR if different packages require different version bumps, we still need separate PRs for each package, right? Do you have something in mind for this?

andrewi-wd:
> @andrewi-wd This is a great improvement and initiative, thank you! 🙏
> 
> I have a couple of questions about how this will work: 😅
> 
> * I'm not too familiar with how the `needs` work. Could this approach introduce circular dependency issues? For example: `Package A -> Package B -> Package A`.

I guess it can 🤔 And both are going to fail probably.

> * With this change, we'll be able to work on multiple packages within the same PR and release them together. However, since we can only use a single semver tag (patch, minor, or major) per PR if different packages require different version bumps, we still need separate PRs for each package, right? Do you have something in mind for this?

That's the exact way it is going to work currently. I wasn't able to solve the puzzle yet 🙂 But anyways, labels won't fit here. Semver should come from some another place 🤔

.github/workflows/release.packages.yml
andrewi-wd: I've already lost some context from a time I did this for Maestro service. Let me check if it possible.

CarlosGamero:
> I guess it can 🤔 And both are going to fail probably.

We need to think about how to solve this 😅, we may already have circular deps 😅

.github/workflows/release.packages.yml
andrewi-wd: Renamed in https://github.com/lokalise/shared-ts-libs/pull/632/commits/f7b649248e9703da9c7d7315994bd02acd408944

andrewi-wd:
> We need to think about how to solve this 😅, we may already have circular deps 😅

Not sure I was able to spot any in current configurations. I mean, it is theoretically possible to add such deps by hand, but you'll end up with a failed release 🙂

CarlosGamero (approved):

PR #634, opened 2025-05-19
Update zod for backend-http-client

kibertoad:
## Changes

Underlying libraries need to be updated to start using and expecting v4 zod, meanwhile we can start preparing by bumping the dependency version and pinning classic version explicitly.

## Checklist

### DIFF
--- a/packages/app/frontend-http-client/src/client.spec.ts
+++ b/packages/app/frontend-http-client/src/client.spec.ts
@@ -232,7 +232,7 @@ describe('frontend-http-client', () => {
       expect(responseBody satisfies z.infer<typeof responseBodySchema>).toEqual({
         data: {
           code: 99,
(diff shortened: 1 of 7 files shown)


packages/app/frontend-http-client/src/client.spec.ts
```diff
@@ -232,7 +232,7 @@ describe('frontend-http-client', () => {
       expect(responseBody satisfies z.infer<typeof responseBodySchema>).toEqual({
         data: {
           code: 99,
-          url: 'http://localhost:8000/users/1?sort=-startDate%2CendDate',
+          url: `${mockServer.url}/users/1?sort=-startDate%2CendDate`,
```
kibertoad, line 235: this was flaky before and failed if mockserver would take a different port.

CarlosGamero (approved):

PR #637, opened 2025-05-21
Add option to set job options as a function

CatchMe2:
## Changes

This PR allows the use of a function that will be resolved for every scheduled job for `jobOptions`
```
  {
    queueId: LqaAsyncReviewJobProcessor.QUEUE_ID,
    jobPayloadSchema: LQA_ASYNC_REVIEW_JOB_DATA,
    jobOptions(payload:any): {
      return {
        group: {
          id: payload.withTenantGrouping ? payload.tenantId : payload.key,
        }
      }
    }
  },
```

## Checklist

### DIFF
--- a/packages/app/background-jobs-common/src/background-job-processor/managers/types.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/managers/types.ts
@@ -32,8 +32,8 @@ export type QueueConfiguration<
   queueOptions?: Omit<QueueOptionsType, 'connection' | 'prefix'>
   jobPayloadSchema: PayloadSchema
   jobOptions?:
-      | JobOptionsWithDeduplicationIdBuilder<JobOptionsType>
-      | ((payload: z.infer<PayloadSchema>) => JobOptionsWithDeduplicationIdBuilder<JobOptionsType>)
+    | JobOptionsWithDeduplicationIdBuilder<JobOptionsType>
+    | ((payload: z.infer<PayloadSchema>) => JobOptionsWithDeduplicationIdBuilder<JobOptionsType>)
 }
 
 export type SupportedQueueIds<Config extends QueueConfiguration[]> = Config[number]['queueId']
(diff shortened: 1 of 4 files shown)


packages/app/background-jobs-common/src/background-job-processor/managers/types.ts
```diff
@@ -28,8 +28,11 @@ export type QueueConfiguration<
   /** Used to compose the queue name and allow bull dashboard grouping feature */
   bullDashboardGrouping?: string[]
   queueOptions?: Omit<QueueOptionsType, 'connection' | 'prefix'>
-  jobPayloadSchema: z.ZodType<BaseJobPayload> // should extend JobPayload
-  jobOptions?: JobOptionsWithDeduplicationIdBuilder<JobOptionsType>
+  jobPayloadSchema: z.ZodType<BaseJobPayload>
+  jobOptions?:
+    | JobOptionsWithDeduplicationIdBuilder<JobOptionsType>
+    // biome-ignore lint/suspicious/noExplicitAny: We cannot infer type of payload, but we have run time validation
+    | ((payload: any) => JobOptionsWithDeduplicationIdBuilder<JobOptionsType>)
```
CarlosGamero, line 35: Do we really have run time validation?

CarlosGamero: Not sure how easy is to have it actually 😅, if it is complex I think we can go without it, ideally we would use the method for edge cases when there is no other option.

If so, please adjust the comment and add something about it on readme

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/managers/types.ts
CatchMe2: We do, it is using the same paylod as in the current idBuilder -> https://github.com/lokalise/shared-ts-libs/blob/a5851a762d2132397a1636a9a1d94d55ccc2aa29/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts#L210
The payload is parsed with provided zod schema in a higher scope -> https://github.com/lokalise/shared-ts-libs/blob/a5851a762d2132397a1636a9a1d94d55ccc2aa29/packages/app/background-jobs-common/src/background-job-processor/managers/QueueManager.ts#L159

kibertoad (approved):

PR #639, opened 2025-05-24
Feature: processor payload validation

CarlosGamero:
## Changes

We were only validating job data during the scheduling part, while it is a valid and useful feature, we also need to validate the payload on the processor side to cover the following cases:
- Payload schema migrations
- Jobs added to the queue, bypassing the manager

## Checklist


packages/app/background-jobs-common/src/background-job-processor/errors/ZodUnrecoverableError.ts
```diff
@@ -0,0 +1,11 @@
+import { UnrecoverableError } from 'bullmq'
+import { ZodError } from 'zod'
+
+export class ZodUnrecoverableError extends ZodError implements UnrecoverableError {
+  constructor(error: ZodError) {
+    super(error.errors)
+
+    this.name = UnrecoverableError.name // Needed to be fully compatible with BullMQ UnrecoverableError
```
CatchMe2, line 8: Wouldn't it make sense to extend `Unrecoverable` error instead? We might add zod.issues as an error prop.

CarlosGamero: Even if we extend from `Unrecoverable`, we still need to set up this property, and I preferred to extend from ZodError as, if I remember correctly, we have some utilities to work with them properly in error handling/reporting

CatchMe2 (approved):

PR #654, opened 2025-06-19
fastify-bullboard plugin

kibertoad:
## Changes

This extracts the bull-board plugin implementation from Flow

ToDo - add readme on how to actually integrate this into an app

## Checklist


packages/app/fastify-bullboard-plugin/README.md
```diff
@@ -0,0 +1,34 @@
+# bull-board
```
CarlosGamero, line 1: I guess we need to update the `readme`, we do not need to do it in this PR, but if we prefer to do it independently, I would suggest removing references to AP or service

kibertoad: good point! will do it in a follow-up pr, together with the rest of the docs

packages/app/fastify-bullboard-plugin/package.json
```diff
@@ -0,0 +1,66 @@
...
+        "lint:fix": "biome check --write",
+        "prepublishOnly": "npm run build"
+    },
+    "dependencies": {
+        "@bull-board/api": "^6.10.1",
+        "@bull-board/fastify": "^6.10.1",
+        "@fastify/auth": "^5.0.2",
+        "@fastify/basic-auth": "^6.0.3",
+        "@lokalise/backend-http-client": "^5.0.0",
+        "@lokalise/background-jobs-common": "^12.6.0"
```
CarlosGamero, line 30: Is `backend-http-client` needed here? 
Could `background-jobs-common` be a peer dep? the service using this plugin will have to use the common library too

packages/app/fastify-bullboard-plugin/src/bullBoard.ts
```diff
@@ -0,0 +1,108 @@
...
+} from '@lokalise/background-jobs-common'
+import { resolveGlobalErrorLogObject } from '@lokalise/node-core'
+import type { Queue } from 'bullmq'
+import type { FastifyInstance } from 'fastify'
+import fp from 'fastify-plugin'
+import type { Redis } from 'ioredis'
+import { AsyncTask, SimpleIntervalJob } from 'toad-scheduler'
+
+export type BullBoardOptions = {
+  queueProConstructor: QueueProConstructor
```
CarlosGamero, line 19: 😍

CarlosGamero, line 19: Should it be `queueConstructor`? I mean, to support basic bull mq and pro one

packages/app/fastify-bullboard-plugin/src/errorReporterPlugin.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    done()
+  })
+
+  next()
+}
+
+export const errorReporterPlugin = fp<ErrorReporterOptions>(plugin, {
+  fastify: '>=4.0.0',
+  name: 'error-reporter',
```
CarlosGamero, line 72: Not sure, but maybe we can use the node-core error handler here 🤔.

If the objective of this package is to allow services to have their own bull-board, I guess we will use that error handlers instead of this one

kibertoad: will try doing this change in a separate PR, don't want to block this one further

PR #678, opened 2025-07-18
feat: assert is never utility function

ntrpilot-lokalise:
## Changes

FINALLY lift a utility function that I have written a number of times into a shared utility.

The `assertIsNever` method asserts that the provided value is of type `never` and will cause a TypeScript error if it is not.

This is useful for exhaustive type checking in switch statements or conditional types, ensuring that all possible cases have been handled.

```typescript
const value: "case1" | "case2" = getSomeValue();

switch (value) {  
    case 'case1':    
        return 'Handled case1';  

    case 'case2':    
        return 'Handled case2';  

    default:    
        // This will trigger an error if there are unhandled cases    
        return assertIsNever(value);
}
```

## Checklist

### DIFF
--- a/packages/app/universal-ts-utils/src/node.ts
+++ b/packages/app/universal-ts-utils/src/node.ts
@@ -29,6 +29,7 @@ export * from './public/object/transformToKebabCase.ts'
 // string
 export * from './public/string/trimText.ts'
 // type
+export * from './public/type/assertIsNever.ts'
 export * from './public/type/FreeformRecord.ts'
 export * from './public/type/hasMessage.ts'
 export * from './public/type/isError.ts'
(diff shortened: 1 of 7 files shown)


packages/app/universal-ts-utils/src/node.ts
```diff
@@ -29,6 +29,7 @@ export * from './public/object/transformToKebabCase.ts'
 // string
 export * from './public/string/trimText.ts'
 // type
+export * from './public/type/assertIsNever.ts'
```
kibertoad, line 32: can you also add readme entry for it?

packages/app/universal-ts-utils/src/public/type/assertIsNever.ts
```diff
@@ -0,0 +1,38 @@
...
+ *   case 'case2':
+ *     return 'Handled case2';
+ *
+ *   default:
+ *     // This will trigger an error if there are unhandled cases
+ *     return assertIsNever(value);
+ * }
+ * ```
+ */
+export const assertIsNever = (value: never): never => {
```
kibertoad, line 27: should this be `any` and the function itself be a `is never` type guard?

kibertoad: can you add these details to the jsdoc of the method?

PR #697, opened 2025-07-31
Bull board should not fail if scheduler is already registered

CarlosGamero:
## Checklist


packages/app/fastify-bullboard-plugin/src/bullBoard.ts
```diff
@@ -88,7 +88,9 @@ async function scheduleUpdates(
     (e) => fastify.log.error(resolveGlobalErrorLogObject(e)),
   )
 
-  await fastify.register(fastifySchedule)
+  // if scheduler is not registered, register it
+  if (!fastify.scheduler) await fastify.register(fastifySchedule)
+
```
CarlosGamero, line 93: The issue was that in conversion service we were registering it twice

kibertoad (approved):

PR #700, opened 2025-08-01
Pin to known good bullboard version

kibertoad:
## Changes

This is the same version that AP uses.

## Checklist


kibertoad:
unnecessary

PR #721, opened 2025-08-27
Make releases sequential and use local workflow definition

kjamrog:
## Changes

- making release process sequential to avoid race conditions and collisions
- using local workflow yml instead for the one from main branch

## Checklist


kjamrog:
**Note:** the only relevant change is in `.github/workflows/release.yml`, the remaining part is only fixing tests and linting issues

packages/app/api-common/src/apiSchemas.spec.ts
```diff
@@ -117,7 +117,7 @@ describe('apiSchemas', () => {
               "origin": "string",
               "code": "invalid_format",
               "format": "uuid",
-              "pattern": "/^([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-8][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}|00000000-0000-0000-0000-000000000000)$/",
+              "pattern": "/^([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-8][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}|00000000-0000-0000-0000-000000000000|ffffffff-ffff-ffff-ffff-ffffffffffff)$/",
```
kibertoad, line 120: did this get broken after some package update?

kibertoad (approved):
thank you, this is awesome!

packages/app/api-common/src/apiSchemas.spec.ts
kjamrog: yes, I think it must have been zod update, not sure why it wasn't failing before

PR #890, opened 2026-03-13
Extend api contracts with route contract

CatchMe2:
## Changes

## Summary                                                                                                                                                                                              
                                                                                                                                                                                                          
- Introduces `defineRouteContract` — a new way to define contracts that unifies REST and SSE use-cases
- Adds `sendByRouteContract` to both `backend-http-client` and `frontend-http-client`, and `injectByRouteContract` + `defineFastifyRoute` to `fastify-api-contracts`, so the same contract drives HTTP  
  clients and Fastify route registration end-to-end 
- Ships `inferTypes` helpers (`InferSuccessResponse`, `InferSuccessSchema`, `HasAnyNonJsonSuccessResponse`) for consuming contract response types in application code; introduces `ContractNoBody` /  `ContractNonJsonResponse` symbols for non-JSON and empty-body scenarios
- Deprecates `buildContract` / `buildRestContract`, `buildFastifyRoute`, and `injectByContract` in favour of the new route-contract-first pattern; updates all three package READMEs accordingly

## Old vs new contract definition                                                                                                                                                                       

The old `buildContract` / `buildRestContract` API threaded every aspect of the route through explicit generic type parameters. A typical payload route required up to **9 type parameters** just to get full inference: 

```ts           
  // Old
  const contract = buildRestContract({                                                                                                                                                                    
    method: 'post',                                                                                                                                                                                       
    successResponseBodySchema: ResponseSchema,                                                                                                                                                            
    requestBodySchema: RequestBodySchema,                                                                                                                                                                 
    requestPathParamsSchema: PathSchema,                                                                                                                                                                  
    requestQuerySchema: QuerySchema,                                                                                                                                                                      
    isEmptyResponseExpected: false,                                                                                                                                                                       
    isNonJSONResponseExpected: false,                                                                                                                                                                     
    pathResolver: ({ projectId }) => `/projects/${projectId}/keys`,                                                                                                                                       
})                                                                                                                                                                                                      
```
The response type lived on a single successResponseBodySchema field with no per-status-code granularity, and flags like isEmptyResponseExpected / isNonJSONResponseExpected had to be specified as both a runtime value and a boolean generic to keep inference correct.

The new defineRouteContract drops all explicit generics. Types are inferred entirely from the shape of the object you pass in — the same DX, with significantly less noise:                             

// New
```ts
  const contract = defineRouteContract({
    method: 'post',
    requestBodySchema: RequestBodySchema,
    requestPathParamsSchema: PathSchema,
    requestQuerySchema: QuerySchema,
    pathResolver: ({ projectId }) => `/projects/${projectId}/keys`,                                                                                                                                       
    responseSchemasByStatusCode: {                                                                                                                                                                        
      200: ResponseSchema,                                                                                                                                                                                
      204: ContractNoBody,                                                                                                                                                                                
    },            
  })                                                                                                                                                                                                      
```
Per-status-code responses are first-class, and ContractNoBody / ContractNonJsonResponse symbols replace the boolean flags entirely — no generics, no duplication.                                       

▎ Note: The new API achieves the same level of type safety without being heavily dependent on explicit generics. This results in cleaner contract definitions that are easier to read, write, and maintain, while TypeScript still infers everything correctly.

## Checklist

## AI Assistance Tracking


packages/app/api-contracts/src/route-contract/defineRouteContract.ts
```diff
@@ -0,0 +1,158 @@
...
+  [K in keyof T]: K extends keyof U ? T[K] : never
+}
+
+type TypedPathRouteContract<T extends RequestPathParamsSchema> = Omit<
+  RouteContract,
+  'pathResolver' | 'requestPathParamsSchema'
+> & {
+  pathResolver: RoutePathResolver<InferSchemaOutput<T>>
+  requestPathParamsSchema?: T
+}
```
CarlosGamero: Which is the difference/reason to introduce this type instead of having the generic on `CommonRouteContract`

CarlosGamero: Makes sense, thank you 🙏

packages/app/api-contracts/src/route-contract/defineRouteContract.ts
```diff
@@ -0,0 +1,155 @@
...
+  isTextResponse,
+  type ResponseSchemasByStatusCode,
+  type SseSchemaByEventName,
+} from './contractResponse.ts'
+
+export type RequestPathParamsSchema = z.ZodObject
+
+export type CommonRouteContract = {
+  // biome-ignore lint/suspicious/noExplicitAny: Required for compatibility with generics
+  pathResolver: RoutePathResolver<any>
```
kibertoad: does this mean we lose type-safety for manual execution of pathResolver on the contract?

PR #908, opened 2026-04-02
http client send by api contract

CatchMe2:
## Changes

Adds a contract-driven HTTP request function — `sendByApiContract` — to both `frontend-http-client` and `backend-http-client`, alongside the supporting type infrastructure in `api-contracts`.

---

### `@lokalise/api-contracts`

- **`clientTypes.ts`** (new):  
  `InferSseClientResponse<T, THeaders>` and `InferNonSseClientResponse<T, THeaders>` — typed discriminated unions of `{ statusCode, headers, body }` derived from a contract's `responsesByStatusCode`.  
  SSE bodies on success codes are `AsyncIterable`; non-SSE bodies are inferred from the Zod schema / response kind.

- **`hasAnySuccessSseResponse`**:  
  Utility that checks whether any 2xx entry in a contract is an SSE response (used to auto-detect streaming mode).

- **`resolveContractResponse`**:  
  Gains a `strict` parameter (default `true`). In non-strict mode, a missing or mismatched `content-type` falls back to the entry's declared kind instead of returning `null`.  
  `anyOfResponses` always requires a content-type regardless of `strict`.

---

### `sendByApiContract` (both clients)

A unified, contract-typed request function covering all HTTP methods (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`).  

#### Key options (all default to `true`)

| Option               | Behavior |
|---------------------|----------|
| `captureAsError`    | `true` → non-2xx responses go to `Either.error`; `false` → all contract status codes go to `Either.result` with a narrowed type union |
| `validateResponse`  | Runs the response body through the contract's Zod schema; disable to skip parsing |
| `strictContentType` | Throws when the response `content-type` doesn't match the contract entry |
| `signal`            | `AbortSignal` for mid-flight cancellation |

---

### Additional behavior

- **SSE streaming**:  
  Contracts with `sseResponse` default to streaming mode automatically.  
  `result.body` is an `AsyncIterable` of schema-validated, typed events.

- **Dual-mode (`anyOfResponses`)**:  
  Pass `streaming: true | false` explicitly to select between SSE and JSON; the return type narrows accordingly.

## AI Assistance Tracking


packages/app/backend-http-client/src/client/sendByApiContract.spec.ts
```diff
@@ -0,0 +1,571 @@
...
+              return attemptNumber < 2 ? 1 : -1
+            },
+          },
+        },
+      )
+
+      expect(callCount).toBe(2)
+      expect(result.error).toMatchObject({
+        message: expect.stringContaining('Could not map response'),
+      })
```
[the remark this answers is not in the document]
CatchMe2: Outdated

packages/app/api-contracts/src/new/inferTypes.ts
```diff
@@ -88,14 +81,31 @@ export type SseEventOf<S> = {
...
+export type ContractResponseMode<T extends ResponsesByStatusCode> =
   HasAnySseSuccessResponse<T> extends true
-    ? IsUnion<AvailableResponseModes<T>> extends true
-      ? true
-      : false
-    : false
+    ? HasAnyNonSseSuccessResponse<T> extends true
+      ? 'dual'
+      : 'sse'
+    : 'non-sse'
```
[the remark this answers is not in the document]
CatchMe2: It is dual

packages/app/api-contracts/src/new/defineApiContract.ts
```diff
@@ -105,6 +108,28 @@ export const getSseSchemaByEventName = (routeConfig: ApiContract): SseSchemaByEv
...
+    } else if (isAnyOfResponses(value)) {
+      for (const response of value.responses) {
+        if (isSseResponse(response)) {
+          return true
+        }
+      }
+    }
+  }
+
+  return false
```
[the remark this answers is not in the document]
CatchMe2: This is intended

packages/app/api-contracts/src/new/clientTypes.ts
```diff
@@ -0,0 +1,109 @@
...
+      InferSchemaInput<TApiContract['requestHeaderSchema']>,
+      'headers',
+      HeadersParam<InferSchemaInput<TApiContract['requestHeaderSchema']>>
+    > & { pathPrefix?: string }
+>
+
+type InferClientResponseHeaders<TApiContract extends ApiContract> =
+  TApiContract['responseHeaderSchema'] extends z.ZodType
+    ? InferSchemaOutput<TApiContract['responseHeaderSchema']> & Record<string, string | undefined>
+    : Record<string, string | undefined>
```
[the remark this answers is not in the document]
CatchMe2: Good catch, fixed.

packages/app/backend-http-client/src/client/sendByApiContract.ts
```diff
@@ -0,0 +1,287 @@
...
+  let event = 'message'
+  let data = ''
+
+  for (const line of block.split('\n')) {
+    if (line.startsWith('event:')) {
+      event = line.slice(6).trim()
+    } else if (line.startsWith('data:')) {
+      data = line.slice(5).trim()
+    }
+  }
```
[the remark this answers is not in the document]
CatchMe2: Fixed

packages/app/backend-http-client/src/client/sendByApiContract.ts
```diff
@@ -0,0 +1,287 @@
...
+  const dispatcher = options.retryConfig
+    ? client.compose(interceptors.retry(toUndiciRetryOptions(options.retryConfig)))
+    : client
+
+  let response: Dispatcher.ResponseData
+  try {
+    response = await dispatcher.request({
+      method: apiContract.method.toUpperCase(),
+      path: buildRequestPath(apiContract.pathResolver(params.pathParams), params.pathPrefix),
+      body: params.body ? JSON.stringify(params.body) : undefined,
```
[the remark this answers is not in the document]
CatchMe2: It is fine

packages/app/frontend-http-client/src/sendByApiContract.ts
```diff
@@ -0,0 +1,233 @@
...
+  for await (const sseEvent of parseSseStream(reader, signal)) {
+    const schema = schemaByEventName[sseEvent.event]
+
+    if (!schema) {
+      throw new Error(`Schema for event "${sseEvent.event}" not found.`)
+    }
+
+    const parsed = JSON.parse(sseEvent.data)
+    yield { event: sseEvent.event, data: schema.parse(parsed) }
+  }
```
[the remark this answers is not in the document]
CatchMe2: It is fine

packages/app/frontend-http-client/docs/send-by-api-contract.md
```diff
@@ -0,0 +1,263 @@
...
+type Either<E, R> = { error: E; result: undefined } | { error: undefined; result: R }
+```
+
+`result` (when defined) has the shape:
+
+```ts
+{
+  statusCode: number
+  headers: Record<string, string>
+  body: <inferred from contract>
```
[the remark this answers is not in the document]
CatchMe2: Done

packages/app/backend-http-client/docs/sendByApiContract.md
```diff
@@ -0,0 +1,305 @@
...
+  { captureAsError: false },
+)
+
+// response.result is typed for both 200 and 404
+if (response.result.statusCode === 404) {
+  response.result.body // { message: string }
+}
+```
+
+Status codes absent from the contract always surface as `Either.error`, regardless of this option.
```
CarlosGamero: This means that if a 404 response isn’t defined in the contract, it will still end up in `Either.error`, even when `captureAsError: false` is used, right?

Just to confirm my understanding, I think it is a good behaviour

packages/app/backend-http-client/docs/sendByApiContract.md
```diff
@@ -0,0 +1,305 @@
...
+      sseResponse({ chunk: z.object({ delta: z.string() }) }),
+      z.object({ text: z.string() }),
+    ]),
+  },
+})
+
+const stream = await sendByApiContract(client, chat, { body: { message: 'hi' }, streaming: true })
+// stream.result.body: AsyncIterable<{ type: 'chunk'; data: { delta: string }; lastEventId: string; retry: number | undefined }>
+
+const json = await sendByApiContract(client, chat, { body: { message: 'hi' }, streaming: false })
```
CarlosGamero, line 171: Is streaming mandatoy in this case? Ideally it should, but if it is not and is complex to have it maybe worth mentioning the default on the doc

packages/app/backend-http-client/docs/sendByApiContract.md
```diff
@@ -0,0 +1,305 @@
...
+    signal: AbortSignal.any([AbortSignal.timeout(15_000), controller.signal]),
+  },
+)
+```
+
+When the total signal fires, the request rejects with an `AbortError` and any pending retry is stopped immediately.
+
+## Retry
+
+Pass a `retry` option to retry failed requests. Pass `true` to use all defaults, or a config object to customise behaviour:
```
CarlosGamero, line 221: I would suggest to use retry: true as the default, we want retries in almost all cases

packages/app/backend-http-client/src/new/retry.ts
```diff
@@ -0,0 +1,232 @@
...
+   * Has no effect when `timeout` is not configured.
+   *
+   * @default true
+   */
+  retryOnTimeout?: boolean
+}
+
+const defaultRetryConfig: Required<RetryConfig> = {
+  maxRetries: 2,
+  statusCodes: [
```
CarlosGamero, line 73: In some cases we are also retying `423` (Resource is blocked), not sure if such case a retry really makes sense 🤔, maybe not but just sharing

packages/app/backend-http-client/docs/sendByApiContract.md
CatchMe2: Yeah, any response that cannot be mapped with a contract will return `UnexpectedResponseError` in Either.error. I'll update the docs to make it cleaner.

packages/app/backend-http-client/src/new/retry.ts
```diff
@@ -0,0 +1,232 @@
...
+ */
+// biome-ignore lint/complexity/noExcessiveCognitiveComplexity: it is acceptable
+export async function withRetry(
+  fn: () => Promise<Dispatcher.ResponseData>,
+  config: ResolvedRetryConfig,
+  signal?: AbortSignal,
+): Promise<Dispatcher.ResponseData> {
+  let retryNumber = 0
+
+  while (true) {
```
CarlosGamero, line 186: Nit: just wondering if by using a `do while` we could make this a bit more clear by moving the "retry or not" condition to the end. Not completely sure, just thinking out loud

packages/app/backend-http-client/docs/sendByApiContract.md
CatchMe2: Yeah, it is mandatory for dual contracts:
```
// streaming param: required for dual-mode, forbidden otherwise
type StreamingParam<T extends ResponsesByStatusCode, TIsStreaming extends boolean> =
  ContractResponseMode<T> extends 'dual' ? { streaming: TIsStreaming } : { streaming?: never }
```
I'll add a note about it in the docs.

packages/app/backend-http-client/src/new/sendByApiContract.spec.ts
```diff
@@ -0,0 +1,859 @@
...
+        responsesByStatusCode: { 200: responseSchema },
+      })
+
+      await mockServer
+        .forGet('/products/1')
+        .thenJson(200, { id: 1, title: 'Backpack' }, JSON_HEADERS)
+
+      const result = await sendByApiContract(client, contract, {})
+
+      expectTypeOf(result.result).toMatchTypeOf<
```
CarlosGamero, line 49: 🟡 to use type assertions (`expectTypeOf`) there is an option that needs to be enabled on Vitest config, please check if it is already there 🙏

packages/app/backend-http-client/docs/sendByApiContract.md
CatchMe2: Yeah, I generally agree that retry should be enabled by default.

The main question for me is whether we want `retry: true` by default for idempotent HTTP methods only (non POST/PATCH), or just for all methods and rely on the communication layer to handle idempotency properly (e.g. via an idempotency-key header).

packages/app/backend-http-client/package.json
```diff
@@ -38,11 +38,12 @@
         "postversion": "biome check --write package.json"
     },
     "dependencies": {
+        "parse-sse": "^0.1.0",
```
CarlosGamero, line 41: Mmm, I don't know this library but I see it is an early stage, do we want to rely on it? worth checking with @kibertoad

packages/app/backend-http-client/src/new/retry.ts
CatchMe2: For the default list of status codes, I've copied the one from `undici-retry` https://github.com/kibertoad/undici-retry/blob/main/lib/undiciRetry.ts#L11

But in general, I wouldn't add `423 Locked` in default retryable codes. This code is generally not retryable because it represents a persistent state on the server where the resource is intentionally locked.

packages/app/backend-http-client/docs/sendByApiContract.md
CarlosGamero: I’d enable it by default and leave idempotency to the communication layer, or to the user when needed. In most cases, retries shouldn’t cause issues since we’re retrying on status codes that are generally safe. The only potential concern I see is around timeouts.

packages/app/backend-http-client/src/new/retry.ts
CarlosGamero: Makes sense thank you!

packages/app/backend-http-client/src/new/retry.ts
CatchMe2: I was thinking about that as well, but in the end decided to go with `while (true)`, because we have multiple throw and return points mid-loop (and some scope issues with the response object), `while(true)` actually keeps the logic more linear. Moving the check to the end might require extra state flags.

packages/app/backend-http-client/package.json
CatchMe2: It is a spec-compliant SSE parser, written by a well-known open-sourcer https://github.com/sindresorhus (author of type-fest or ky). I've checked its source code, and it seems fine, so I decided to use it. We could just copy-paste it, but since it is used in both be-http-client and fe-http-client, IMO it is cleaner to just install it from npm.
@kibertoad WDYT?

packages/app/backend-http-client/docs/sendByApiContract.md
CatchMe2: I'm fine with enabling it for all methods, @kibertoad WDYT?

packages/app/backend-http-client/docs/sendByApiContract.md
CatchMe2: Updated

packages/app/backend-http-client/docs/sendByApiContract.md
CatchMe2: Added

packages/app/backend-http-client/src/new/sendByApiContract.spec.ts
CatchMe2: It is already enabled
```
    typecheck: {
      enabled: true,
      include: ['src/**/*.spec.ts'],
    },
```

CarlosGamero: Awesome! thanks for checking 🙏

packages/app/backend-http-client/src/new/retry.ts
CarlosGamero: All good then! thanks for considering it!

packages/app/backend-http-client/package.json
kibertoad: yeah, sindresorhus is cool. let's keep it as a dependency and see how it goes

packages/app/backend-http-client/src/api-contract/retry.ts
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * Parses the HTTP `Retry-After` header into a delay.
+ *
+ * Per RFC 9110:
+ * - Supports either seconds or an HTTP-date
+ * - Invalid values are treated as absent
+ *
+ * Returns delay in milliseconds, or `null` if missing or invalid.
+ */
+export function parseRetryAfterHeader(headers: Dispatcher.ResponseData['headers']): number | null {
```
kibertoad, line 130: can you add comment explaining why we are not relying on built-in undici retry logic?

packages/app/backend-http-client/src/new/retry.ts
```diff
@@ -0,0 +1,232 @@
...
+
+      continue
+    }
+
+    // Retryable status code — consume the body and wait before the next attempt.
+    if (!isLastAttempt && config.statusCodes.includes(response.statusCode)) {
+      const delay = computeRetryDelay(response, retryNumber, config)
+
+      // undici response body must always be consumed or explicitly dumped.
+      await response.body.dump()
```
kibertoad: I wonder if this should be in a finally block. there is no risk of throwing before reaching this step?

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
```diff
@@ -0,0 +1,282 @@
...
+   * - `true` (default): error status codes are returned as `Either.error`, and the result type is
+   *   narrowed to success status codes only.
+   * - `false`: all status codes defined in `responsesByStatusCode` are returned as `Either.result`.
+   *
+   * Status codes absent from the contract always surface as `Either.error` regardless of this option.
+   */
+  captureAsError?: DoCaptureAsError
+}
+
+type Either<TError, TResult> =
```
kibertoad, line 68: why are we not relying on node-core for this?

packages/app/backend-http-client/src/new/sendByApiContract.ts
```diff
@@ -0,0 +1,282 @@
...
+        headers.append(key, element)
+      }
+    } else {
+      headers.append(key, value)
+    }
+  }
+
+  const result: Record<string, string | undefined> = {}
+
+  headers.forEach((value, key) => {
```
kibertoad: why do we need this second iteration?

packages/app/backend-http-client/src/new/sendByApiContract.ts
```diff
@@ -0,0 +1,282 @@
...
+): Promise<ReturnTypeForContract<TApiContract, TIsStreaming, TCaptureAsError>> {
+  const strictContentType = params.strictContentType ?? true
+  const captureAsError = params.captureAsError ?? true
+
+  const useStreaming: boolean = params.streaming ?? hasAnySuccessSseResponse(apiContract)
+
+  const requestHeaders: Record<string, string> = (await resolveRequestHeaders(params.headers)) ?? {}
+
+  if (params.reqContext) {
+    requestHeaders['x-request-id'] = params.reqContext.reqId
```
kibertoad: can we use single constant everywhere for this value?

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
```diff
@@ -0,0 +1,282 @@
...
+  const requestHeaders: Record<string, string> = (await resolveRequestHeaders(params.headers)) ?? {}
+
+  if (params.reqContext) {
+    requestHeaders['x-request-id'] = params.reqContext.reqId
+  }
+  if (useStreaming) {
+    requestHeaders.accept = 'text/event-stream'
+  }
+  if (params.body) {
+    requestHeaders['content-type'] = 'application/json'
```
kibertoad, line 245: are you sure this is a safe assumption? there are different variations of json payloads out there, and message body may not be json at all (e. g. it can be form data, XML etc)

why are we not using contracts as a source of possible request content types?

also there should be an escape hatch, user should be able to provide any values for any headers if they choose to do so, and they should always win

packages/app/backend-http-client/src/new/sendByApiContract.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+  if (useStreaming) {
+    requestHeaders.accept = 'text/event-stream'
+  }
+  if (params.body) {
+    requestHeaders['content-type'] = 'application/json'
+  }
+
+  const requestFn = () => {
```
kibertoad: this is the hottest path, and closures are not cheap.

I'd lean towards something more like

```ts
const signal =
  params.signal && params.timeout !== undefined
    ? AbortSignal.any([params.signal, AbortSignal.timeout(params.timeout)])
    : params.signal
      ? params.signal
      : params.timeout !== undefined
        ? AbortSignal.timeout(params.timeout)
        : undefined

const request = {
  method: apiContract.method.toUpperCase(),
  path: buildRequestPath(apiContract.pathResolver(params.pathParams), params.pathPrefix),
  body: params.body ? JSON.stringify(params.body) : undefined,
  query: params.queryParams,
  headers: requestHeaders,
  reset: params.disableKeepAlive ?? false,
  signal,
}

const response = params.retry
  ? await withRetry(() => client.request(request), resolveRetryConfig(params.retry), params.signal)
  : await client.request(request)
```


and tbh, I would even adjust withRetry to accept parameters rather than function, especially since we don't need it for many different cases

packages/app/api-contracts/src/new/contractResponse.ts
```diff
@@ -104,28 +104,83 @@ const matchTypedResponse = (
...
+  if (isTextResponse(entry)) {
+    return { kind: 'text' }
+  }
+  if (isBlobResponse(entry)) {
+    return { kind: 'blob' }
+  }
+  if (isSseResponse(entry)) {
+    return { kind: 'sse', schemaByEventName: entry.schemaByEventName }
+  }
+  return { kind: 'json', schema: entry }
```
kibertoad, line 117: what about all other cases, e. g. when you receive xml. yaml, html etc?

shouldn't we also utilize contract here, and/or allow an explicit escape hatch from the caller?

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
CatchMe2: This inlined `Either` type pattern is used in many of our packages. Instead of installing the whole lib, we just copied 3 lines of code.
We might consider extracting `Either` to some dedicated lib so that it could be used in every package (including those used in the browser)

packages/app/backend-http-client/src/api-contract/retry.ts
CatchMe2: Sure, added!

packages/app/backend-http-client/src/new/retry.ts
CatchMe2: This is safe because computeRetryDelay is pure synchronous arithmetic, so there is no throw path between receiving the response and starting dump().

Anyway, thanks for the comment, because I found another issue. The body drain and the retry delay were sequential, drain first, then the full delay on top, so the actual wait was drain duration +  delay. Fixed it by starting both concurrently, making the effective wait max(drain duration, delay).

packages/app/backend-http-client/src/new/sendByApiContract.ts
CatchMe2: Good catch, can be covered by a single iteration. Updated

packages/app/backend-http-client/src/new/sendByApiContract.ts
CatchMe2: Sure, updated

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
CatchMe2: This isn't covered yet, for now we're going with the pragmatic default of `application/json` which covers the vast majority of cases.

There is a plan to add support for passing `requestBodyContentType` field on `defineApiContract`, I just added it to api-contracts README, but that's a separate PR once we have the basic implementation finished.

packages/app/api-contracts/src/new/contractResponse.ts
CatchMe2: It is already covered by non-json responses. I've adjusted the docs a bit to make it clearer.

### Non-JSON responses

Use `textResponse` for text-based responses (plain text, CSV, HTML, XML, YAML, etc.):

```ts
import { defineApiContract, textResponse } from '@lokalise/api-contracts'

const exportCsv = defineApiContract({
  method: 'get',
  pathResolver: () => '/export.csv',
  responsesByStatusCode: { 200: textResponse('text/csv') },
})

const getPage = defineApiContract({
  method: 'get',
  pathResolver: () => '/page',
  responsesByStatusCode: { 200: textResponse('text/html') },
})

const getDocument = defineApiContract({
  method: 'get',
  pathResolver: () => '/document',
  responsesByStatusCode: { 200: textResponse('application/xml') },
})
```

Use `blobResponse` for binary responses (images, PDFs, etc.):

```ts
import { defineApiContract, blobResponse } from '@lokalise/api-contracts'

const downloadPhoto = defineApiContract({
  method: 'get',
  pathResolver: () => '/photo.png',
  responsesByStatusCode: { 200: blobResponse('image/png') },
})
```

packages/app/backend-http-client/src/new/sendByApiContract.ts
CatchMe2: We can’t avoid a closure entirely because `AbortSignal.timeout()` must be called fresh on each attempt, so something needs to capture `params.timeout` and invoke it per iteration.

That said, I’ve improved this by precomputing the static parts outside the closure, so it now only assembles the signal. I also switched to an explicit ternary for signal construction, which is cleaner than the array filter.

kibertoad (approved):

PR #920, opened 2026-04-22
feat: Metrics utils refactor

CarlosGamero:
## Changes

Hey team! As discussed, we’re going to introduce transaction observability managers in metrics-utils. These will help us record metrics at our entry points (BullMQ, MQT, etc.).

To do this, I first want to standardise the API for our metrics classes, so the new managers can use them under the hood without type issues.

## Checklist

## AI Assistance Tracking


packages/app/metrics-utils/src/AbstractMetric.ts
```diff
@@ -1,38 +1,32 @@
...
-export type BaseDimensionalMetricParams = {
+export type LabeledMetricParams = CommonMetricParams & {
+  name: string
+}
+
+export type DimensionalMetricParams<TDimensions extends string[]> = CommonMetricParams & {
   namePrefix: string
-  nameSuffix: string
-  helpDescription: string
+  nameSuffix?: string
```
andrewi-wd: 🟢 When it might be useful to omit suffix? As per metric guidelines - type of a metric is a must 🤔

andrewi-wd (approved):

packages/app/metrics-utils/src/AbstractMetric.ts
CarlosGamero: I thought it was useful in the cases we want to have something like `translation:counter_${dimension}`, if by metric guidelines it is a must and should be at the end maybe we could add it always as a suffix without having to define it? 

I was just trying to reduce the amount of required params but if you think it is better to keep it mandatory I will revert this change, no strong opinion from my side

andrewi-wd: > maybe we could add it always as a suffix without having to define it?

Sounds good btw!

packages/app/metrics-utils/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { AbstractCounterMetric } from './AbstractCounterMetric.ts'
-export { AbstractDimensionalCounterMetric } from './AbstractDimensionalCounterMetric.ts'
-export { AbstractDimensionalHistogramMetric } from './AbstractDimensionalHistogramMetric.ts'
-export { AbstractHistogramMetric } from './AbstractHistogramMetric.ts'
+export { AbstractDimensionalCounterMetric } from './dimensional/AbstractDimensionalCounterMetric.ts'
+export { AbstractDimensionalHistogramMetric } from './dimensional/AbstractDimensionalHistogramMetric.ts'
+export { AbstractLabeledCounterMetric } from './labeled/AbstractLabeledCounterMetric.ts'
+export { AbstractLabeledHistogramMetric } from './labeled/AbstractLabeledHistogramMetric.ts'
```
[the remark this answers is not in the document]
CarlosGamero: This is a breaking release already, all good

kibertoad:
Review from Claude:

```
  Summary

  Solid refactor: introduces Labeled / Dimensional class families, extracts getOrRegisterMetric, and threads name
  through createMetric. Breaking API changes (renames + histogram registerMeasurement signature) are labeled major.
  Overall clean, but a few real gaps.

  Bugs / concerns

  1. inc(undefined) risk in dimensional counter — dimensional/AbstractDimensionalCounterMetric.ts:46-49
  for (const [dimension, value] of Object.entries(measurement) as [TDimensions[number], number][]) {
    this.metrics.get(dimension)?.inc(value)
  }
  The cast-to-number lies. DimensionalCounterMeasurement is Partial<Record<..., number>>, so a caller can legitimately
  pass { successful: undefined } (explicit undefined is common when the value is conditional). Counter.inc(undefined)
  throws in prom-client. Either filter out undefined, coerce via ?? 0, or refine the type so undefined keys aren't
  representable. Covered by no test.

  2. Silent behavior change: existing counters no longer re-seeded with inc(0)
  Old AbstractDimensionalCounterMetric called counter.inc(0) on every construction (fresh or reused). New code only
  calls it inside createMetric, i.e. only on the factory path. The spec was updated to codify this
  (incMock).not.toHaveBeenCalled() for the "already registered" case). In practice prom-client Counters already expose
  at 0, so this is probably fine — but it is a behavior change worth calling out in the release notes, and if any
  downstream consumer depended on re-init-to-zero semantics when instantiating twice in the same process, they'll be
  surprised.

  3. No test for the new optional nameSuffix
  README prominently advertises the new {namePrefix}_{dimension} format when nameSuffix is omitted (the whole reason
  buildMetricName branches). Neither AbstractDimensionalCounterMetric.spec.ts nor
  AbstractDimensionalHistogramMetric.spec.ts exercises the no-suffix path. Easy regression surface — add one case per
  spec.

  4. Unsafe cast preserved in getOrRegisterMetric.ts:9
  return existing ? (existing as T) : factory()
  If a different class registered a metric with the same name (e.g. a Gauge), the cast succeeds and you get runtime
  failures on the wrong method. Same issue as before the refactor, but now it's a single chokepoint where an instanceof
  client.Counter/Histogram check would be cheap. Worth at least a comment.

  Code quality

  5. Redundant type alias — dimensional/AbstractDimensionalCounterMetric.ts:6-7
  type DimensionalCounterMetricConfiguration<TDimensions extends string[]> =
    DimensionalMetricParams<TDimensions>
  Pure rename, adds no fields. Inline DimensionalMetricParams<TDimensions> or delete it.

  6. buildMetricName is private — AbstractDimensionalMetric.ts:25
  No subclass can change naming (e.g. . instead of : separator). Making it protected costs nothing and leaves an escape
  hatch.

  7. Helper name — getOrRegisterMetric doesn't actually call register.registerMetric; it relies on prom-client
  auto-registering inside the factory. getOrCreateMetric would be more accurate.

  8. createMetric(name, client) signature is an abstraction leak
  name is per-dimension for dimensional metrics but always metricConfig.name for labeled. For labeled classes the
  parameter is always identical to the field, so it's just noise. Minor — fine as-is since it keeps both families on one
   abstract method.

  9. Vague PR title — CodeRabbit already flagged this; agree. Also: the description says the motivation is "so the new
  managers can use them under the hood without type issues" — worth linking to or describing the intended
  AbstractMetric<_, _, TMeasurement> consumer pattern so reviewers can verify the generic shape is actually sufficient.

  Nits

  - CommonMetricParams with only helpDescription is a lot of ceremony for one field, but it's forward-compatible so
  fine.
  - AbstractDimensionalMetric takes a client? and no-ops if undefined — same "metrics disabled" pattern as before, still
   undocumented. One sentence in the README would help.
  - README heading anchors were lowercased — good, GitHub requires that.

  Verdict

  Ship-worthy after addressing #1 (real bug surface), #3 (test gap for advertised feature), and ideally #2
  (release-notes mention). The rest are polish.
```

packages/app/metrics-utils/src/AbstractMetric.ts
CarlosGamero: Discussed offline and agreed to include a metric name builder instead of relying on internal method

CarlosGamero:
@kibertoad I think all important feedback is addressed. Point 4 is a pre-existing issue (not introduced by this PR) and non-trivial to solve cleanly, so I'd prefer to handle it separately to avoid expanding the scope of this PR if it sounds good to you.

kibertoad (approved):

PR #923, opened 2026-04-24
feat: Metrics-utils -> Prometheus transaction managers

CarlosGamero:
## Changes

Context -> https://lokalise.slack.com/archives/C0AUMN3BDMF/p1776762818009099

## Checklist

## AI Assistance Tracking


packages/app/metrics-utils/src/dimensional/AbstractDimensionalMetric.ts
```diff
@@ -1,25 +1,79 @@
...
+    | {
+        /**
+         * Lazy mode. No pre-registration happens at construction time.
+         *
+         * - If `dimensions` is provided, it acts as an allow-list: only those dimensions are
+         *   registered (lazily, on first measurement); a measurement for a dimension outside the
+         *   allow-list throws an error.
+         * - If `dimensions` is omitted, any dimension is accepted and registered lazily.
+         */
+        lazyInit: true
```
CarlosGamero, line 27: Lazy initialization is required for the transaction managers, as the transaction names are not known at creation time.

andrewi-wd (commented):
Agree with Coderabbits comment. Would prefer to have memory cleanups even when errors are thrown.

When `emitMeasurement(...)` throws

andrewi-wd (approved):

.changeset/breezy-hounds-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lokalise/metrics-utils": minor
```
CatchMe2, line 2: I'd make it major. We've added a new peerDependency, so the update is not straightforward.

CatchMe2 (approved):

.changeset/breezy-hounds-share.md
CarlosGamero: I’m okay with making it a major. If nobody objects, I’ll update it before the release.

kibertoad:
## kibertoad-review — findings

Reviewed the full diff against the `TransactionObservabilityManager` contract in `@lokalise/node-core`. Solid, well-tested PR (nice coverage bump to 100/100/95/100 and thorough lazy-init specs). A few things worth addressing before merge, ordered by priority.

### 1. `addCustomAttributes` replaces instead of merging — likely a bug ⚠️
`PrometheusLabeledTransactionManager.addCustomAttributes` rebuilds `filtered` from scratch and does `this.customLabelsByKey.set(key, filtered)`. The interface doc says *"Adds all custom attributes in an object to the current transaction"*, and the method name implies accumulation. Today:

```ts
manager.addCustomAttributes('k', { tenant_id: 'x' })
manager.addCustomAttributes('k', { project_type: 'y' }) // tenant_id is now lost
```

NewRelic-style managers accumulate across calls. Recommend merging into the existing entry (`{ ...existing, ...filtered }`). There's no test for repeated calls on the same key, which is why this slipped through — please add one.

### 2. Dimensional metrics now **throw** on an unknown dimension — observability breaking the caller 🤔
`AbstractDimensionalCounterMetric`/`HistogramMetric` previously *silently ignored* a measurement for an undeclared dimension (`this.metrics.get(dim)?.inc(...)` / early `return`). After this change, eager mode throws via `getOrRegisterMetric`. Two concerns:

- **Fail-safe vs fail-loud**: a metrics utility throwing means a single instrumentation mistake can crash a request/queue-consumer path. Instrumentation usually fails safe. If throwing is intentional (it's documented), it deserves a prominent release note — `minor` undersells a runtime-behavior change for existing consumers.
- **Partial application**: in the counter, the throw happens mid-loop in `registerMeasurement`, so earlier dimensions in the same measurement object are already `inc`'d before it throws.

The manager classes use lazy-open mode so they're unaffected — this only bites existing direct subclasses of the dimensional abstracts.

### 3. Built-in label override via `customLabels` collision (low)
In `PrometheusLabeledTransactionManager.emitMeasurement`, `...customLabels` is spread *after* `status`/`transaction_name`:

```ts
this.counter?.registerMeasurement({ status, transaction_name: transactionName, ...customLabels, increment: 1 })
```

A caller declaring `customLabels: ['status']` (or `transaction_name`) would override the built-in value and produce a duplicate in `labelNames`. Cheap guard: reject/validate `customLabels` that collide with `prometheusTransactionManagerBuiltInLabels` in the constructor.

### 4. Unbounded metric cardinality from runtime `transactionName` (low)
`PrometheusDimensionalTransactionManager` registers one label-free metric per `(transactionName, status)` in lazy-open mode, and `PrometheusLabeledTransactionManager` puts `transaction_name` in a label — both driven by caller-supplied `transactionName`. The underlying abstract supports an allow-list, but the managers don't expose it. The README rightly notes the `addCustomAttributes` no-op is there to bound cardinality, but a high-cardinality `transactionName` (e.g. one containing an id) blows up the registry just as badly. Worth a doc warning, or optionally surfacing the allow-list on the dimensional manager.

### 5. Minor / nits
- **Dead branch**: in `AbstractPrometheusTransactionManager.stop`, `this.startTimeByKey.get(key) ?? Date.now()` — `start()` always sets both maps together and `stop()` already bailed if `transactionName` is undefined, so the `?? Date.now()` fallback is unreachable.
- **Leak on abandoned transactions**: `start()` without `stop()` leaves `transactionNameByKey`/`startTimeByKey`/`customLabelsByKey` entries forever. Inherent to the interface, but there's no TTL/cleanup. Also: in the labeled manager `customLabelsByKey.delete` is the last line of `emitMeasurement`, so if `registerMeasurement` throws, that entry leaks.
- **`ManagerDimensional{Counter,Histogram}`** smuggle the already-built metric name through the `dimension` param with `buildMetricName: (dim) => dim`. It works and you've commented it, but it means the `metrics` Map is keyed by full metric name and the abstract's allow-list path is bypassed — slightly surprising for future maintainers.

### Verdict
Nothing blocking the design; (1) is the one I'd fix before merge (plus a regression test), and (2) deserves an explicit decision + release-note callout. The rest are low-priority hardening.

kibertoad (approved):

CarlosGamero:
Thanks for the thorough review @kibertoad — really useful, addressed point by point:

1. **`addCustomAttributes` replacing instead of merging** — fixed. It now merges into the existing entry (later calls add to / override earlier ones). Added regression tests for repeated calls and key override.
2. **Dimensional metrics throwing on unknown dimension** — reverted to the original fail-safe behavior: out-of-set dimensions (eager mode and lazy allow-list) are silently ignored, so instrumentation never throws into the caller's path. This also removes the partial-application issue (valid dimensions in the same measurement still apply). Since it matches pre-PR behavior, no release-note/breaking concern. Docs and tests updated.
3. **Built-in label collision via `customLabels`** — fixed. The constructor now throws if a `customLabel` collides with `status`/`transaction_name` (config error at startup, not on the runtime path). Added a test.
4. **Unbounded cardinality from `transactionName`** — went with the doc route: prominent warning in both README sections and the class-level JSDoc of both managers. Didn't expose the allow-list to keep the API surface small.

PR #945, opened 2026-05-15
feat(fastify-bullboard-plugin): bull-board 7.1.5 + BullMQ Pro support

kibertoad:
## Summary

- Upgrade `@bull-board/api` and `@bull-board/fastify` from `^6.18.0` to `^7.1.5`.
- Add support for [BullMQ Pro](https://docs.bullmq.io/bullmq-pro/introduction) queues via the official `BullMQProAdapter` introduced upstream in bull-board 7.1.x.
- Allow mixing BullMQ Pro and standard BullMQ connections in a single plugin registration via a per-connection `isPro: true` flag on each `redisConfigs` entry.

`queueConstructor` stays required; `queueProConstructor` is the only new constructor option, and its absence is validated only when a `redisConfigs` entry sets `isPro: true`.

```ts
await app.register(bullBoard, {
  queueConstructor: Queue,
  queueProConstructor: QueuePro,
  redisConfigs: [
    { host: 'redis-1', port: 6379, useTls: false },
    { host: 'redis-2', port: 6379, useTls: false, isPro: true },
  ],
  basePath: '/bull',
})
```

## Test plan

- [x] `tsc --noEmit` passes
- [x] `biome check src` passes
- [x] `vitest run` — 16/16 passing, including new coverage for Pro constructor validation
- [ ] Manual smoke test against a deployment with both Pro and non-Pro Redis connections

## Checklist

## AI Assistance Tracking

### DIFF
--- a/packages/app/fastify-bullboard-plugin/src/bullBoard.spec.ts
+++ b/packages/app/fastify-bullboard-plugin/src/bullBoard.spec.ts
@@ -1,7 +1,7 @@
 import fastifySchedule from '@fastify/schedule'
 import { Queue } from 'bullmq'
 import fastify, { type FastifyInstance } from 'fastify'
-import { beforeAll, expect, vi } from 'vitest'
+import { beforeAll, expect } from 'vitest'
 import {
   type BullBoardOptions,
   bullBoard,
@@ -9,16 +9,7 @@ import {
   type QueueProConstructor,
 } from './bullBoard.ts'
 
-vi.mock('@lokalise/background-jobs-common', async (orig) => {
-  const actual = await orig<typeof import('@lokalise/background-jobs-common')>()
-  return {
-    ...actual,
-    backgroundJobProcessorGetActiveQueueIds: vi.fn().mockResolvedValue([]),
-  }
-})
-
-const { backgroundJobProcessorGetActiveQueueIds } = await import('@lokalise/background-jobs-common')
-
+const QueuePro: QueueProConstructor = Queue as unknown as QueueProConstructor
 const BullMqQueue: QueueConstructor = Queue
 
 async function initApp(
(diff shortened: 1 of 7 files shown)


packages/app/fastify-bullboard-plugin/src/bullBoard.ts
```diff
@@ -15,16 +17,34 @@ import fp from 'fastify-plugin'
...
+  isPro?: boolean
+}
+
 export type BullBoardOptions = {
-  queueConstructor: QueueProConstructor | QueueConstructor
+  /**
+   * Constructor for non-Pro BullMQ queues. Required if any entry of `redisConfigs`
+   * does not have `isPro: true`.
+   */
+  queueConstructor?: QueueConstructor
```
CarlosGamero: Wondering if this is really needed, I mean, could we use the basic constructor when isPro is undefined or false?

CarlosGamero (approved):

packages/app/fastify-bullboard-plugin/src/bullBoard.spec.ts
```diff
@@ -105,4 +111,37 @@ describe('bull board', () => {
...
+
+  describe('mixed pro/non-pro configs', () => {
+    it('accepts both pro and non-pro redis configs in the same registration', async () => {
+      app = await initApp({
+        queueConstructor: QueueCtor,
+        queueProConstructor: QueuePro,
+        redisConfigs: [],
+        basePath: '/test-mixed',
+      })
+
```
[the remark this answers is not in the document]
kibertoad: Addressed. Renamed the misleading test in e39d29e, and added a real mixed-config test in c4abf69 that mocks `backgroundJobProcessorGetActiveQueueIds` and uses fake constructors to exercise both Pro and non-Pro construction paths. Coverage rose from ~69% to ~84% lines as a result.

packages/app/fastify-bullboard-plugin/src/bullBoard.ts
kibertoad: Good catch — simplified. `queueConstructor` is now required (it covers the common case), and only `queueProConstructor` stays optional, validated at startup when any `redisConfigs` entry sets `isPro: true`. Fixed in e39d29e.

PR #948, opened 2026-05-15
Migrate release pipeline to Changesets

CatchMe2:
## Changes

Replaces the label-based release system with [Changesets](https://github.com/changesets/changesets).

## Problems with the old system

- One semver bump for all packages in a PR — if a PR touched two packages that deserved different bump types, the PR had to be split
- No changelog — releases were a bare version bump commit with no record of what changed
- Fragile pipeline — `release.yml` had 33 manually chained jobs; adding a package meant editing it by hand

## New workflow

1. Run `pnpm changeset`, select which packages changed and the bump type for each (independently, in one PR)
2. Commit the generated `.changeset/*.md` file
3. Merge → the release workflow creates a “Release Packages” PR with bumped versions and updated changelogs
4. The "Release Packages" PR is automatically squash-merged and packages are published to npm

## Required: install the Changeset bot

Install the [Changeset bot](https://github.com/apps/changeset-bot) on this repository — it comments on PRs to show whether a changeset is present and links directly to adding one.

## Checklist

## AI Assistance Tracking


kibertoad:
> Run pnpm changeset, select which packages changed and the bump type for each (independently, in one PR)

@CatchMe2 is LLM capable of doing that programmatically and does that by default? do we need to add extra CLAUDE.md instructions, telling them to generate a changeset for every change?

CatchMe2:
> > Run pnpm changeset, select which packages changed and the bump type for each (independently, in one PR)
> 
> @CatchMe2 is LLM capable of doing that programmatically and does that by default? do we need to add extra CLAUDE.md instructions, telling them to generate a changeset for every change?

LLM should be capable of generating it if asked. It doesn't do it by default, just as it doesn't generate CHANGELOG by default. Using CLAUDE.md seems like a good idea. I'd push the migration first, and then once we confirm it works fine, we can work on some LLM instructions. WDYT?

kibertoad:
@CatchMe2 I'm a bit concerned about regression in DX for developers. If there is a time period when our automations don't work as expected, change will not necessarily be a net positive. if developers will need to add one extra instruction to generate changesets, many will forget to do so, and we will have gaps. for that reason it feels like having LLM instructions from the day one is important.

I think we also need explicit overrides for PRs without changesets - if one is not intentionally present, user needs to put an appropriate label, or CI should fail.

CatchMe2:
@kibertoad your concerns are valid.
I've found a repo that has CLAUDE.md with guidance for changesets. We can copy-paste this part
https://github.com/palantir/osdk-ts/blob/main/CLAUDE.md

In terms of enforcing changesets on CI, most repos use [changeset-bot](https://github.com/apps/changeset-bot), which adds a comment to each PR if a changeset is missing. This approach is used by many OS repos (many of them with a lot of contributions and open PRs). It should work fine for us as well.

kibertoad (approved):

PR #975, opened 2026-05-21
feat(background-jobs-common): add FlowManager for BullMQ FlowProducer

kibertoad:
## Summary

- Adds `FlowManager`, `FakeFlowManager`, `ModuleAwareFlowManager` to support BullMQ `FlowProducer` with parity to `QueueManager`: typed payload-per-queue, Zod validation, merged job options + default retry/retention, deduplication `idBuilder` on flow roots (children cannot carry `deduplication`/`debounce`/`parent` per BullMQ), dashboard grouping via `resolveQueueId`, spy support in test mode, and lazy initialization
- Adds `BullmqFlowProducerFactory` interface and implements `buildFlowProducer` on `CommonBullmqFactoryNew`
- Refactors `QueueRegistry` to expose a read-only `QueueConfigRegistry` base class for typed queue-config lookup without queue instantiation — `FlowManager` reuses it for `getQueueConfig` and `queueIds`

## Test plan

- [x] `pnpm exec tsc --noEmit` passes
- [x] `pnpm exec biome check .` passes
- [x] `pnpm exec vitest run` — 200 tests pass (27 files), including new FlowManager.spec.ts (16 tests) and ModuleAwareFlowManager.spec.ts (2 tests)
- [ ] Verified on review

## AI Assistance Tracking


packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
```diff
@@ -0,0 +1,342 @@
...
+      }
+    }
+  }
+
+  public async start(): Promise<void> {
+    if (this.flowProducer) return
+    if (!this.startPromise) this.startPromise = this.doStart()
+    await this.startPromise
+    this.startPromise = undefined
+  }
```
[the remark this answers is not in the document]
kibertoad: Fixed in 2bb9f29 — wrapped the await in `try { ... } finally { this.startPromise = undefined }` so a failed `doStart()` clears the cached promise and a subsequent `start()` can retry. Added a regression test (`start failure recovery > clears startPromise on failure so a subsequent start can succeed`) that fakes a `waitUntilReady` rejection and verifies the next `start()` succeeds. Also closed the leaked `FlowProducer` connection in `doStart()` on failure, and made `dispose()` await any in-flight start so the producer is closed even when dispose races startup.

kibertoad:
CodeRabbit nit on `ModuleAwareFlowManager.spec.ts:82-98` (dispose lazy-init managers): fixed in 2bb9f29 — the `sets lazyInitEnabled based on isTest` test now `await`s `testManager.dispose()` and `prodManager.dispose()` before returning so we do not leak handles from the eager-init path.

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
```diff
@@ -0,0 +1,380 @@
...
+  QueueManagerConfig,
+  SupportedJobPayloads,
+  SupportedQueueIds,
+} from './types.ts'
+
+/**
+ * Options accepted on a child flow node. Mirrors BullMQ's `FlowChildJob.opts`:
+ * `debounce`, `deduplication`, `parent` and `repeat` are not supported on children.
+ */
+export type FlowChildJobOptions = Omit<
```
CarlosGamero: We have a types file in this folder, should we move those types there?

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  /**
+   * Convenience factory that pairs a `FlowManager` to an existing `QueueManager`,
+   * sharing its registry. Use this when you have a `QueueManager` instance handy —
+   * TypeScript infers `Queues` cleanly through it (which it does not always do when
+   * passing `queueManager.queueRegistry` to the regular constructor, especially
+   * for `ModuleAwareQueueManager` registries).
+   */
+  public static fromQueueManager<
```
CarlosGamero: Mmm, do we actually have any use cases where the Flow Manager is started without the Queue Manager? If not, I’d prefer to enforce starting it through the Queue Manager only. That would allow us to share a single spy instance across the whole flow and remove a good amount of duplicated code.

Regarding the spy instance itself, I think it’s beneficial to keep it shared. This way, whether you inspect the Flow Manager spy, the Queue Manager spy, or the Processor spy, they would all expose the same data.

CarlosGamero: btw, queue manager is exposing queue config and queue getters, so we can use them directly from this class

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
kibertoad: Good catch — moved in 4523658. `FlowChildJobOptions`, `FlowRootJobOptions`, `FlowChildJobInput`, `FlowJobInput` and `FlowManagerConfig` now live in `managers/types.ts` alongside the QueueManager types.

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
kibertoad: Agreed — done in 4523658. `FlowManager` now takes a `QueueManager` directly in its constructor. Key effects:

- Single spy stream: `flowManager.getSpy(qid)` delegates to `queueManager.getSpy(qid)`, returning the same instance the processor sees. A new `QueueManager.recordScheduledJob(qid, job)` is what FlowManager calls in `recordNode` (no-op outside test mode), so flow nodes show up in the same spy as a direct `queueManager.schedule()` call.
- No duplicated state in FlowManager: queue configs come from `queueManager.queueRegistry`, `isTest`/`redisConfig` from `queueManager.config`, so `FlowManagerConfig` shrinks to just `{ lazyInitEnabled? }`.
- `ModuleAwareQueueManager` pairing keeps `[serviceId, moduleId]` grouping automatically.
- The `fromQueueManager` static helper and the `applyModuleGrouping` public utility are gone — both were workarounds for the registry-based wiring that no longer exists.
- `FakeFlowManager(queueManager, config?)` takes the paired QueueManager; `lazyInitEnabled` defaults to true.

Publish-only services pair a never-started QueueManager (no Redis connection, just a config holder) with a FlowManager — documented in the README under "Publish-only services".

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
```diff
@@ -0,0 +1,330 @@
...
+    flow: FlowJobInput<Queues, QueueId>,
+    opts?: FlowOpts,
+  ): Promise<JobNode> {
+    // Validate and resolve queue names up front so invalid payloads fail fast,
+    // before any network round-trip (matches `QueueManager.schedule`).
+    const rootFlowJob = this.buildRootFlowJob(flow)
+
+    await this.startIfNotStarted()
+
+    const node = await this.getFlowProducer().add(rootFlowJob, opts)
```
CarlosGamero, line 167: I don't have deep knowledge about how flow producer works, so please ignore this if it does not make sense.

On queue manager schedule we are applying some default jobs to jobs, like retry options or id generation. Also via queue config we are allowing some job config override that will apply to all jobs in the queue, are thoe things relevant here?

CarlosGamero (approved):

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
CarlosGamero: nvm, I see job options are resolved below with what I mentioned, all good :D

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
```diff
@@ -0,0 +1,330 @@
...
+  // Reverse lookup of resolved queueName (including dashboard grouping) ->
+  // queueId, so `recordNode` doesn't rescan every queue config per node.
+  private readonly queueIdByResolvedName: Map<string, SupportedQueueIds<Queues>>
+
+  private flowProducer?: FlowProducerType
+  private startPromise?: Promise<void>
+
+  constructor(
+    flowProducerFactory: BullmqFlowProducerFactory<FlowProducerType, FlowProducerOptionsType>,
+    queueManager: QueueManager<Queues, QueueType, QueueOptionsType, JobOptionsType>,
```
CarlosGamero: Small nit: for consistency, maybe would makes sense to group those in a dependencies object as we do in BackgroundJobProcessorDependenciesNew

CarlosGamero: In future we might want to add optional things there like logger, error reporter etc and that will allow doing it without breaking changes

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
kibertoad: Right — `resolveRootJobOptions` / `resolveChildJobOptions` are exact mirrors of `QueueManager.resolveJobOptions`: queue-level `jobOptions` defaults (function or static) get merged, `prepareJobOptions` slaps on the default retry/retention + a fresh `jobId`, deduplication `idBuilder` runs on the root (BullMQ does not allow it on children, so child defaults strip `deduplication`/`debounce`/`parent`/`repeat`). So every guarantee `QueueManager.schedule` makes per job carries over to every node in a flow.

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
kibertoad: Done in d95c30a. Added `FlowManagerDependencies` next to `FlowManagerConfig` in `managers/types.ts` and switched the constructor to `new FlowManager({ flowProducerFactory, queueManager }, config?)`. `FakeFlowManager(queueManager, config?)` wraps that internally so the test-facing API stays unchanged. Now optional fields like `logger`/`errorReporter` can be tacked on later without breaking the signature.

PR #977, opened 2026-05-21
chore: update changesets config with per deps configuration

CatchMe2:
## Changes

This PR adds `onlyUpdatePeerDependentsWhenOutOfRange` to prevent dependents bumps.
When set to true, Changesets will only bump peer dependents when peerDependencies are leaving the range.

## Checklist

## AI Assistance Tracking

- [ ] **Yes, AI assisted with this PR**
- [x] **No, AI did not assist with this PR**

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,8 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
+  }
 }


.changeset/config.json
```diff
@@ -7,5 +7,8 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
```
kibertoad, line 11: should we also update changesets/cli to the latest 2.31.0?

kibertoad (approved):

.changeset/config.json
CatchMe2: We're already using 2.31.0

PR #981, opened 2026-05-25
feat(api-contracts): add wildcard status code keys to api contract responses

CatchMe2:
## Changes

  - Add `WildcardStatusCodeKey` (`'1xx'`–`'5xx'` and `'default'`) as valid keys in `responsesByStatusCode`, matching OpenAPI / Fastify conventions
  - `resolveResponseEntry` now resolves with precedence: exact code → range key → `'default'`
  - `'2xx'` range participates in SSE-mode detection (`hasAnySuccessSseResponse`) and in `InferSseClientResponse` / `InferNonSseClientResponse` success-side type narrowing the same way explicit 2xx
  codes do
  - `'default'` is split into a `SuccessfulHttpStatusCode` half and a `Exclude<HttpStatusCode, SuccessfulHttpStatusCode>` half so `captureAsError` type narrowing stays accurate

## Checklist

## AI Assistance Tracking

### DIFF
--- a/packages/app/api-contracts/README.md
+++ b/packages/app/api-contracts/README.md
@@ -217,34 +217,36 @@ getSseSchemaByEventName(chatCompletion)
 ### All fields
 
 ```ts
-defineApiContract({
+type ApiContractOptions = {
   // Required
-  method: 'get' | 'post' | 'put' | 'patch' | 'delete',
-  pathResolver: (pathParams) => string,
-  responsesByStatusCode: {
-    // exact status codes
-    [statusCode: HttpStatusCode]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
-    // range keys (OpenAPI-style): '1xx' | '2xx' | '3xx' | '4xx' | '5xx' | 'default'
-    [rangeKey: WildcardStatusCodeKey]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
-  },
+  method: 'get' | 'post' | 'put' | 'patch' | 'delete'
+  pathResolver: (pathParams: Record<string, string>) => string
+  // Accepts exact codes, OpenAPI-style range keys ('1xx'–'5xx'), and a catch-all 'default'.
+  // Lookup precedence at runtime: exact code → range key → 'default'.
+  responsesByStatusCode: Partial<
+    Record<
+      HttpStatusCode | '1xx' | '2xx' | '3xx' | '4xx' | '5xx' | 'default',
+      z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
+    >
+  >
 
   // Path params — links pathResolver parameter type to the schema
-  requestPathParamsSchema: z.ZodObject,
+  requestPathParamsSchema?: z.ZodObject<z.ZodRawShape>
 
   // Request
-  requestBodySchema: z.ZodType | ContractNoBody, // required for POST / PUT / PATCH, forbidden otherwise
-  requestQuerySchema: z.ZodObject,
-  requestHeaderSchema: z.ZodObject,
+  requestBodySchema?: z.ZodType | ContractNoBody  // required for POST / PUT / PATCH, forbidden otherwise
+  requestQuerySchema?: z.ZodObject<z.ZodRawShape>
+  requestHeaderSchema?: z.ZodObject<z.ZodRawShape>
 
   // Response
-  responseHeaderSchema: z.ZodObject,
+  responseHeaderSchema?: z.ZodObject<z.ZodRawShape>
 
   // Documentation
-  summary: string,
-  description: string,
-  tags: readonly string[],
-  metadata: Record<string, unknown>,
-})
+  summary?: string
+  description?: string
+  tags?: readonly string[]
+  metadata?: Record<string, unknown>
+}
 ```
 
 ### Header schemas
--- a/packages/app/api-contracts/README.md
+++ b/packages/app/api-contracts/README.md
@@ -121,6 +121,52 @@ const chatCompletion = defineApiContract({
 })
 ```
 
+### Wildcard and default response keys
+
+In addition to exact status codes, `responsesByStatusCode` accepts OpenAPI-style range keys (`'1xx'`–`'5xx'`) and `'default'` as fallbacks.
+
+Lookup precedence at runtime: **exact code → range key → `'default'`**.
+
+```ts
+import { defineApiContract } from '@lokalise/api-contracts'
+import { z } from 'zod/v4'
+
+// '2xx' covers all 200–299 responses
+const listItems = defineApiContract({
+  method: 'get',
+  pathResolver: () => '/items',
+  responsesByStatusCode: {
+    '2xx': z.object({ items: z.array(z.string()) }),
+    '4xx': z.object({ message: z.string() }),
+  },
+})
+
+// exact code takes precedence over the range key
+const createItem = defineApiContract({
+  method: 'post',
+  pathResolver: () => '/items',
+  requestBodySchema: z.object({ name: z.string() }),
+  responsesByStatusCode: {
+    201: z.object({ id: z.string() }),
+    '4xx': z.object({ message: z.string() }),
+  },
+})
+
+// 'default' matches any status code not covered by a more specific entry
+const flexible = defineApiContract({
+  method: 'get',
+  pathResolver: () => '/data',
+  responsesByStatusCode: {
+    200: z.object({ data: z.unknown() }),
+    default: z.object({ error: z.string() }),
+  },
+})
+```
+
+The `'2xx'` range key participates in SSE detection and success/error type narrowing exactly like explicit 2xx codes: `InferNonSseClientResponse` maps it to `SuccessfulHttpStatusCode`, and `hasAnySuccessSseResponse` returns `true` when it holds an SSE schema.
+
+`'default'` is split into a success half (`SuccessfulHttpStatusCode`) and a non-success half in `InferSseClientResponse` / `InferNonSseClientResponse` so that `captureAsError` type narrowing stays correct regardless of the actual status code.
+
 ### OpenAPI response descriptions
 
 All response factories accept an optional `ResponseOptions` object as their last argument.
@@ -176,7 +222,10 @@ defineApiContract({
   method: 'get' | 'post' | 'put' | 'patch' | 'delete',
   pathResolver: (pathParams) => string,
   responsesByStatusCode: {
-    [statusCode]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
+    // exact status codes
+    [statusCode: HttpStatusCode]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
+    // range keys (OpenAPI-style): '1xx' | '2xx' | '3xx' | '4xx' | '5xx' | 'default'
+    [rangeKey: WildcardStatusCodeKey]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
   },
 
   // Path params — links pathResolver parameter type to the schema
@@ -236,7 +285,7 @@ type CsvResponse = InferNonSseSuccessResponses<typeof exportCsv['responsesByStat
 
 **`InferSseSuccessResponses<T>`** — extracts the SSE event schema map type from a `responsesByStatusCode` map. Returns `never` when no SSE schemas are present.
 
-**`HasAnySseSuccessResponse<T>`** — `true` if any 2xx entry is a `TypedSseResponse` or an `AnyOfResponses` containing one.
+**`HasAnySseSuccessResponse<T>`** — `true` if any 2xx entry (exact code or `'2xx'` range key) is a `TypedSseResponse` or an `AnyOfResponses` containing one.
 
 **`HasAnyJsonSuccessResponse<T>`** — `true` if any 2xx entry is a JSON Zod schema or an `AnyOfResponses` containing one.
 
@@ -264,9 +313,9 @@ These types are primarily consumed by HTTP client implementations.
 
 **`ClientRequestParams<TApiContract, TIsStreaming>`** — infers the request parameter object for a contract. Includes `pathParams`, `body`, `queryParams`, `headers` (required when the corresponding schema is defined), `pathPrefix` (always optional), and `streaming` (required for dual-mode contracts, forbidden otherwise).
 
-**`InferSseClientResponse<TApiContract>`** — discriminated union of `{ statusCode, headers, body }` for SSE mode. Success status codes yield `AsyncIterable<SseEventOf<...>>`; error codes yield the declared body type.
+**`InferSseClientResponse<TApiContract>`** — discriminated union of `{ statusCode, headers, body }` for SSE mode. Exact 2xx codes and the `'2xx'` range key yield `AsyncIterable<SseEventOf<...>>`; error codes, other range keys, and `'default'` yield the declared body type. `'default'` is split into a `SuccessfulHttpStatusCode` half and a non-success half.
 
-**`InferNonSseClientResponse<TApiContract>`** — same shape as above for non-SSE mode. Success status codes yield JSON / `string` / `Blob` / `null`; SSE entries are excluded (`never`).
+**`InferNonSseClientResponse<TApiContract>`** — same shape as above for non-SSE mode. Exact 2xx codes and the `'2xx'` range key yield JSON / `string` / `Blob` / `null` (SSE excluded); error codes, other range keys, and `'default'` yield the declared body type as-is. `'default'` is split the same way.
 
 **`DefaultStreaming<T>`** — `true` for SSE-only contracts, `false` for everything else.
 
@@ -324,7 +373,7 @@ getIsEmptyResponseExpected(deleteUser) // true
 getIsEmptyResponseExpected(getUser)    // false
 ```
 
-**`hasAnySuccessSseResponse`** — `true` when any 2xx entry is an SSE response (including inside `anyOfResponses`).
+**`hasAnySuccessSseResponse`** — `true` when any 2xx entry (exact code or `'2xx'` range key) is an SSE response (including inside `anyOfResponses`).
 
 ```ts
 import { hasAnySuccessSseResponse } from '@lokalise/api-contracts'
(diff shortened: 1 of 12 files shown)


packages/app/api-contracts/src/new/defineApiContract.ts
```diff
@@ -110,7 +110,7 @@ export const getSseSchemaByEventName = (routeConfig: ApiContract): SseSchemaByEv
 }
 
 export const hasAnySuccessSseResponse = (apiContract: ApiContract): boolean => {
-  for (const code of SUCCESSFUL_HTTP_STATUS_CODES) {
+  for (const code of [...SUCCESSFUL_HTTP_STATUS_CODES, '2xx' as const]) {
```
kibertoad: **P0 — `'default': sseResponse(...)` is misclassified as non-SSE.** This iterator now covers `'2xx'` but still ignores `'default'`. A contract declaring `{ default: sseResponse(...) }` will return `false` here → `DefaultStreaming = false` → the client picks the wrong consumption mode. The same gap exists at the type level in `inferTypes.ts:9` (`ExtractSuccessResponses`).

Either also detect `'default'` here (treating its success-half as SSE-capable, mirroring how `WildcardSseEntry`'s default branch already splits into a `SuccessfulHttpStatusCode` half), or explicitly document and reject SSE under `'default'`.

packages/app/api-contracts/src/new/inferTypes.ts
```diff
@@ -6,7 +6,7 @@ import type { ResponsesByStatusCode } from './contractResponse.ts'
 
 type ExtractSuccessResponses<T extends ResponsesByStatusCode> = ValueOf<
   T,
-  Extract<keyof T, SuccessfulHttpStatusCode>
+  Extract<keyof T, SuccessfulHttpStatusCode | '2xx'>
```
kibertoad: **P0/P3 — `'default'` not propagated to success-side classification.** This picks up `'2xx'` but not `'default'`. As a result, every helper flowing through `ExtractSuccessResponses` — `IsNoBodySuccessResponse`, `HasAnyJsonSuccessResponse`, `HasAnySseSuccessResponse`, `InferNonSseSuccessResponses`, `InferJsonSuccessResponses`, `InferSseSuccessResponses`, `ContractResponseMode`, `AvailableResponseModes` — silently misclassifies contracts whose only declared response is `default`.

Concrete consequence: `defineApiContract({ ..., responsesByStatusCode: { default: ContractNoBody } })` ⇒ `ExtractSuccessResponses = never` ⇒ `IsNoBodySuccessResponse = true` by vacuity (`[never] extends [ContractNoBody | undefined]`). That's technically what the docstring on line 13 says ("or no success status codes defined"), but it's true regardless of what `default` actually holds — which is misleading.

Pair this with the runtime gap in `defineApiContract.ts:113` (`hasAnySuccessSseResponse`).

packages/app/api-contracts/src/new/clientTypes.ts
```diff
@@ -73,38 +79,135 @@ type SseInferClientResponseBody<T> = Extract<InferClientResponseBody<T>, AsyncIt
  */
 type NonSseInferClientResponseBody<T> = Exclude<InferClientResponseBody<T>, AsyncIterable<unknown>>
 
+// Body helpers for wildcard response keys.
+// '2xx' maps to success mode (SSE-filtered or non-SSE-filtered); all other ranges and 'default' use
+// the full body union since those entries are always on the error side of captureAsError.
```
kibertoad: **P0 — Comment contradicts the code.** This block says "all other ranges and `'default'` use the full body union", but `WildcardSseEntry` at lines 111-119 below uses `SseInferClientResponseBody<...>` (AsyncIterable-only) for the `'default'` success half — i.e., the SSE-filtered narrowing, not the full body union.

The behavior is arguably correct (SSE-mode success half must be `AsyncIterable`), but the comment will mislead future maintainers. Either reword to clarify that `'default'`-success in SSE mode is still SSE-filtered, or move the explanation into the two `WildcardSseEntry` / `WildcardNonSseEntry` blocks where the asymmetric treatment actually happens.

packages/app/api-contracts/src/new/clientTypes.ts
```diff
@@ -73,38 +79,135 @@ type SseInferClientResponseBody<T> = Extract<InferClientResponseBody<T>, AsyncIt
...
+            ExactStatusCodes<TApiContract> | RangeStatusCodes<TApiContract>
+          >
+          headers: InferClientResponseHeaders<TApiContract>
+          body: InferClientResponseBody<NonNullable<TApiContract['responsesByStatusCode'][K]>>
+        }
+  : {
+      statusCode: Exclude<ExpandStatusRangeKey<K>, ExactStatusCodes<TApiContract>>
+      headers: InferClientResponseHeaders<TApiContract>
+      body: WildcardSseBody<NonNullable<TApiContract['responsesByStatusCode'][K]>, K>
+    }
```
kibertoad, line 137: **P2 — ~25-line near-duplicates.** `WildcardSseEntry` (107-132) and `WildcardNonSseEntry` (134-≈160) differ only in the body extractor (`SseInferClientResponseBody` vs `NonSseInferClientResponseBody`). TS HKT limits make full deduplication awkward, but at minimum the four copies of `Exclude<…, ExactStatusCodes<TApiContract> | RangeStatusCodes<TApiContract>>` could share intermediate type aliases (one for the unreached-success and one for the unreached-non-success status sets) to keep the structure visible and the diff small when this evolves.

packages/app/api-contracts/src/new/contractResponse.ts
```diff
@@ -206,21 +212,43 @@ export const resolveContractResponse = (
...
   responsesByStatusCode: ResponsesByStatusCode,
   statusCode: number,
   contentType: string | undefined,
   strictContentType: boolean,
 ): ResponseKind | null {
-  const schemaEntry = responsesByStatusCode[statusCode as HttpStatusCode]
+  const exactEntry = responsesByStatusCode[statusCode as HttpStatusCode]
+  if (exactEntry) {
+    return resolveContractResponse(exactEntry, contentType, strictContentType)
+  }
```
kibertoad, line 238: **P1 — Exact-match short-circuit isn't tested.** If the exact entry is e.g. `textResponse('text/csv')` but the server returns `application/json`, `resolveContractResponse` returns `null` and this function returns `null` immediately — it never falls through to the range or `default` entry. That's arguably correct ("exact code wins absolutely"), but it's an undocumented invariant and no test asserts it. Add an explicit assertion in `contractResponse.spec.ts` so a future refactor doesn't silently change the precedence.

packages/app/api-contracts/src/new/contractResponse.ts
```diff
@@ -206,21 +212,43 @@ export const resolveContractResponse = (
...
 }
 
+function getRangeKey(statusCode: number): HttpStatusCodeRange | null {
+  if (statusCode >= 100 && statusCode < 200) return '1xx'
+  if (statusCode >= 200 && statusCode < 300) return '2xx'
+  if (statusCode >= 300 && statusCode < 400) return '3xx'
+  if (statusCode >= 400 && statusCode < 500) return '4xx'
+  if (statusCode >= 500 && statusCode < 600) return '5xx'
+  return null
+}
```
kibertoad, line 222: **P1 — Boundary coverage thin.** The only "out of range" test uses `statusCode = 0`. Add explicit boundary cases — these are exactly where silent off-by-ones hide:

- 99 → `null`
- 100 → `'1xx'`
- 199 → `'1xx'`
- 200 → `'2xx'`
- 299 → `'2xx'`
- 300 → `'3xx'`
- 599 → `'5xx'`
- 600 → `null`

packages/app/api-contracts/src/HttpStatusCodes.ts
```diff
@@ -1,70 +1,44 @@
...
+export type ServerErrorHttpStatusCode = (typeof SERVER_ERROR_HTTP_STATUS_CODES)[number]
+
+export type HttpStatusCode =
+  | InformationalHttpStatusCode
+  | SuccessfulHttpStatusCode
+  | RedirectionHttpStatusCode
+  | ClientErrorHttpStatusCode
+  | ServerErrorHttpStatusCode
+
+export type HttpStatusCodeRange = '1xx' | '2xx' | '3xx' | '4xx' | '5xx'
```
kibertoad, line 42: **P2 — Public API surface grew without JSDoc.** `HttpStatusCodeRange`, `WildcardStatusCodeKey`, `ExpandStatusRangeKey`, and the four new code groups (`Informational`, `Redirection`, `ClientError`, `ServerError` — both the `*_HTTP_STATUS_CODES` arrays and the corresponding type aliases above) are now exported via `src/index.ts` but undocumented. A one-line JSDoc on each surfaces them in IntelliSense and documents intent for consumers.

packages/app/api-contracts/src/HttpStatusCodes.ts
```diff
@@ -1,70 +1,44 @@
...
+  ? InformationalHttpStatusCode
+  : K extends '2xx'
+    ? SuccessfulHttpStatusCode
+    : K extends '3xx'
+      ? RedirectionHttpStatusCode
+      : K extends '4xx'
+        ? ClientErrorHttpStatusCode
+        : K extends '5xx'
+          ? ServerErrorHttpStatusCode
+          : HttpStatusCode // 'default'
```
kibertoad: **P2 — Flatten the 6-level ternary.** Reads poorly and the `'default'` branch is tucked into a comment. A mapped-lookup form makes the relationship explicit and the `'default'` case as visible as the rest:

```ts
type RangeExpansion = {
  '1xx': InformationalHttpStatusCode
  '2xx': SuccessfulHttpStatusCode
  '3xx': RedirectionHttpStatusCode
  '4xx': ClientErrorHttpStatusCode
  '5xx': ServerErrorHttpStatusCode
  default: HttpStatusCode
}
export type ExpandStatusRangeKey<K extends WildcardStatusCodeKey> = RangeExpansion[K]
```

packages/app/api-contracts/src/HttpStatusCodes.ts
kibertoad, line 42: **P2 — Case-sensitivity divergence from OpenAPI.** OpenAPI 3.0 uses uppercase `'1XX'`–`'5XX'`; Fastify uses lowercase. The PR description says "matching OpenAPI / Fastify conventions" but only lowercase is accepted. Two reasonable choices:

(a) Accept both — widen the type union to `'1xx' | '1XX' | ...` and lowercase inside `getRangeKey`.
(b) Document the lowercase-only decision prominently in the README so users cross-referencing OpenAPI specs don't get confused when their keys silently fail to match.

(a) is more forgiving; (b) is simpler.

packages/app/api-contracts/src/new/contractResponse.spec.ts
```diff
@@ -287,4 +287,118 @@ describe('resolveResponseEntry', () => {
...
+    })
+
+    it('falls through to default when status code is outside all ranges', () => {
+      const schema = z.object({ error: z.string() })
+      expect(resolveResponseEntry({ default: schema }, 0, 'application/json', true)).toEqual({
+        kind: 'json',
+        schema,
+      })
+    })
+  })
```
kibertoad, line 461: **P1 — Test coverage gaps for default/range interactions.** A few cases worth adding:

1. `{ '4xx': ..., '5xx': ..., default: ... }` — assert routing for 4xx, 5xx, and a non-range code (e.g. 304) goes through the right branch and `default` is not invoked for 4xx/5xx codes. The current "range key takes precedence over default" test only exercises `'5xx'`.
2. `{ '2xx': ..., default: ... }` — symmetric to the existing `'4xx' + default` case (currently only the error-side combo is tested).
3. `default: anyOfResponses([sseResponse(...), zodSchema])` — verify the resolver picks the right kind from a composite `default` entry by content-type.
4. `resolveResponseEntry({ 200: textResponse('text/csv'), '2xx': someSchema }, 200, 'application/json', true)` — assert this returns `null` (i.e., does NOT fall through to `'2xx'` when the exact entry's content-type doesn't match). This is currently an undocumented invariant — see the inline comment on `contractResponse.ts`.

packages/app/api-contracts/src/new/inferTypes.spec.ts
```diff
@@ -165,6 +165,30 @@ describe('inferTypes', () => {
...
+      const contract = defineApiContract({
+        method: 'get',
+        pathResolver: () => '/test',
+        responsesByStatusCode: {
+          '4xx': sseResponse({ chunk: z.object({ delta: z.string() }) }),
+        },
+      })
+      type Result = HasAnySseSuccessResponse<(typeof contract)['responsesByStatusCode']>
+      expectTypeOf<Result>().toEqualTypeOf<false>()
+    })
```
kibertoad, line 207: **P1 — Only `HasAnySseSuccessResponse` got `'2xx'` coverage.** Every type in this file flows through the updated `ExtractSuccessResponses` (now `SuccessfulHttpStatusCode | '2xx'`), but the new tests only cover one helper.

At minimum, add one `'2xx'` test each for:

- `IsNoBodySuccessResponse` — with `'2xx': ContractNoBody` should be `true`
- `HasAnyJsonSuccessResponse` — with `'2xx': z.object(...)` should be `true`
- `InferNonSseSuccessResponses` — with `'2xx': z.object(...)` should produce that type
- `InferJsonSuccessResponses` — same
- `ContractResponseMode` — with `'2xx': sseResponse(...)` should be `'sse'`; with `'2xx': z.object(...)` should be `'non-sse'`
- `AvailableResponseModes` — should include `'sse'`/`'json'` accordingly

Without these, future regressions in `ExtractSuccessResponses` would only be caught by the one SSE detection test.

packages/app/api-contracts/README.md
```diff
@@ -176,7 +222,10 @@ defineApiContract({
   method: 'get' | 'post' | 'put' | 'patch' | 'delete',
   pathResolver: (pathParams) => string,
   responsesByStatusCode: {
-    [statusCode]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
+    // exact status codes
+    [statusCode: HttpStatusCode]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
+    // range keys (OpenAPI-style): '1xx' | '2xx' | '3xx' | '4xx' | '5xx' | 'default'
+    [rangeKey: WildcardStatusCodeKey]: z.ZodType | NoBodyResponse | TypedTextResponse | TypedBlobResponse | TypedSseResponse | AnyOfResponses
```
kibertoad: **P3 — Doc syntax is ambiguous.** `[statusCode: HttpStatusCode]: …` and `[rangeKey: WildcardStatusCodeKey]: …` are TS *index-signature* syntax, valid in type declarations but not inside a value-level object literal. Readers who copy this block will hit syntax errors.

Either mark this as schematic (e.g., a `// schematic — not valid TS` comment header on the block), or render it as a `type ResponsesByStatusCode = { ... }` declaration alongside the `defineApiContract({...})` example rather than inline within it.

kibertoad:
**P0 — `universal-testing-utils/ApiContractMockttpHelper` is incompatible with wildcard contracts.** Out of scope for this PR's diff (the file isn't touched), but it directly affects users adopting `'2xx'` / `'4xx'` / `default` in real contracts:

- `packages/app/universal-testing-utils/src/api-contracts/types.ts:55-58` — `StatusCodeBodyPair` uses `keyof T & HttpStatusCode`. A `{ '2xx': schema }`-only contract collapses the discriminated union to `never`, so the mock helper can't be configured against it at all.
- `packages/app/universal-testing-utils/src/api-contracts/ApiContractMockttpHelper.ts:54` — runtime does `responsesByStatusCode[params.responseStatus]` directly. Mocking status `200` against a `'2xx'`-only contract throws *"Specified responseStatus cannot be mapped with contract"*.

Fix is to mirror the exact→range→default lookup that `resolveResponseEntry` now does, and widen `StatusCodeBodyPair` to also enumerate wildcard keys (probably by reusing `ExpandStatusRangeKey`). Either fold into this PR or open a tracking issue and note the limitation in the changeset/README so adopters aren't surprised.

packages/app/api-contracts/src/HttpStatusCodes.ts
CatchMe2: I wouldn't complicate types by supporting both lower and uppercase. We will use lowercase only, fastify/openapi matching will be handled by `opinionated-machine`

packages/app/api-contracts/src/new/defineApiContract.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/new/inferTypes.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/new/clientTypes.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/new/contractResponse.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/new/contractResponse.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/HttpStatusCodes.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/new/contractResponse.spec.ts
CatchMe2: Fixed.

packages/app/api-contracts/src/new/inferTypes.spec.ts
CatchMe2: Fixed.

packages/app/api-contracts/README.md
CatchMe2: Fixed.

CatchMe2:
> **P0 — `universal-testing-utils/ApiContractMockttpHelper` is incompatible with wildcard contracts.** Out of scope for this PR's diff (the file isn't touched), but it directly affects users adopting `'2xx'` / `'4xx'` / `default` in real contracts:
> 
> * `packages/app/universal-testing-utils/src/api-contracts/types.ts:55-58` — `StatusCodeBodyPair` uses `keyof T & HttpStatusCode`. A `{ '2xx': schema }`-only contract collapses the discriminated union to `never`, so the mock helper can't be configured against it at all.
> * `packages/app/universal-testing-utils/src/api-contracts/ApiContractMockttpHelper.ts:54` — runtime does `responsesByStatusCode[params.responseStatus]` directly. Mocking status `200` against a `'2xx'`-only contract throws _"Specified responseStatus cannot be mapped with contract"_.
> 
> Fix is to mirror the exact→range→default lookup that `resolveResponseEntry` now does, and widen `StatusCodeBodyPair` to also enumerate wildcard keys (probably by reusing `ExpandStatusRangeKey`). Either fold into this PR or open a tracking issue and note the limitation in the changeset/README so adopters aren't surprised.

Let's handle it in a separate PR.

kibertoad (approved):

packages/app/api-contracts/src/HttpStatusCodes.ts
CatchMe2: Fixed.

PR #1007, opened 2026-05-26
Added odata-filter package

wawrzyniec-madej:
## Changes

Please describe

## Checklist

- [ ] I've added a changeset, or no published packages were changed

## AI Assistance Tracking

- [ ] **Yes, AI assisted with this PR**


packages/app/odata-filter/README.md
```diff
@@ -0,0 +1,149 @@
+# @lokalise/odata-filter
```
kibertoad, line 1: is this conceptually different from https://github.com/lokalise/shared-ts-libs/tree/main/packages/app/odata-mapper and does not belong there?

PR #1011, opened 2026-05-27
Add support for wildcard status in testing utils

CatchMe2:
## Changes

  - MockResponseParams now accepts any concrete numeric code covered by a range key ('2xx', 'default', …) — exact codes and range expansions are separate discriminants so the union stays unambiguous
  - resolveContractEntry replaces direct key lookup with exact → range → 'default' fallback, mirroring the runtime client
  - NoBodyResponse handled alongside ContractNoBody in both InferBodyParam and the no-body reply branch
  - Peer dep bumped to >=6.13.0; docs updated

## Checklist

## AI Assistance Tracking

### DIFF
--- a/packages/app/universal-testing-utils/src/api-contracts/docs.md
+++ b/packages/app/universal-testing-utils/src/api-contracts/docs.md
@@ -17,7 +17,7 @@ afterEach(() => mockServer.stop())
 
 ## mockResponse
 
-Registers a mock rule for the given contract. The `responseStatus` field is required and must match a key in the contract's `responsesByStatusCode` — it drives both which schema to use for validation and which response type to return.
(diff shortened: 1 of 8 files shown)


packages/app/universal-testing-utils/src/api-contracts/docs.md
```diff
@@ -17,7 +17,7 @@ afterEach(() => mockServer.stop())
 
 ## mockResponse
```
CarlosGamero, line 19: Should we add new features doc to readme?

CarlosGamero (approved):

packages/app/universal-testing-utils/src/api-contracts/docs.md
CatchMe2: Yeah, we could deprecate the previous class to handle mockttp and move /src/api-contracts/docs.md directly to the README. I'd do it in a separate PR, tho. WDYT?

CarlosGamero: Makes sense!

PR #1032, opened 2026-06-02
feat(fastify-api-contracts): add buildFastifyRouteByApiContract

kibertoad:
## What

Adds `buildFastifyRouteByApiContract` and `buildFastifyRouteHandlerByApiContract` — the `defineApiContract` counterparts of the existing `buildFastifyRoute` / `buildFastifyRouteHandler` (which target the deprecated `buildRestContract`/`buildGetRoute`/`buildPayloadRoute` builders).

This mirrors how the request injectors are split: `injectByApiContract` (new) vs `injectByContract` (deprecated).

## Behavior

- Derives the route **method**, **URL** (via `mapApiContractToPath`), **request schemas** (`body`/`params`/`querystring`/`headers`), `summary`/`description`/`tags` and **response schema** directly from the contract.
- Infers handler request/reply types from the contract:
  - GET/DELETE → handler without `req.body`
  - POST/PUT/PATCH → handler with `req.body` (typed `undefined` for `ContractNoBody`)
- Maps `responsesByStatusCode` into `schema.response`, keeping only entries that carry a JSON body. `ContractNoBody`, `textResponse`, `blobResponse` and `sseResponse` entries are skipped (no Zod serializer schema); `anyOfResponses` entries contribute the union of their JSON members.
- Exposes the contract on the route config (`req.routeOptions.config.apiContract`) and accepts the same optional metadata-to-route-options mapper as `buildFastifyRoute`.

## Notes

- `ExtendedFastifySchema` gains an optional `tags` field so contract tags flow through to the route schema (useful for swagger).
- Response-entry detection uses `isJsonResponse` (available in the declared `@lokalise/api-contracts` peer range) rather than the newer `isNoBodyResponse`.

## Tests / docs

- New `buildFastifyRouteByApiContract.spec.ts` covering GET/DELETE/POST/PUT/PATCH, `ContractNoBody` request/response, response-schema mapping, the metadata mapper, and type-level inference. Full package suite: 184 tests passing, no type errors.
- README updated with dedicated sections and cross-links between the new and deprecated builders.
- Changeset added (`minor`).

## AI Assistance Tracking


.changeset/fastify-route-by-api-contract.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lokalise/fastify-api-contracts": minor
+---
+
+Add `buildFastifyRouteByApiContract` and `buildFastifyRouteHandlerByApiContract` builders compatible with `defineApiContract` contracts. They derive the route method, URL, request schemas and response schema from the contract and infer the handler request/reply types, mirroring `buildFastifyRoute`/`buildFastifyRouteHandler` for the deprecated contract builders.
```
kibertoad: let's simplify naming - buildFastifyApiRoute and buildFastifyApiRouteHandler is much simpler

CatchMe2:
Do we even need these functions for APiContracts? Building api routes logic is already covered by opinionated-machine

.changeset/fastify-route-by-api-contract.md
kibertoad: Done in 8d3d143 — renamed to `buildFastifyApiRoute` / `buildFastifyApiRouteHandler` across the implementation, exports, tests, README and changeset (the source/spec files were renamed to match too).

kibertoad:
@CatchMe2 Good question. This package is the low-level primitive for turning a contract into a Fastify route definition, and `opinionated-machine` is a higher-level, opinionated layer on top — not every consumer adopts it, and the ones that do still benefit from a shared, well-typed builder underneath rather than re-implementing method/URL/schema derivation.

More concretely: `fastify-api-contracts` already ships `buildFastifyRoute`/`buildFastifyRouteHandler` for the **deprecated** `buildRestContract` style. `defineApiContract` is the replacement contract API, so these builders just give it the same first-class Fastify integration — without them, teams migrating off the deprecated contracts would lose the building block (or have to go all-in on opinionated-machine). Happy to revisit if we decide opinionated-machine should be the only supported path going forward.

CatchMe2:
I'm all aligned on moving api route definition to fastify-api-contracts and utilising it in opinionated-machine. One thing to note is that we have to add support for SSE responses, which means that we'd have to add @fastify/sse as peerDependency

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
```diff
@@ -0,0 +1,123 @@
...
+    schema.querystring = contract.requestQuerySchema
+  }
+  if (contract.requestHeaderSchema) {
+    schema.headers = contract.requestHeaderSchema
+  }
+  if (contract.requestBodySchema !== undefined && contract.requestBodySchema !== ContractNoBody) {
+    schema.body = contract.requestBodySchema
+  }
+
+  schema.response = buildResponseSchemas(contract)
```
kibertoad, line 133: `summary`, `description` and `tags` from the contract never reach the route schema. Both existing builders set them (`fastifyApiContracts.ts:156-158`, `fastifyRouteBuilder.ts:292-294`):

```ts
describe: apiContract.description,
description: apiContract.description,
summary: apiContract.summary,
```

Since v7 makes contract `summary` mandatory, every OpenAPI operation generated from this builder ends up without a summary, description or tags. Suggest returning `ExtendedFastifySchema` and mapping them through, same as the other two builders.

packages/app/fastify-api-contracts/src/types.ts
```diff
@@ -26,6 +26,7 @@ export type ExtendedFastifySchema = FastifySchema & {
   describe?: string
   description?: string
   summary?: string
+  tags?: readonly string[]
```
kibertoad: Nothing writes this field. `buildFastifyApiSchema` returns a plain `FastifySchema` and never sets `tags` (nor `summary`/`description`), and the two existing builders don't set it either. Either wire contract tags through in `buildFastifyApiSchema` or drop this addition.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,306 @@
...
+    onReconnect: _onReconnect,
+    ...fastifyOptions
+  } = options ?? {}
+
+  const eventSchemas = getSseSchemaByEventName(contract) ?? {}
+  const responseContentTypes = getContractResponseContentTypes(contract)
+  const contractMetadata = contractMetadataToRouteMapper?.(contract.metadata) ?? {}
+  const sseCapable = hasAnySseResponse(contract)
+
+  return {
```
kibertoad, line 354: The contract is not exposed on the route config, so `req.routeOptions.config.apiContract` (documented in the README's "Accessing the contract" section, and claimed in the PR description) does not work for routes built here.

`fastifyRouteBuilder.ts:270-280` merges it in:

```ts
const mergedConfig = routeMetadata?.config
  ? { ...routeMetadata.config, apiContract }
  : { apiContract }
```

Worth doing the same so hooks and handlers can reach the contract. Note this also interacts with the merge-order issue below: a `contractMetadataToRouteMapper` returning a `config` currently replaces the whole object rather than merging into it.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,306 @@
...
+    ...fastifyOptions,
+    ...contractMetadata,
```
kibertoad, line 355: Merge order contradicts the JSDoc on `contractMetadataToRouteMapper` (`apiHandlerTypes.ts:214`), which says its return value "is merged into the Fastify route options as a base". Spread last, contract metadata wins over options the caller passed explicitly, i.e. it is the override, not the base. Pick one and align code with docs.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -0,0 +1,166 @@
...
+      const session: SSESession<typeof eventSchemas, Context> = {
+        id: connectionId,
+        request,
+        reply,
+        context: startOptions?.context,
+        connectedAt: new Date(),
+        send,
+        isConnected: () => reply.sse.isConnected,
+        getStream: () => reply.sse.stream(),
+        sendStream: async (messages: AsyncIterable<SSEStreamMessage>) => {
```
kibertoad, line 213: The return value of `send` is discarded. `send` swallows write errors and returns `false`, so when the client disconnects mid-stream this loop keeps pulling from `messages` until the source is exhausted.

For the use case the README highlights (LLM token streams) that means paying for a full completion nobody receives, plus whatever side effects the generator performs.

```ts
sendStream: async (messages) => {
  for await (const message of messages) {
    const sent = await send(message.event, message.data, { id: message.id, retry: message.retry })
    if (!sent || !reply.sse.isConnected) break
  }
}
```

Breaking out of a `for await` calls the iterator's `return()`, so an `async function*` source gets to run its `finally` and clean up.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+  if (options?.heartbeat !== undefined) {
+    sseOptions.heartbeat = options.heartbeat
+  }
+
+  return sseOptions
+}
+
+/** Collects the response content-types a contract declares (a bare Zod schema is `application/json`). */
+function getContractResponseContentTypes(contract: ApiContract): string[] {
```
kibertoad, line 57: This unions the content types of **every** status code, including error responses, and that set is what `context.expectedContentType` is negotiated against.

For an SSE-only success with a JSON `404` the candidates are `['text/event-stream', 'application/json']`. A client sending `Accept: application/json` then gets `expectedContentType: 'application/json'` — a content type the 200 cannot produce — and the README's recommended branch (`if (expectedContentType === 'text/event-stream') ...`) silently takes the non-streaming path.

Negotiating over the success entry's content types would match what handlers actually branch on. If the current behavior is deliberate, it needs a caveat in the README next to the branching example.

kibertoad (commented):
Review of `buildFastifyApiRoute`. The type-level work is the strongest part: `ContentMapResults` making `contentType` required only for multi-media-type statuses, `ApiHandlerContext` conditionally adding `sse`, and `ApiHandlerReply` stripping fluent `.send()` chains are all well done and covered by `expectTypeOf` tests. Test coverage is genuinely thorough (90+ cases across all four body kinds, mixed content maps, header validation, hijack, SSE lifecycle, `Accept` edge cases).

Main things to look at, detailed inline:

1. `summary`/`description`/`tags` never reach the route schema, so every generated OpenAPI operation loses them (regression vs the existing builders), and the `tags` field added to `ExtendedFastifySchema` is dead.
2. The contract is not exposed on route config, so `req.routeOptions.config.apiContract` does not work for this builder.
3. `sendStream` discards `send`'s failure return, so a client disconnect mid-stream drains the whole source iterable.
4. `expectedContentType` negotiates across error responses too, which can name a content type the success status cannot produce.
5. `@fastify/sse` is a required peer but only type-imported and only needed for SSE contracts.
6. `contractMetadataToRouteMapper` merge order contradicts its own JSDoc.

Test gaps that map to the above: no assertion on `summary`/`description`/`tags` in the schema, no client-disconnect-mid-`sendStream` case, no case for `onConnect`/`onReconnect` throwing.

Security: nothing concerning. Bodies go through the Zod serializer, SSE event data is validated per event before write, blob bodies are passed natively with a contract-declared content type.

Also: the PR description is stale. It describes `buildFastifyRouteByApiContract`/`buildFastifyRouteHandlerByApiContract`, `anyOfResponses`, `textResponse` and `isJsonResponse` vs `isNoBodyResponse` — none of which match the shipped diff. Worth rewriting from the changeset before merge.

packages/app/fastify-api-contracts/package.json
```diff
@@ -43,19 +43,23 @@
         "postversion": "biome check --write package.json"
     },
     "dependencies": {
-        "light-my-request": "^6.6.0"
+        "light-my-request": "^6.6.0",
+        "negotiator": "^1.0.0"
     },
     "peerDependencies": {
-        "@lokalise/api-contracts": ">=6.5.1",
+        "@fastify/sse": ">=0.5.0",
```
kibertoad, line 50: This should be an optional peer. The README says the plugin is only needed for contracts that declare an SSE response, and the only reference is `import type { SSERouteOptions } from '@fastify/sse'`, which is erased at build time. As a required peer, every consumer with JSON-only contracts gets an unmet-peer warning for a package they will never install.

```json
"peerDependenciesMeta": {
    "@fastify/sse": { "optional": true }
}
```

packages/app/fastify-api-contracts/package.json
```diff
@@ -43,19 +43,23 @@
...
+        "@lokalise/api-contracts": ">=7.0.0",
```
kibertoad, line 51: Raising the peer floor from `>=6.5.1` to `>=7.0.0` breaks consumers still on v6 — they can no longer take any release of this package. The changeset marks it `minor`; arguably this is a `major` for `@lokalise/fastify-api-contracts`.

packages/app/fastify-api-contracts/package.json
```diff
@@ -43,19 +43,23 @@
...
     },
     "devDependencies": {
         "@biomejs/biome": "^2.5.0",
+        "@fastify/sse": "^0.5.0",
         "@lokalise/api-contracts": "workspace:^",
         "@lokalise/biome-config": "workspace:^",
         "@lokalise/tsconfig": "workspace:^",
+        "@types/negotiator": "^0.6.4",
         "@vitest/coverage-v8": "^4.1.9",
         "fastify-type-provider-zod": "^6.1.0",
```
kibertoad: The changeset states that generating a spec from content maps requires `fastify-type-provider-zod >= 7`, but the devDependency here is `^6.1.0`, so the whole content-map schema path (including `z.file()` for `blobBody()` and the SSE envelope unions) is only ever exercised against v6 in CI. Worth bumping the devDependency, or at least noting that the v7 path is untested.

Separately: `fastify-type-provider-zod` is required at runtime by this module (per the README) but is not declared as a peer dependency at all. Pre-existing, but this PR makes the version constraint load-bearing.

packages/app/fastify-api-contracts/src/api-contracts/apiHandlerTypes.ts
```diff
@@ -0,0 +1,222 @@
...
+ * the contract provides (`method`, `url`, `schema`, `handler`, `sse`), plus the SSE lifecycle
+ * options (`onConnect`/`onClose`/`onReconnect`, …) that apply only to SSE-capable contracts.
+ */
+export type ApiRouteOptions = Omit<RouteOptions, 'method' | 'url' | 'schema' | 'handler' | 'sse'> &
+  FastifySSERouteOptions & {
+    /**
+     * Maps contract metadata to additional Fastify route options.
+     *
+     * Called with the contract's `metadata` field; its return value is merged into
+     * the Fastify route options as a base — useful for cross-cutting concerns (auth,
```
kibertoad: "merged into the Fastify route options as a base" does not match the implementation — `buildFastifyApiRoute.ts:287` spreads `contractMetadata` after `fastifyOptions`, so metadata overrides explicitly-passed options rather than acting as a base.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,306 @@
...
+ * unambiguous: a `body: null` on an `allowNoBody` entry, a bare Zod schema
+ * (`application/json`), or a content map's first declared content-type.
+ */
+function resolveResponseRepresentation(
+  contract: ApiContract,
+  { status, contentType, body }: ApiHandlerResult,
+): ResolvedApiResponse {
+  const entry = contract.responsesByStatusCode[status as HttpStatusCode]
+  if (!entry) {
+    throw new Error(`Contract does not declare a response for status ${status}.`)
```
kibertoad, line 154: These contract-mismatch paths throw bare `Error` (also lines 125, 142 and 237), while `validateApiResponseHeaders` and the SSE event validation in `sseUtils.ts` throw `InternalError` with an `errorCode`. They are all the same class of programmer error surfaced as a 500. Using `InternalError` consistently gives them structured error codes in logs, matching the rest of the package.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  if (isJsonResponse(entry)) {
+    return { status, body, contentType: 'application/json', isSse: false }
+  }
+
+  if (entry.allowNoBody && body === null) {
+    return { status, body, contentType: null, isSse: false }
+  }
+
+  const descriptorEntry = Object.entries(entry.content ?? {})[0]
```
kibertoad, line 175: When `contentType` is omitted and the content map declares several media types, this silently picks the first one. The type system prevents that from TS callers, but a JS consumer (or anything that went through an `as any`) gets a wrong representation and a wrong serializer instead of an error. Cheap guard:

```ts
const entries = Object.entries(entry.content ?? {})
if (entries.length > 1) throw new InternalError({ ... }) // contentType is required here
```

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  // Response body validation is handled by the `fastify-type-provider-zod` serializer
+  // compiler (a required dependency), which throws a 500 when a JSON body doesn't match the
+  // contract's schema for this status code and content-type (raw and SSE bodies bypass the
+  // serializer). Response headers are not covered by the serializer, so they are validated
+  // explicitly here.
+  validateApiResponseHeaders(contract, reply)
+
+  if (reply.sent) {
```
kibertoad: `validateApiResponseHeaders` runs before this check, so on a hijacked reply a header-validation failure throws after the response has already gone out — Fastify's error handler then tries to send again. Moving the `reply.sent` early return above the validation call avoids that.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -0,0 +1,166 @@
...
+          }
+        },
+        close: () => {
+          closedByServer = true
+          reply.sse.close()
+        },
+      }
+
+      if (options?.onConnect) {
+        void Promise.resolve(options.onConnect(session)).catch(() => {})
```
kibertoad: `.catch(() => {})` discards user errors from `onConnect` (and the same pattern for `onClose` just below) with no trace at all. At minimum log them: `request.log.error({ err }, 'SSE onConnect hook failed')`. Silent-drop makes a broken hook effectively invisible in production.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -0,0 +1,166 @@
...
+          void Promise.resolve(onClose(session, closedByServer ? 'server' : 'client')).catch(
+            () => {},
+          )
+        })
+      }
+
+      if (options?.onReconnect && reply.sse.lastEventId) {
+        const onReconnect = options.onReconnect
+        const lastEventId = reply.sse.lastEventId
+        void reply.sse.replay(async () => {
```
kibertoad: Unlike the `onConnect`/`onClose` paths, this `void`-ed promise has no `.catch`. If `onReconnect` (or a `reply.sse.send` inside the replay loop) rejects and `reply.sse.replay` doesn't guard internally, that's an unhandled rejection. Add a `.catch` that logs, same as the other hooks.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -0,0 +1,166 @@
...
+ * the list win ties (e.g. under a full wildcard).
+ *
+ * Returns `null` when the request carries no `Accept` header or none of the candidates
+ * is acceptable — the caller decides the fallback.
+ *
+ * @param request - The Fastify request whose `Accept` header drives the negotiation
+ * @param contentTypes - Candidate response content-types, in server preference order
+ * @returns The preferred candidate, or `null` when there is no acceptable match
+ */
+export function determineResponseContentType<TContentType extends string>(
```
kibertoad, line 28: This is called unconditionally in `handleApiRoute` for every request, so every plain JSON route allocates a `Negotiator` and parses the `Accept` header for a value the handler will usually never read. Making `expectedContentType` a lazy getter on the context object would make it free for the common case:

```ts
const context = {
  get expectedContentType() { return determineResponseContentType(request, responseContentTypes) },
  ...
}
```

(Memoize if a handler may read it more than once.)

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Well, consumer can fake any argument that can lead to runtime errors. We should trust that passed data is meaningful or be consistent and apply runtime type check everywhere (which doesn't feel right, we have TS for it)

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: This throw is only to satisfy TS. It shouldn't happen in real application unless someone intentionally pass invalid result with like `return {} as any`. I'd keep it as is.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
CatchMe2: Good catch, fixed.

packages/app/fastify-api-contracts/src/types.ts
CatchMe2: Good point, dropped.

packages/app/fastify-api-contracts/package.json
CatchMe2: Good point, fixed.

packages/app/fastify-api-contracts/package.json
CatchMe2: Good point, fixed.

packages/app/fastify-api-contracts/package.json
CatchMe2: Good point, fixed.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Good point, fixed.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Good point, fixed.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Good idea. Added

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Added docs about it.

packages/app/fastify-api-contracts/src/api-contracts/apiHandlerTypes.ts
CatchMe2: Adjusted

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+ * unambiguous: a `body: null` on an `allowNoBody` entry, a bare Zod schema
+ * (`application/json`), or a content map's first declared content-type.
+ */
+function resolveResponseRepresentation(
+  contract: ApiContract,
+  { status, contentType, body }: ApiHandlerResult,
+): ResolvedApiResponse {
+  const entry = contract.responsesByStatusCode[status as HttpStatusCode]
+  if (!entry) {
+    throw new Error(`Contract does not declare a response for status ${status}.`)
```
kibertoad, line 154: **Blocking: wildcard status keys are unsupported at runtime.**

`ResponsesByStatusCode` is `Partial<Record<HttpStatusCode | WildcardStatusCodeKey, ...>>`, so a contract may legally declare `{ 200: userSchema, '4xx': errorSchema }`. This lookup is exact, so a handler returning `{ status: 404 }` against a `'4xx'` entry falls into this throw and the client gets a 500.

`@lokalise/api-contracts` already exports `resolveResponseEntry`, which implements exact -> range (`'4xx'`) -> `'default'` precedence. Either use it here, or reject wildcard keys explicitly in `buildFastifyApiRoute` with a clear message and document the limitation.

No test in any of the three new spec files covers a wildcard status key.

packages/app/fastify-api-contracts/src/api-contracts/apiHandlerTypes.ts
```diff
@@ -0,0 +1,226 @@
...
+          ? ContentMapResults<TStatusCode, TContent>
+          : never)
+      | (TEntry extends { allowNoBody: true } ? { status: TStatusCode; body: null } : never)
+
+/**
+ * Discriminated union of `{ status, contentType?, body }` results for every response a
+ * contract declares. `contentType` exists only for content-map responses — required (and a
+ * discriminant) when a status declares several media types, optional when it declares one.
+ */
+export type InferApiHandlerResult<TApiContract extends ApiContract> = {
```
kibertoad, line 102: **Blocking (same root cause as the runtime wildcard issue).**

This maps over `keyof TApiContract['responsesByStatusCode']`, which includes wildcard keys. For a contract declaring `'4xx': errorSchema`, the inferred handler result is `{ status: '4xx', body: ... }` and returning `{ status: 404, body: ... }` becomes a type error. That is backwards from what a handler author wants.

Wildcard keys need expanding to their concrete status union (`ExpandStatusRangeKey` exists in `api-contracts`) or excluding here with a documented restriction.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
```diff
@@ -0,0 +1,130 @@
...
+ * per-media-type response schema (`{ content: { '<mediaType>': { schema } } }`): JSON
+ * descriptors keep their Zod schema — selected by the response `content-type`, both for
+ * serialization and in the OpenAPI spec — while `blobBody()` gets a binary placeholder and
+ * `sseBody()` a union of its event envelopes. A no-body entry (`noBodyResponse()`) becomes
+ * a `z.null()` schema, describing a body-less response.
+ */
+function buildResponseSchemas(contract: ApiContract): Record<string, FastifyResponseSchema> {
+  const schemas: Record<string, FastifyResponseSchema> = {}
+
+  for (const [statusCode, entry] of Object.entries(contract.responsesByStatusCode)) {
```
kibertoad, line 73: Third facet of the wildcard problem: this passes contract keys through verbatim, so a `'default'` entry becomes `schema.response['default']`. Fastify's serializer recognises `'2xx'`-style range keys but not `'default'` (that is an OpenAPI convention), so such an entry silently never applies.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+  }
+
+  if (contentType !== undefined) {
+    const descriptor = isContentResponseEntry(entry) ? entry.content?.[contentType] : undefined
+    if (descriptor === undefined) {
+      throw new Error(
```
kibertoad, line 160: **Blocking: internal messages leak to clients.**

This and the throws at lines 119, 142 and 237 are plain `Error`. Fastify's default error handler serialises `err.message` into the 500 body, so a client sees `Contract does not declare content-type "application/xml" for status 200.`

`validateApiResponseHeaders` a few lines up already uses `InternalError` from `@lokalise/node-core` with an `errorCode`, which is the package convention and gets masked by the Lokalise error handler. Make all four consistent.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+ */
+async function sendResponse(
+  contract: ApiContract,
+  reply: FastifyReply,
+  { status, contentType, body }: ResolvedApiResponse,
+): Promise<void> {
+  // (@fastify/sse commits its `text/event-stream` headers lazily on stream start, so nothing
+  // is pre-set here on the non-streaming path of an SSE-capable route.)
+  if (contentType !== null && reply.getHeader('content-type') === undefined) {
+    reply.type(contentType)
```
kibertoad, line 205: **The `reply.sent` guard is 10 lines too late.**

After `reply.hijack()`, `reply.sent` is true and headers set on `reply.raw` are invisible to `reply.getHeader()`, so `reply.type()` runs against an already-sent reply. `validateApiResponseHeaders` then runs on it too.

The hijack test (`buildFastifyApiRoute.spec.ts:491`) passes only because it sets `content-type` on `raw` and asserts on the final body. Move the `if (reply.sent) return` check to the top of `sendResponse`: there is nothing to set or validate on a reply that has already gone out.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+  }
+  if (options?.heartbeat !== undefined) {
+    sseOptions.heartbeat = options.heartbeat
+  }
+
+  return sseOptions
+}
+
+/** Collects the response content-types a contract declares (a bare Zod schema is `application/json`). */
+function getContractResponseContentTypes(contract: ApiContract): string[] {
```
kibertoad, line 57: **`expectedContentType` negotiates across error responses.**

This collects content types from every status code, so a contract with `200: sseResponse(...)` and `400: z.object(...)` offers `application/json` as a candidate even though no success path can produce it. The JSDoc on `ApiHandlerContext` acknowledges this, but it makes the value awkward for its main stated use (branching on which representation to emit). `api-contracts` already has `hasAnySuccessSseResponse`, so the success-status-only notion exists upstream. Negotiating over 2xx content types only would be more predictable.

Separately: candidate order here comes from JS object key iteration (numeric keys ascending), and that order decides the winner under `Accept: */*`, which is what most non-browser clients send. Worth stating explicitly in the docs since it is currently implicit.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+  sseCapable,
+  request,
+  reply,
+}: HandleApiRouteParams): Promise<void> {
+  const apiSSEContext = sseCapable
+    ? buildApiSSEContext(request, reply, eventSchemas, options)
+    : undefined
+
+  const context = {
+    expectedContentType: determineResponseContentType(request, responseContentTypes),
```
kibertoad: **Every request pays for `Accept` parsing, including JSON-only routes.**

This constructs a `Negotiator` and parses the `Accept` header on every request even when `responseContentTypes` has a single entry and the answer is fixed. Either short-circuit when `responseContentTypes.length <= 1`, or make `expectedContentType` a lazy getter on the context object so handlers that never read it pay nothing.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -0,0 +1,175 @@
...
+      started = true
+      sessionMode = mode
+
+      if (mode === 'keepAlive') {
+        reply.sse.keepAlive()
+      }
+
+      // sendHeaders() calls writeHead(200) but only queues headers in the buffer.
+      // flushHeaders() forces them onto the wire so the client's fetch() returns.
+      reply.sse.sendHeaders()
```
kibertoad, line 171: **No guard for a missing `@fastify/sse` registration.**

`@fastify/sse` is an optional peer, and Fastify silently ignores the unknown `sse` route option when the plugin is not registered. The first request to an SSE contract then dies here with `Cannot read properties of undefined (reading 'sendHeaders')`.

A check in `buildApiSSEContext` (or at route build time via an `onRoute` hook) throwing something like "contract declares an SSE response but @fastify/sse is not registered" would save real debugging time.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+    contractMetadataToRouteMapper,
+    serializer: _serializer,
+    heartbeat: _heartbeat,
+    onConnect: _onConnect,
+    onClose: _onClose,
+    onReconnect: _onReconnect,
+    ...fastifyOptions
+  } = options ?? {}
+
+  const eventSchemas = getSseSchemaByEventName(contract) ?? {}
```
kibertoad: `getSseSchemaByEventName` merges every SSE event map across all statuses with `Object.assign` (last wins), and `session.send` validates against that flat map. Two SSE statuses declaring the same event name with different schemas validate against the wrong one, and an event declared only on status A is accepted while streaming status B.

The flattening originates upstream in `api-contracts`, but this builder is what makes it observable. A JSDoc note here at minimum.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+    for (const contentType of Object.keys(entry.content)) {
+      contentTypes.add(contentType)
+    }
+  }
+
+  return [...contentTypes]
+}
+
+/** True when any response entry of the contract (any status code) declares an SSE body. */
+export const hasAnySseResponse = (apiContract: ApiContract): boolean =>
```
kibertoad: `hasAnySseResponse` is exported from this module but not re-exported from `api-contracts/index.ts`, so it is public surface only for the spec. Either export it properly or drop the `export`.

It also nearly duplicates `hasAnySuccessSseResponse` in `api-contracts` with different semantics (any status vs success statuses). A one-line comment on why "any status" is right here would help.

Same question for `buildFastifyApiSchema`, which the spec imports directly but `index.ts` does not re-export.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+      await session.sendStream(resolved.body as AsyncIterable<SSEStreamMessage>)
+      apiSSEContext.markHandlerDone()
+      return
+    }
+    // Any other status/body is sent as a regular HTTP response.
+    await sendResponse(contract, reply, resolved)
+    return
+  }
+
+  throw new Error(
```
kibertoad: This message fires on non-SSE routes too, where `sse.start()` does not exist and the advice is misleading. Branch the message on `sseCapable`.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+      errorCode: 'RESPONSE_HEADERS_VALIDATION_FAILED',
+      details: { validationError: result.error.message },
+    })
+  }
+}
+
+/** The runtime shape of a handler's return value (`InferApiHandlerResult` erased of its generics). */
+type ApiHandlerResult = { status: number; contentType?: string; body: unknown }
+
+function isApiHandlerResult(value: unknown): value is ApiHandlerResult {
```
kibertoad, line 109: `'status' in value` is loose enough that a returned `FastifyReply` passes (`reply.status` is a method), and `resolveResponseRepresentation` then throws a confusing contract error. Tightening to `typeof (value as ApiHandlerResult).status === 'number'` is cheap.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
```diff
@@ -0,0 +1,322 @@
...
+    return { status, body, contentType: 'application/json', isSse: false }
+  }
+
+  if (entry.allowNoBody && body === null) {
+    return { status, body, contentType: null, isSse: false }
+  }
+
+  const descriptorEntry = Object.entries(entry.content ?? {})[0]
+  if (descriptorEntry === undefined) {
+    throw new Error(`Contract does not declare a content-type for status ${status}.`)
```
kibertoad, line 177: Minor: an explicit `contentType: 'application/json'` against a bare-Zod entry is rejected (asserted as a 500 at `buildFastifyApiRoute.spec.ts:390`). Unreachable at type level, but accepting it when it matches `application/json` would be friendlier for JS consumers and costs one condition.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -0,0 +1,175 @@
...
+            }
+          }
+        },
+        close: () => {
+          closedByServer = true
+          reply.sse.close()
+        },
+      }
+
+      if (options?.onConnect) {
```
kibertoad, line 233: `onConnect` and `onClose` rejections are swallowed with a bare `.catch(() => {})`. The tests assert the swallowing, so it is deliberate, but a silently dropped user hook error is painful to debug. Consider `request.log.error` instead of discarding.

.changeset/fastify-route-by-api-contract.md
```diff
@@ -0,0 +1,9 @@
...
...est, contentTypes)` helper is also exported for custom candidate lists). For advanced streaming (keep-alive, lifecycle hooks, reconnection) `context.sse.start(...)` can drive the session imperatively instead. The route method, URL, request schemas and response schemas are derived from the contract; a `blobBody()` descriptor accepts a `string`/`Buffer`/`Readable` body sent with the declared media type, and a `noBodyResponse()` status requires `body: null`. Every response entry is mapped into `schema.response` so the full contract shows up in a generated OpenAPI spec: content-map entries become Fastify per-media-type response schemas (JSON descriptors keep their Zod schema and drive content-type-aware serialization; `blobBody()` maps to a binary string, `sseBody()` to the union of its event envelopes `{ event, data, id?, retry? }` per the OpenAPI convention for `text/event-stream`), and `noBodyResponse()` maps to `z.null()`, rendered by `@fastify/swagger` as a body-less response (spec generation from content maps requires `fastify-type-provider-zod` >= 7). SSE lifecycle hooks (`onConnect`, `onClose`, `onReconnect`, `serializer`, `heartbeat`) are configurable via the options argument.
```
kibertoad: This is a single ~400 word paragraph and lands verbatim in `CHANGELOG.md`. Split it into a short summary line plus bullets.

The filename (`fastify-route-by-api-contract`) also no longer matches the API it describes (`buildFastifyApiRoute`).

kibertoad (commented):
Review of `buildFastifyApiRoute`. Design is solid: `ApiHandlerReply` making `.send()` a compile error, the `contentType` discriminant for multi-media-type statuses, and the `sendStream` break-on-failed-write so `async function*` sources run their `finally` are all well thought through. Test coverage is unusually thorough.

Findings posted inline. Two I would fix before merge:

1. Wildcard status keys (`'4xx'`, `'2xx'`, `'default'`) are unsupported at both type and runtime level, and untested.
2. Internal error messages are thrown as plain `Error` and leak into 500 response bodies.

Process notes:

- The PR description is stale. It describes `buildFastifyRouteByApiContract` / `buildFastifyRouteHandlerByApiContract`, a `minor` changeset, and an `ExtendedFastifySchema` gaining `tags`. None of that is in this diff.
- `major` is the correct bump: the `@lokalise/api-contracts` peer goes `>=6.5.1` to `>=7.0.0`, `fastify-type-provider-zod` becomes a required peer, and `FastifyContextConfig.apiContract` widens to `CommonRouteDefinition | ApiContract`, which breaks consumers reading it as the narrower type.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Good point, improved.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Good point, done.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: The same answer as to https://github.com/lokalise/shared-ts-libs/pull/1032#discussion_r3657966285

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Good point. Fixed.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/apiHandlerTypes.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
CatchMe2: Fixed

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Fixed by forcing client to set statusCode & contentType when a single contract has multiple sseResponses

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Fixed.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Good point, I've changed it to negotiate on success responses only.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: These are internal functions, exported only for unit tests.

.changeset/fastify-route-by-api-contract.md
CatchMe2: Agree, improved.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: After changing accept header negotiation logic, `hasAnySseResponse` was no longer needed and was removed.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: Added `typeof value.status === 'number'` to this type guard.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.ts
CatchMe2: I think that we should trust TS, no need to add additional guards here or complicate types, if a single contentType is allowed by contract, there is no need in setting it explicitly).

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Fixed.

CatchMe2 (approved):

PR #1034, opened 2026-06-05
feat(opentelemetry-fastify-bootstrap): join Datadog inferred-service entities via db.namespace stamping

ppletscher:
## Changes

Adds an optional `dbNamespaceBySystem` option to `initOpenTelemetry`. When provided, the bootstrap wraps **only the Datadog-bound OTLP trace exporter** in a `DbNamespaceSpanExporter`. For each outbound span carrying `db.system: <key>` (falling back to `peer.db.system` when `db.system` is absent), that exporter adds to its **own export payload**:

- `db.namespace: <mapped-value>` — the OTel-canonical attribute. Datadog derives `peer.db.name` from it when constructing inferred-service entity keys.

We only ever set the conventional `db.namespace`; we never write any `peer.*` tag — Datadog derives `peer.db.name`/`peer.db.system` itself from the short-form `db.*` attributes.

The option is fully opt-in (no default mapping, exporter unwrapped when omitted) and only affects spans whose `db.system` matches a key in the supplied map. An existing non-empty `db.namespace` is never replaced.

```ts
initOpenTelemetry({
  dbNamespaceBySystem: { elasticsearch: 'lokalise' },
})
```

### Exporter, not a span processor

`db.namespace: lokalise` is a Datadog entity-keying heuristic, not the span's true OTel namespace. Following review feedback from Igor and Ervin, this **does not mutate the span**: the value is added only to the Datadog exporter's payload via a non-mutating `Proxy` view of the `ReadableSpan`. The single shared span object seen by every other processor/exporter — console, user-supplied processors, future OTel-native tooling — is left untouched, so nothing else reads `lokalise` as the real namespace. Because the original is never written to, this also can't be broken by a future SDK that freezes span attributes, and the previous `onEnd` mutation hardening (disarm-on-throw, registration-order pin) is gone.

`DbNamespaceSpanExporter` is also exported for wrapping an exporter directly.

## Problem this tries to solve

Screenshot 2026-06-05 at 18 14 18

Node services on this bootstrap (e.g. tm-service, autopilot's `backend-translation-storage`) use the v8 `@elastic/elasticsearch` client. Its `@elastic/transport` library emits OTel spans natively but only sets `db.system: elasticsearch` — never `db.namespace`. The ES nodes are reached by raw IP, so Datadog's APM service catalog can't classify the downstream from `peer.hostname` alone:

- `peer.hostname` resolves to an IP (e.g. `10.22.38.72`).
- Datadog's classifier flags the span (it stamps a verbatim `issue_type: pod_like_hostname` tag — the name is misleading here since these are VMs, but it's the same DD heuristic) and routes the span into the synthetic `blocked-ip-address` service.
- Result: TM/autopilot's ES traffic doesn't link to the existing `(peer.db.system:elasticsearch, peer.db.name:lokalise)` inferred-service entity that the PHP `expert-*` services feed (~2k spans/h between them) — so the dependency map shows an unidentified downstream instead of the real ES cluster.

The PHP services dodge this because their custom in-house instrumentation (`lokalise-main/services/app/src/Apm/Implementation/OtelTracer.php:212`) sets `db.namespace` explicitly, with the value hard-coded as the literal string `lokalise` in `LoggingTransport.php:160` — a documented dummy constant whose only job is to produce a stable Datadog entity. There is no equivalent step on the Node side, so the spans drop into the IP bucket.

This PR fills that gap, matching what `OtelTracer.php` already does — but scoped to the Datadog export payload rather than the span itself.

## Verification

Verified end-to-end against the real `@elastic/elasticsearch` v8 client (`@elastic/transport` 8.10.1, as used by tm-service) pointed at a live ES 8.16.2, with `getNodeAutoInstrumentations()` + the wrapped exporter:

- **Export payload → Datadog:** `db.system=elasticsearch`, `db.namespace=lokalise` on all ES spans (`indices.create/delete`, `index`, `search`); no `peer.*` written.
- **Shared span → every other consumer:** `db.namespace` absent — the source span is untouched.

This confirms the premise: the ES auto-instrumentation sets `db.system` but not `db.namespace`, and the exporter fills the gap only in its own payload. Datadog's `db.namespace → peer.db.name` derivation isn't independently verified here — the plan is to release, ship in tm-service, and confirm in the live dependency map.

## Alternative Solutions

- A dedicated `@opentelemetry/instrumentation-elasticsearch` install that already sets `db.namespace`, if there's one we trust (I didn't find a maintained one for ES v8).
- Configuring peer-service inference on the Datadog OTel collector instead of in the SDK — centralised but outside this team's control.

## Checklist

## AI Assistance Tracking


ervinb:
Datadog adds the `peer.*` prefix, so we can use the short format seen on the right side of the table eg. `db.name`. This would also be more vendor-neutral.

image


https://docs.datadoghq.com/tracing/services/inferred_services/?tab=agentv7600#peer-tags

packages/app/opentelemetry-fastify-bootstrap/package.json
```diff
@@ -55,6 +55,7 @@
         "@biomejs/biome": "^2.5.0",
         "@lokalise/biome-config": "workspace:^",
         "@lokalise/tsconfig": "workspace:^",
+        "@opentelemetry/api": "^1.9.0",
```
kibertoad: I'd suggest setting this to "^1.9.1" instead to avoid affecting previous lockfile resolution

kibertoad (approved):

ervinb (approved):

PR #1035, opened 2026-06-08
fix(background-jobs-common): demote bullmq control-flow errors to debug

Drodevbar:
## Summary

`BackgroundJobProcessorMonitor.jobAttemptError` was treating every error thrown from `processInternal` as a job failure and logging it at `error` level with `"<jobName> try failed"`. This included BullMQ's cooperative control-flow errors - `DelayedError`, `WaitingChildrenError`, `RateLimitError` - which a processor throws after `moveToDelayed` / `moveToWaitingChildren` / `rateLimit` to hand the job back to BullMQ.

These aren't failures, so logging them at `error` level was misleading (the package itself does this in its own barrier path, so any barrier-driven delay was producing a noisy error log in prod).

  ## Changes

  - Added `isBullmqControlFlowError` helper in `errors/utils.ts` (matches the existing `error.name === X.name` pattern used by `isUnrecoverableJobError`).
  - `BackgroundJobProcessorMonitor.jobAttemptError` now detects control-flow errors and logs `"<jobName> deferred via <ErrorName>"` at `debug` level instead of the `error` log. Real failures still log at `error` unchanged.

  ### Why `debug` and not silent / `warn` / `info`?

  - Silent is hard to debug later when investigating "why is this job stuck in delayed?".
  - `warn` would cry wolf - these are normal control flow.
  - `info` would add steady noise on healthy rate-limit / retry-heavy workers.
  - `debug` matches the existing precedent in `AbstractBackgroundJobProcessor.barrier` (`"Did not pass the barrier"` is also debug).

  ## Test plan

  - [x] New unit tests in `BackgroundJobProcessorMonitor.spec.ts` asserting `logger.error` is NOT called and `logger.debug` IS called with the expected payload for each of
  `DelayedError`, `WaitingChildrenError`, `RateLimitError`.
  - [x] Existing `jobAttemptError` "should log error" test still passes (real errors still log at `error`).
  - [x] Full package suite green (211 tests passing).
  - [x] Barrier integration specs run clean - no more `"try failed"` error logs for delayed jobs.
  - [x] `pnpm run lint` (biome + tsc) clean.

## AI Assistance Tracking

### DIFF
--- a/.changeset/bullmq-control-flow-error-log.md
+++ b/.changeset/bullmq-control-flow-error-log.md
@@ -1,0 +1,5 @@
+---
+"@lokalise/background-jobs-common": patch
+---
+
+Stop logging BullMQ control-flow errors (`DelayedError`, `WaitingChildrenError`, `RateLimitError`) as job attempt failures. These errors are cooperative signals a processor throws after `moveToDelayed`/`moveToWaitingChildren`/`rateLimit` to hand the job back to BullMQ, not real failures. They now emit a `"<jobName> deferred via <ErrorName>"` debug log instead of the `"<jobName> try failed"` error log.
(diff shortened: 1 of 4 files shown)


CarlosGamero (approved):
🚀

PR #1055, opened 2026-06-11
Fix api contract path types

CatchMe2:
## Changes

This PR fixes three compiler-verified type-safety holes in the new contract API (@lokalise/api-contracts) found during a detailed review, plus one deprecation. All changes are patch-level; runtime
  behavior is unchanged — every fix makes the type level tell the truth about what the runtime already does.                                                                            

### 1. defineApiContract did not enforce method/body rules

  TypedPathApiContract used Omit<ApiContract, ...>, and Omit over a union is not distributive — it collapsed the `GetApiContract | DeleteApiContract | PayloadApiContract` discriminated union into a single object type with method: 'get' | 'delete' | 'post' | 'put' | 'patch' and an optional requestBodySchema. As a result, both of these compiled with zero errors:

  defineApiContract({ method: 'get', requestBodySchema: z.object({...}), ... })  // GET with a body
  defineApiContract({ method: 'post', ... })  // POST without its required requestBodySchema

  Fix: new DistributiveOmit utility in typeUtils.ts; TypedPathApiContract now transforms each union member separately. Both cases above are compile errors again.

###  2. Tagged noBodyResponse() was invisible to type-level inference

  The runtime treats the ContractNoBody symbol and the tagged noBodyResponse() object as equivalent, but the type level only knew the symbol. For a { 204: noBodyResponse() } contract:

  - IsNoBodySuccessResponse returned false (symbol: true) → now true
  - AvailableResponseModes resolved to never instead of 'noContent' → now 'noContent'
  - client response inference produced body: never — an untypeable value (symbol: null) → now null

  The no-body convention is now also documented explicitly: handler/server-side inference (InferNonSseSuccessResponses) maps no-body to undefined, client-side inference maps it to null — matching what
  each side actually produces.

###  3. pathResolver param was loosely typed without a requestPathParamsSchema

  When requestPathParamsSchema was omitted, the schema generic fell back to its constraint (z.ZodObject), typing the resolver param as `Record<string, unknown>` — while the runtime always calls `pathResolver(undefined)` in that case. So a resolver could read path params that are undefined at runtime (/users/undefined, or a TypeError for (params: { id: string }) => ``).

  Fix: the generic now defaults to undefined, so without a schema the resolver param is typed undefined and reading params (or declaring a param shape) is a compile error. With a schema, params stay fully typed as before.

###  4. Deprecated IsNoBodySuccessResponse

  No known consumers — marked @deprecated, to be removed in a future release. Tests retained until removal.

## Checklist

## AI Assistance Tracking

### DIFF
--- a/.changeset/deprecate-is-no-body-success-response.md
+++ b/.changeset/deprecate-is-no-body-success-response.md
@@ -1,0 +1,5 @@
+---
+"@lokalise/api-contracts": patch
+---
+
+Deprecate `IsNoBodySuccessResponse` — no known consumers; will be removed in a future release.
--- a/packages/app/api-contracts/src/new/inferTypes.ts
+++ b/packages/app/api-contracts/src/new/inferTypes.ts
@@ -2,7 +2,7 @@ import type { z } from 'zod/v4'
 import type { SuccessfulHttpStatusCode } from '../HttpStatusCodes.ts'
 import type { ValueOf } from '../typeUtils.ts'
 import type { ContractNoBody } from './constants.ts'
-import type { ResponsesByStatusCode } from './contractResponse.ts'
+import type { NoBodyResponse, ResponsesByStatusCode } from './contractResponse.ts'
 
 type ExtractSuccessResponses<T extends ResponsesByStatusCode> = ValueOf<
   T,
@@ -10,11 +10,14 @@ type ExtractSuccessResponses<T extends ResponsesByStatusCode> = ValueOf<
 >
 
 /**
- * Returns true if all success responses have no body (ContractNoBody or no success status codes defined).
+ * Returns true if all success responses have no body
+ * (ContractNoBody, noBodyResponse(), or no success status codes defined).
+ *
+ * @deprecated No known consumers — will be removed in a future release.
  */
 export type IsNoBodySuccessResponse<T extends ResponsesByStatusCode> = [
   ExtractSuccessResponses<T>,
-] extends [typeof ContractNoBody | undefined]
+] extends [typeof ContractNoBody | NoBodyResponse | undefined]
   ? true
   : false
 
@@ -63,7 +66,7 @@ type NonSseBodyOf<T> = T extends { _tag: 'SseResponse' }
 /**
  * Infers the TypeScript output type of all non-SSE success responses.
  * JSON schemas → z.output<T>. TextResponse → string. BlobResponse → Blob.
- * ContractNoBody → undefined. SseResponse → never (excluded).
+ * ContractNoBody and noBodyResponse() → undefined. SseResponse → never (excluded).
  * AnyOfResponses are unpacked before mapping.
  */
 export type InferNonSseSuccessResponses<T extends ResponsesByStatusCode> = NonSseBodyOf<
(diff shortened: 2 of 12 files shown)


.changeset/deprecate-is-no-body-success-response.md
CarlosGamero, line 1: There are 4 changesets, should will it generate 4 releases? if so, maybe worth grouping them in a single one

.changeset/deprecate-is-no-body-success-response.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lokalise/api-contracts": patch
+---
+
+Deprecate `IsNoBodySuccessResponse` — no known consumers; will be removed in a future release.
```
kibertoad, line 5: is this true? nothing in autopilot uses it?

packages/app/api-contracts/src/new/inferTypes.ts
```diff
@@ -2,19 +2,22 @@ import type { z } from 'zod/v4'
...
   T,
   Extract<keyof T, SuccessfulHttpStatusCode | '2xx' | 'default'>
 >
 
 /**
- * Returns true if all success responses have no body (ContractNoBody or no success status codes defined).
+ * Returns true if all success responses have no body
+ * (ContractNoBody, noBodyResponse(), or no success status codes defined).
+ *
+ * @deprecated No known consumers — will be removed in a future release.
```
CarlosGamero, line 16: Why will it be removed? I feel it might be useful right 🤔

packages/app/api-contracts/src/typeUtils.ts
CarlosGamero, line 1: A bunch of this utilities might be useful outside of api-contracts, maybe universal-ts-utils is better fit for them

packages/app/api-contracts/src/typeUtils.ts
```diff
@@ -20,6 +20,15 @@ export type Exactly<T, U> = T & {
   [K in keyof T]: K extends keyof U ? T[K] : never
 }
 
+/**
+ * Like `Omit`, but distributes over unions: each union member is transformed
+ * separately, preserving discriminated-union relationships between keys.
+ * (Plain `Omit` collapses a union into a single object type of its common keys.)
+ */
+export type DistributiveOmit<TUnion, TKeys extends PropertyKey> = TUnion extends unknown
```
CarlosGamero: Hmm, shouldn’t `TUnion` extend `object`? Also, could `TKeys` be `keyof TUnion`?

.changeset/deprecate-is-no-body-success-response.md
CatchMe2: It will be a single release with 4 fixes described in the CHANGELOG

CarlosGamero: Good to know! I learn something new today, thank you

.changeset/deprecate-is-no-body-success-response.md
CatchMe2: Yeah, it is a legacy type that I added in the initial api-contracts implementation, but then once I started working on fe/be http client, I added better-suited util types in `clientTypes.ts`, especially `InferClientResponseBody`.

packages/app/api-contracts/src/new/inferTypes.ts
CatchMe2: Please take a look at my comment here https://github.com/lokalise/shared-ts-libs/pull/1055#discussion_r3394410123

packages/app/api-contracts/src/typeUtils.ts
CatchMe2: This is true, but I just extended what we had there before. If we'd like to use an additional package for types, I'd prefer to pick https://github.com/sindresorhus/type-fest, which should cover all of the existing type utils, and is well tested.

packages/app/api-contracts/src/typeUtils.ts
CatchMe2: There is no need to narrow it to `object`, it would still allow to pass any in theory unwanted types, but in reality it will pass any non-primitive (arrays/functions/maps/...).
The implementation follows https://github.com/sindresorhus/type-fest/blob/main/source/distributed-omit.d.ts

Good catch on `keyof`, I added a dedicated `KeysOfUnion`, because standard `keyof` doesn't work well with unions.

CarlosGamero (approved):

kibertoad (approved):

PR #1090, opened 2026-06-23
Add api contracts multi content response

CatchMe2:
## Changes

Adds a { content: { <mediaType>: descriptor }, allowNoBody? } way to declare responses that return multiple content types — replacing anyOfResponses. Body descriptors are a bare Zod schema (JSON), blobBody(), or sseBody().

Why: anyOfResponses can't express multiple JSON media types, matches content-types by substring (false positives), is order-dependent, and has no no-body option. Content maps fix all of these, match media types exactly, and map 1:1 to the OpenAPI Response Object.

Additive — anyOfResponses and the other legacy entries keep working and can mix per status code.

## Checklist

## AI Assistance Tracking


packages/app/api-contracts/src/new/contractResponse.ts
```diff
@@ -170,6 +254,75 @@ const resolveByKind = (entry: TypedApiContractResponse): ResponseKind => {
...
+  if (!contentType) {
+    if (entry.allowNoBody) {
+      return { kind: 'noContent' }
+    }
+    return !strict && onlyDescriptor ? descriptorToKind(onlyDescriptor) : null
+  }
+
+  const target = normalizeMediaType(contentType)
+  for (const [mediaType, descriptor] of Object.entries(entry.content)) {
+    if (normalizeMediaType(mediaType) === target) {
```
kibertoad: Minor: this walks the content map twice (`Object.values` for `onlyDescriptor`, then `Object.entries` in the loop) and re-runs `normalizeMediaType(mediaType)` for every key on each resolution rather than normalizing the keys once. Content maps are tiny so the cost is negligible, but it is easy to fold into a single pass with pre-normalized keys. Flagging for cleanup.

kibertoad (commented):
Reviewed the content-map response entries. The resolver and the `clientTypes.ts` / `inferTypes.ts` inference are well-built, and existing entries route around the new code untouched. Two things to address in this PR.

## 1. Content-map entries crash the Mockttp mock helper (not in this diff, but enabled by it)

`packages/app/universal-testing-utils/src/api-contracts/ApiContractMockttpHelper.ts:143`

Widening `ResponsesByStatusCode` to include `ResponseEntry` flows into `ApiContract`, which `ApiContractMockttpHelper` consumes. `mockResponse` type-switches with `isNoBodyResponse` / `isTextResponse` / `isBlobResponse` / `isSseResponse` / `isAnyOfResponses`. A content entry (`{ content: {...} }` or `{ allowNoBody: true }`) matches none of them and falls through to:

```ts
const body = responseEntry.parse(anyParams.responseJson)
```

`responseEntry` is a plain object with no `.parse`, so this throws `TypeError: responseEntry.parse is not a function` at runtime.

There is no compile-time guard either: `InferBodyParam<ResponseEntry>` falls to its `object` default (`types.ts:58`), so `MockResponseParams` for such a status code is just `{ responseStatus }`, and `resolveContractEntry`'s declared return type (`ApiContractResponse | undefined`) excludes `ResponseEntry`.

Repro: any contract using a content entry, e.g. `200: { content: { 'application/json': schema } }`, passed to `helper.mockResponse(contract, { responseStatus: 200 })`, compiles cleanly and throws at runtime. Please handle content entries in the mock helper (and the `InferBodyParam` type) so the feature works end-to-end across the package set.

## 2. Minor efficiency in `resolveContentEntry`

See the inline comment.

Please address both within this PR.

kibertoad (approved):

PR #1096, opened 2026-06-30
feat(api-contracts): overhaul response API and require contract summary

CatchMe2:
## Summary

A breaking cleanup of the `@lokalise/api-contracts` response model (and the fe/be http clients that consume it), plus making contract `summary` mandatory.

### Response API
- **Removed deprecated APIs**: `textResponse` / `anyOfResponses` (+ their types and guards), `getSuccessResponseSchema`, `getIsEmptyResponseExpected`, `IsNoBodySuccessResponse`, and the `'text'` `ResponseKind`. Use content-map entries (`{ content: { '<mediaType>': descriptor } }`) or a bare Zod schema (JSON).
- **`ContractNoBody` is now request-body-only** — no longer a valid `responsesByStatusCode` entry. Use `noBodyResponse()` for no-body responses. (Still valid as a `requestBodySchema` value.)
- **`noBodyResponse()` / `blobResponse()` / `sseResponse()` kept as authoring helpers** but now build content-map entries (`{ allowNoBody: true }` / `{ content: { … } }`). Call sites are unchanged. The standalone tagged types/guards (`NoBodyResponse`, `TypedBlobResponse`, `TypedSseResponse`, `isNoBodyResponse`, `isBlobResponse`, `isSseResponse`) are removed. Blob/SSE bodies live only in content maps; JSON stays a bare Zod schema.

Net effect: a single uniform response model — a bare Zod schema for JSON, or a content-map `ResponseEntry` for everything else (blob / SSE / multi-media-type / no-body).

### Contract summary
- `summary` is now **mandatory** on every contract.
- It is surfaced in the fe/be `UnexpectedResponseError` (new required constructor arg + `readonly summary` field, included in the error message) for debugging.

## AI Assistance Tracking


packages/app/api-contracts/src/apiContracts.ts
```diff
@@ -322,7 +322,7 @@ export function buildGetRoute<
  *   method: 'delete',
  *   requestPathParamsSchema: z.object({ userId: z.string() }),
  *   pathResolver: ({ userId }) => `/api/users/${userId}`,
- *   responsesByStatusCode: { 204: ContractNoBody },
+ *   responsesByStatusCode: { 204: noBodyResponse() },
```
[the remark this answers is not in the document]
CatchMe2: Fixed

packages/app/api-contracts/src/new/contractResponse.ts
```diff
@@ -205,54 +102,56 @@ export const isContentResponseEntry = (
...
-  }
-
-  if (contentType.includes('application/json')) {
-    return { kind: 'json', schema: entry }
-  }
-
-  return null
-}
+): ResponseKind | null =>
+  contentType.includes('application/json') ? { kind: 'json', schema: entry } : null
```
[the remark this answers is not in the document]
CatchMe2: Fixed

.changeset/mandatory-contract-summary.md
```diff
@@ -0,0 +1,10 @@
+---
+"@lokalise/api-contracts": major
+"@lokalise/frontend-http-client": major
+"@lokalise/backend-http-client": major
```
CarlosGamero, line 4: Should we also bump fastify-api-contracts within this PR?

CarlosGamero (approved):

.changeset/mandatory-contract-summary.md
CatchMe2: Good catch, added it to changesets

packages/app/api-contracts/README.md
```diff
@@ -27,6 +28,7 @@ const getUser = defineApiContract({
 
 // POST
 const createUser = defineApiContract({
+  summary: 'Get user',
```
kibertoad: Every added `summary` in this README is `'Get user'` — including this POST `createUser`, plus `deleteUser` (DELETE), `exportCsv`, `downloadPhoto`, and the SSE/dual examples further down. Since these are the canonical copy-paste examples and the whole point of mandatory `summary` is a *meaningful* human-readable label, the identical `'Get user'` across every verb is misleading. Suggest giving each its correct summary (`'Create user'`, `'Delete user'`, `'Export products as CSV'`, …).

packages/app/frontend-http-client/src/api-contract/sendByApiContract.ts
```diff
@@ -130,10 +130,10 @@ async function parseBody(response: Response, resolvedEntry: ResponseKind) {
   switch (resolvedEntry.kind) {
     case 'noContent':
       return null
-    case 'text':
-      return await response.text()
     case 'blob':
-      return await response.blob()
+      // Hand back the raw stream; the caller decides whether to pipe it or aggregate it
+      // (e.g. `await new Response(body).blob()` / `.text()` / `.arrayBuffer()`).
+      return response.body
```
kibertoad: `response.body` is `ReadableStream<Uint8Array> | null`, but the inferred client body type for a blob entry is non-null `ReadableStream<Uint8Array>`. For a matched-blob response with an empty body, `response.body` can be `null`, so the type is mildly unsound at the edge (the backend's `Readable.toWeb(body)` doesn't have this gap). Consider normalizing `null` → an empty stream, or documenting the possibility.

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
```diff
@@ -182,11 +182,10 @@ async function parseBody(body: Dispatcher.ResponseData['body'], resolvedEntry: R
...
       return null
     }
-    case 'text': {
-      return await body.text()
-    }
     case 'blob': {
-      return await body.blob()
+      // Hand back the raw stream; the caller decides whether to pipe it or aggregate it
+      // (e.g. `await new Response(body).blob()` / `.text()` / `.arrayBuffer()`).
+      return Readable.toWeb(body)
```
kibertoad: Handing back the raw undici stream means the caller now **must** consume or `cancel()` it, otherwise the connection isn't released back to the pool — previously `.blob()` buffered and freed it automatically. This is a silent, production-only footgun; nothing at compile time reminds the caller. Please call it out explicitly in the migration/changeset notes. It's also the main argument for returning a thin body handle exposing `.blob()/.text()/.stream()` instead of a bare stream: it keeps the common (buffer-it) path a one-liner and gives a documented home for the drain obligation.

packages/app/api-contracts/src/new/clientTypes.spec.ts
```diff
@@ -319,36 +261,24 @@ describe('clientTypes', () => {
...
-      type Result = InferNonSseClientResponse<typeof contract>
-      expectTypeOf<Result>().toEqualTypeOf<{
-        statusCode: 200
-        headers: DefaultHeaders
-        body: string
-      }>()
-    })
-
-    it('maps blob success response to Blob body', () => {
+    it('maps blob success response to a readable stream body', () => {
```
kibertoad, line 312: Coverage gap: the dual-mode type tests were removed (`requires streaming: true for dual-mode contracts`, `extracts only SSE body for dual-mode success code`, `maps dual-mode success code to non-SSE body only`) but not replaced with content-map equivalents. Dual mode is still supported via a content map carrying both `application/json` and `text/event-stream` (see `dualContentApiContract`). Recommend re-adding a couple of `expectTypeOf` tests using a content-map dual entry to lock down the `streaming: true` requirement and the `ContractResponseMode → 'dual'` classification.

kibertoad (commented):
Thorough, well-executed breaking cleanup — net −1100 LOC and a genuinely simpler response model (bare Zod schema for JSON, content-map `ResponseEntry` for everything else). I verified the cross-cutting risks and found **no correctness bugs**:

- The barrel (`src/index.ts`) uses `export *`, so the removed symbols don't break re-exports.
- The only non-modified file referencing removed names is `sse/sseContractBuilders.ts`, and those are JSDoc-only (`@deprecated Use … anyOfResponses()`), not code — no compile break. **Nit:** those two comment references (around `sseContractBuilders.ts:205` and `:237`) are now stale and point at a removed helper; worth updating since the rest of the removal was so complete. (Commenting here because the file isn't in the diff.)
- All `new UnexpectedResponseError(...)` call sites are updated to the 4-arg signature.
- Mandatory `summary` is correctly scoped to `defineApiContract`'s `CommonApiContract`; the legacy builders keep their own optional `summary`.
- `Readable` is already imported for `Readable.toWeb(body)`.

The SSE-vs-blob type discrimination (`SseBodyShape = AsyncIterable<{ lastEventId: string }>`) is a nice touch — it correctly stops blob `ReadableStream`s from being mistaken for SSE now that both are `AsyncIterable`.

Inline comments below cover: a dual-mode type-test coverage regression, the README example summaries, a nullable-stream edge case, and the backend stream-drain obligation. All non-blocking. Changesets and the AI-assistance block look correct.

kibertoad:
## Design proposal: a body handle instead of a bare `ReadableStream` for blob responses

Following up on the inline thread about blob bodies. I want to be clear up front: **the direction here is right.** The client shouldn't unilaterally buffer on the caller's behalf, and "content-type is a contract fact, materialization is a call-site choice" is exactly the principle that justified removing `textResponse`. My concern is narrower — that `ReadableStream<Uint8Array>` as the *only, terminal* body type overcorrects. Below is the case against the current shape, then a concrete alternative.

### Downsides of the current stream-only approach

1. **The common case got more verbose, and the ergonomic default flipped.** The majority of blob consumers (CSV/report exports, image/PDF downloads, `octet-stream`) just want the bytes. `await res.blob()` / `.text()` becomes `await new Response(body).blob()`. That `new Response(body)` is a non-obvious idiom — most people don't know it's the canonical "drain a stream to a Blob" trick. We optimized the streaming minority at the expense of the buffering majority.

2. **Silent, production-only footgun on the backend (undici).** A `ReadableStream` the caller forgets to consume or `cancel()` holds the connection open — it isn't released back to the pool. `.blob()` freed it automatically. We've traded a runtime-managed memory cost for a caller-managed resource-lifetime obligation, and **nothing in the type reminds them**. This won't fail at compile time or in a unit test; it shows up as pool exhaustion under load.

3. **`Blob` is a value; `ReadableStream` is a one-shot.** A Blob can be read twice, sized (`.size`), sliced, and passed around. A stream is consumed once and then gone. For a "response body" that people log, assert on in tests, retry with, or branch on, one-shot semantics are a real downgrade — the test churn in this PR (`expect(body.size).toBe(4)` → a two-step drain) is a small preview of that friction.

4. **Web/Node semantic divergence.** `Readable.toWeb()` and Fetch's `response.body` don't have identical cancel/backpressure/error behavior across the two clients. A `Blob` is boringly identical everywhere; two raw streams from two runtimes are not.

(There's also the nullable-stream edge case flagged inline: `response.body` is `ReadableStream | null`, but the inferred type is non-null.)

### Proposal: return a lazy `ResponseBodyHandle`

Keep the raw stream as the primitive, but don't force the round-trip through `new Response(...)` on everyone. Return a lazy, single-consume handle — essentially the accessor surface `fetch`'s `Response`/`Blob` already expose (which is *why* `fetch` returns a `Response` and not a bare stream), minus the parts that don't apply:

```ts
/**
 * Lazy handle over a non-JSON response body. The underlying stream is one-shot:
 * the first accessor you call consumes it; a second call throws. Pick one.
 */
export interface ResponseBodyHandle {
  /** Raw stream — for piping/backpressure. You own draining or cancelling it. */
  stream(): ReadableStream<Uint8Array>
  /** Buffer the whole body. The common case. */
  blob(): Promise<Blob>
  text(): Promise<string>
  arrayBuffer(): Promise<ArrayBuffer>
  bytes(): Promise<Uint8Array>
}
```

**Frontend** — wraps the Fetch `Response`, which already implements all of this, so it's near-zero cost:

```ts
function toBodyHandle(response: Response): ResponseBodyHandle {
  let consumed = false
  const guard = <T>(read: () => T): T => {
    if (consumed) throw new Error('Response body already consumed')
    consumed = true
    return read()
  }
  return {
    stream: () => guard(() => response.body ?? new ReadableStream({ start: (c) => c.close() })),
    blob: () => guard(() => response.blob()),
    text: () => guard(() => response.text()),
    arrayBuffer: () => guard(() => response.arrayBuffer()),
    bytes: () => guard(async () => new Uint8Array(await response.arrayBuffer())),
  }
}
```

**Backend (undici)** — the guarded `blob()/text()` path is exactly where the connection gets drained, so "forgot to consume" stops being a silent leak:

```ts
function toBodyHandle(body: Dispatcher.ResponseData['body']): ResponseBodyHandle {
  let consumed = false
  const guard = <T>(read: () => T): T => {
    if (consumed) throw new Error('Response body already consumed')
    consumed = true
    return read()
  }
  return {
    stream: () => guard(() => Readable.toWeb(body)),
    blob:   () => guard(() => body.blob()),
    text:   () => guard(() => body.text()),
    arrayBuffer: () => guard(() => body.arrayBuffer()),
    bytes:  () => guard(async () => new Uint8Array(await body.arrayBuffer())),
  }
}
```

**Type inference** — one line changes in `inferTypes.ts` / `clientTypes.ts`:

```ts
// before
type NonSseBodyOf<T> = … : T extends { _tag: 'BlobBody' } ? ReadableStream<Uint8Array> : …
// after
type NonSseBodyOf<T> = … : T extends { _tag: 'BlobBody' } ? ResponseBodyHandle : …
```

**Call-site comparison:**

```ts
// this PR
const blob = await new Response(result.body).blob()   // common case: non-obvious idiom
result.body.pipeTo(dest)                              // streaming case: direct

// with handle
const blob = await result.body.blob()                 // common case: one-liner, discoverable
result.body.stream().pipeTo(dest)                     // streaming case: one extra call
```

### Trade-offs, stated honestly

- **Cost:** you give up the property that "the body *is* a standard web type." A caller who wants to hand the raw stream straight to another API now calls `.stream()` first, and it's a little more surface to maintain across two clients.
- **Win:** the buffering majority gets a discoverable one-liner, the one-shot footgun becomes a *thrown error* instead of silent corruption, and the backend connection leak gets a single documented chokepoint (which you could later extend to auto-cancel on GC/timeout).

Not blocking — if we'd rather ship the minimal version, that's defensible too, but then the `stream()`-only contract and the **backend drain obligation** need to be loud in the changeset/migration notes rather than implied. Happy to turn this into an actual diff against the branch if the shape looks right.

packages/app/frontend-http-client/src/api-contract/sendByApiContract.ts
CatchMe2: Resolved by introducing BlobHandle

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
CatchMe2: Resolved by introducing BlobHandle

kibertoad (approved):

PR #1123, opened 2026-07-23
feat: EXP-2169 Opentelemetry-fastify-bootstrap -> Skip SSE endpoints option

CarlosGamero:
## Changes

Adds an opt-in `skipStreamEndpoints` option to `initOpenTelemetry` (default `false`) that excludes streaming / SSE server spans from exported traces.

**Why:** an SSE connection stays open for the whole stream lifetime, so its auto-instrumented HTTP server span's duration reflects the keep-alive window (often minutes), not the time-to-first-byte. Any latency metric or SLO derived from that span duration is then skewed by those multi-minute values. (Concretely, this is what currently breaches the TSS Editor 500ms latency SLO via the `/editor/bulk-actions` SSE endpoint — its span p95 is ~300s.)

**How:**
- Streaming requests are detected by the `Accept: text/event-stream` request header — what browser `EventSource` clients are required to send, and the same signal SSE content-negotiation keys on. A FastifyOtel `requestHook` tags the request's server span with an internal `stream.endpoint` attribute.
- A new `StreamSpanFilteringExporter` (following the existing `DbNamespaceSpanExporter` decorator pattern) wraps the trace exporter and drops tagged spans **before export**, so they never reach the collector or its trace metrics. Generic for every streaming endpoint — no per-route exclusion list to maintain.
- The span still starts (trace context still propagates to child spans) and stays visible to console / user-supplied span processors; only the exported traces omit it.
- Extracted `createFastifyOtelInstrumentation` / `buildTraceExporter` helpers to keep `initOpenTelemetry` within the cognitive-complexity lint limit.

Opt-in and backward-compatible → **minor** bump. Consumers enable it with `initOpenTelemetry({ skipStreamEndpoints: true })`.

## Checklist

## AI Assistance Tracking


packages/app/opentelemetry-fastify-bootstrap/src/index.ts
```diff
@@ -106,6 +174,25 @@ export interface OpenTelemetryOptions {
...
+   * those metrics.
+   *
+   * Streaming requests are detected by the `Accept: text/event-stream` request
+   * header (what browser `EventSource` clients are required to send, and the
+   * same signal SSE content-negotiation keys on); matching spans are tagged and
+   * dropped before export. The span still starts (so trace context propagates
+   * to child spans) and stays visible to console / user-supplied span
+   * processors.
+   *
+   * @default false
```
kibertoad: wouldn't true b e a safer default here? do we ever want to instrument streaming requests?

CarlosGamero: I guess we don't want to, but if we set it to true by default it would be a breaking change. I am happy to do it if you think it is the right approach :D

packages/app/opentelemetry-fastify-bootstrap/src/index.ts
```diff
@@ -71,6 +78,67 @@ const logger = {
...
+ * FastifyOtel `requestHook`, so the span already exists; the marker is later
+ * consumed by {@link StreamSpanFilteringExporter} to drop the span before it is
+ * exported. Browser `EventSource` clients are required to send this header, and
+ * it is the same signal SSE content-negotiation keys on.
+ */
+const markStreamEndpointSpan = (span: Span, request: FastifyRequest): void => {
+  const accept = request.headers.accept
+  if (typeof accept === 'string' && accept.includes('text/event-stream')) {
+    span.setAttribute(STREAM_ENDPOINT_SPAN_ATTRIBUTE, true)
+  }
```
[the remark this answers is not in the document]
CarlosGamero: Done in db6e71fc — `markStreamEndpointSpan` now matches with `accept.toLowerCase().includes(...)` (media types are case-insensitive per RFC 7231 §3.1.1.1), plus a new spec injecting `Accept: Text/Event-Stream` to cover the case variant.

packages/app/opentelemetry-fastify-bootstrap/src/index.spec.ts
```diff
@@ -448,6 +453,42 @@ describe('opentelemetry-fastify-bootstrap', () => {
...
+      await waitForSpans(memoryExporter, 1)
+
+      const plainSpans = memoryExporter
+        .getFinishedSpans()
+        .filter(isFastifySpan)
+        .filter((span) => spanMentions(span, '/plain'))
+      expect(plainSpans.length).toBeGreaterThan(0)
+      expect(
+        plainSpans.every((span) => span.attributes[STREAM_ENDPOINT_SPAN_ATTRIBUTE] === undefined),
+      ).toBe(true)
```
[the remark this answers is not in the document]
CarlosGamero: Done in db6e71fc — both the streaming and non-streaming assertions now filter `.filter((span) => span.kind === SpanKind.SERVER)` before checking the marker, so they verify the server-span contract explicitly. Confirmed `@fastify/otel` starts the request span with `kind: SpanKind.SERVER`.

packages/app/opentelemetry-fastify-bootstrap/src/index.ts
CarlosGamero: After thinking about, decided to set it to true by default, thanks Igor 🙏

kibertoad (approved):

PR #1130, opened 2026-07-29
feat: Bull mq purge job data on success

CarlosGamero:
## Changes

Automatically purge job data on successful completion to reduce Redis footprint.

Previously, purging a completed job's data down to `metadata` was fully manual: consumers had
to call `purgeJobData` inside their own `onSuccess` hook, and most didn't — so completed jobs
kept their full `data` in Redis.

This PR makes purging automatic and configurable:

- **New `purgeJobDataOnSuccess` flag (defaults to `true`)**
  - New processor: on the per-queue `QueueConfiguration`.
  - Deprecated processor: on `BackgroundJobProcessorConfig`.
  - After a job completes successfully, its `data` is purged down to `metadata` (keeping
    `correlationId`). Set `purgeJobDataOnSuccess: false` to preserve the full data in Redis.
- The purge runs **after** the `onSuccess` hook, so the hook still sees the full job data.
- `_spy.addJob('completed')` now runs **after** the `onSuccess` hook (and before the purge).
  Previously it ran *before* the hook, as a workaround for consumers that called `purgeJobData`
  manually inside their own hook — now that purging is automatic and `purgeJobData` is `private`,
  that workaround is no longer needed. Recording the spy after the hook means
  `waitForJobWithId(..., 'completed')` resolves only once the hook's side effects have settled
  (e.g. a DB row written in `onSuccess` is committed by the time a test observes the completed
  job), and it makes the success path symmetric with the failure path, which already records the
  spy after `onFailed`. The spy is unaffected by the purge either way, since it deep-clones
  `job.data` on capture.
- Purge failures are reported to `errorReporter` and never crash the completion flow; job-missing
  errors during purge are silenced. Jobs configured with `removeOnComplete: true`/`1` are
  unaffected (BullMQ already removes them).
- `purgeJobData` is now `private` (was `protected`) since purging happens automatically.

### Breaking changes (major release)

- Job data is now purged on success **by default**. Consumers that read a completed job's full
  `data` from Redis must set `purgeJobDataOnSuccess: false`.
- `purgeJobData` is no longer accessible to subclasses. Remove any manual `purgeJobData` calls
  from `onSuccess` hooks.

## Checklist

## AI Assistance Tracking


packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -83,7 +103,7 @@ describe('AbstractBackgroundJobProcessorNew - success', () => {
     expect(job.data).toMatchObject(jobData)
 
     const resolvedJob = await queueManager.getQueue('queue1').getJob(job.id!)
-    expect(resolvedJob!.data).toMatchObject(jobData)
+    expect(resolvedJob!.data).toMatchObject({ metadata: jobData.metadata })
```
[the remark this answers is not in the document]
CarlosGamero: Good catch. Switched both sites to `toStrictEqual({ metadata: jobData.metadata })` so the assertion fails if any extra payload field survives the purge (`AbstractBackgroundJobProcessorNew.success.spec.ts:106` and `AbstractBackgroundJobProcessor.spec.ts:236`).

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -151,84 +171,100 @@ describe('AbstractBackgroundJobProcessorNew - success', () => {
...
 
-    const jobId = await queueManager.schedule('queue2', jobData)
+    // Then - job data is preserved in full.
+    await waitAndRetry(() => processorWithSuccessHook.runningPromisesSet.size === 0, 10, 5)
+    expect(processorWithSuccessHook.runningPromisesSet).toHaveLength(0)
+
+    expect(processorWithoutPurge.onSuccessCallsCounter).toBe(1)
+    const persistedJob = await queueManager.getQueue('queue3').getJob(jobId)
+    expect(persistedJob?.data).toStrictEqual(jobData)
+  })
```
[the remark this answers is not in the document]
CarlosGamero: Fixed. The assertion now awaits and checks `processorWithoutPurge.runningPromisesSet` (the processor handling `queue3`) instead of the unrelated `processorWithSuccessHook`.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -285,12 +285,18 @@ export abstract class AbstractBackgroundJobProcessorNew<
   private async internalOnSuccess(job: JobType): Promise<void> {
     const requestContext = this.monitor.getRequestContext(job)
 
-    this._spy?.addJob(job, 'completed') // this should be executed before the hook to not be affected by it
```
CarlosGamero, line 288: It was here before because some consumers were calling `purgeJobData` from within the hook.

That should no longer be the case, so we can now move the spy to the correct place. This also avoids race conditions. For example, if a successful execution updates a row in the database, by the time the spy runs, the row update has already been committed, so there are no more timing issues caused by this.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -285,12 +285,18 @@ export abstract class AbstractBackgroundJobProcessorNew<
...
     await this.internalOnHook(
       job,
       requestContext,
       async (job, requestContext) => await this.onSuccess(job, requestContext),
     )
+
+    this._spy?.addJob(job, 'completed')
+
+    // Purge after the onSuccess hook so it still sees the full job data. Enabled by default.
+    if (this.queueManager.getQueueConfig(this.queueId).purgeJobDataOnSuccess !== false) {
```
kibertoad, line 303: `dispose()` can return while this purge is still in flight.

`internalOnSuccess` is fired and forgotten from the `completed` listener in `registerListeners` and is never added to `runningPromises`. Only the inner `updateData`/`clearLogs` promises get registered, and only once `purgeJobData` actually starts, which is after the awaited `onSuccess` hook. So if `dispose()` runs while a hook is still awaiting, `Promise.allSettled(this.runningPromises)` sees an empty set and returns, and the purge then fires against a closing worker.

This gap existed before, but purging was opt-in and rarely used. Making it the default turns it into a routine shutdown path for every consumer.

Suggested fix: track the whole promise in `registerListeners`, e.g.

```ts
this._worker?.on('completed', (job) => {
  // @ts-expect-error
  const promise = this.internalOnSuccess(job, job.requestContext)
    .catch(() => undefined)
    .finally(() => this.runningPromises.delete(promise))
  this.runningPromises.add(promise)
})
```

Then `dispose()` genuinely drains the purge, and the inner per-promise bookkeeping in `purgeJobData` becomes redundant.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -285,12 +285,18 @@ export abstract class AbstractBackgroundJobProcessorNew<
...
     const requestContext = this.monitor.getRequestContext(job)
 
-    this._spy?.addJob(job, 'completed') // this should be executed before the hook to not be affected by it
     await this.internalOnHook(
       job,
       requestContext,
       async (job, requestContext) => await this.onSuccess(job, requestContext),
     )
+
+    this._spy?.addJob(job, 'completed')
```
kibertoad, line 300: Moving the spy call after the hook does not buy what the PR description says it does, and it costs something.

`BackgroundJobProcessorSpy.addJob` already stores `{ ...job, data: deepClone(job.data) }`, and its docblock says verbatim "JobData is cloned, to be protected against purge after the job is completed". The spy was already immune to the purge, so "spies capture the final (unpurged) job data" holds with the call in its original position too.

What the move does change: `waitForJobWithId(..., 'completed')` now resolves only after every consumer's `onSuccess` hook has settled. A slow hook delays all spy waits in consumer test suites, and a hook that never settles blocks them. That is exactly what the deleted `// this should be executed before the hook to not be affected by it` guarded against.

There is a fair argument the other way: `internalOnFailed` already calls `addJob` after its hook, so the move makes the two paths symmetric. If that is the real motivation, worth stating it in the description instead of the data-fidelity rationale. Otherwise I'd restore the original position in both processors.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -349,12 +355,12 @@ export abstract class AbstractBackgroundJobProcessorNew<
...
    * Removes all data associated with the job, keeps only correlationId.
-   * This method only works if the result of the job is not removed right after it is finished.
+   * Runs automatically after a successful job unless `purgeJobDataOnSuccess` is `false` on the
+   * queue configuration. This only works if the job is not removed right after it is finished.
    *
    * @param job
-   * @protected
    */
-  protected async purgeJobData(job: JobType): Promise<void> {
+  private async purgeJobData(job: JobType): Promise<void> {
```
kibertoad, line 369: Nit, pre-existing but now on the default path: the early return only treats `removeOnComplete === true | 1` as "BullMQ already removed it". `removeOnComplete: 0` and `{ count: 0 }` also remove the job, so those queues will now produce a silenced job-missing error on every completed job. Harmless (it is swallowed by `isJobMissingError`), but cheaper to widen the guard than to pay two doomed Redis calls per job.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -360,12 +360,18 @@ export abstract class AbstractBackgroundJobProcessor<
...
     await this.internalOnHook(
       job,
       requestContext,
       async (job, requestContext) => await this.onSuccess(job, requestContext),
     )
+
+    this._spy?.addJob(job, 'completed')
+
+    // Purge after the onSuccess hook so it still sees the full job data. Enabled by default.
+    if (this.config.purgeJobDataOnSuccess !== false) {
```
kibertoad, line 384: Same `dispose()` race as in the new processor, and slightly worse here: `dispose()` does

```ts
await this._worker?.close(this.config.isTest)
await this._queue?.close()
await Promise.allSettled(this.runningPromises)
```

so the queue is closed *before* pending purge promises are awaited. Since `internalOnSuccess` is not itself tracked in `runningPromises`, a purge that has not started yet when `dispose()` is called will run against a closed queue.

Fix is the same: register the `internalOnSuccess(...)` promise in `runningPromises` from the `completed` listener, and drain `runningPromises` before closing the queue.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -360,12 +360,18 @@ export abstract class AbstractBackgroundJobProcessor<
...
     const requestContext = this.monitor.getRequestContext(job)
 
-    this._spy?.addJob(job, 'completed') // this should be executed before the hook to not be affected by it
     await this.internalOnHook(
       job,
       requestContext,
       async (job, requestContext) => await this.onSuccess(job, requestContext),
     )
+
+    this._spy?.addJob(job, 'completed')
```
kibertoad, line 381: Same point as on `AbstractBackgroundJobProcessorNew.ts`: the spy already deep-clones `job.data` specifically to survive the purge, so this move is not needed for data fidelity and it makes `waitForJobWithId(..., 'completed')` depend on consumer `onSuccess` hooks.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -83,7 +103,7 @@ describe('AbstractBackgroundJobProcessorNew - success', () => {
     expect(job.data).toMatchObject(jobData)
 
     const resolvedJob = await queueManager.getQueue('queue1').getJob(job.id!)
-    expect(resolvedJob!.data).toMatchObject(jobData)
+    expect(resolvedJob!.data).toStrictEqual({ metadata: jobData.metadata })
```
kibertoad: Race. `waitForJobWithId(..., 'completed')` now resolves at `_spy.addJob`, which is *before* `purgeJobData` is called, so nothing here waits for the purge. On top of that, the purge writes through the worker's Redis connection while `queueManager.getQueue(...).getJob(...)` reads through the queue's, so there is no ordering guarantee between them.

It passes today because the purge is dispatched synchronously in the same microtask continuation, but it will flake under load. Either wait for the purge explicitly, e.g.

```ts
const resolvedJob = await waitAndRetry(async () => {
  const j = await queueManager.getQueue('queue1').getJob(job.id!)
  return j && !('value' in j.data) ? j : undefined
})
```

or make `dispose()` a real barrier (see my comment on `AbstractBackgroundJobProcessorNew.ts`) and dispose before asserting.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -151,84 +171,104 @@ describe('AbstractBackgroundJobProcessorNew - success', () => {
...
-    await processorWithSuccessHook.start()
+    // When
     const jobId = await queueManager.schedule('queue2', jobData)
+    await processorWithSuccessHook.spy.waitForJobWithId(jobId, 'completed')
 
-    const job = await processorWithSuccessHook.spy.waitForJobWithId(jobId, 'completed')
-
-    await processorWithSuccessHook.dispose()
+    // Then - the onSuccess hook still saw the full data, but the persisted job is purged.
+    await waitAndRetry(() => processorWithSuccessHook.runningPromisesSet.size === 0, 10, 5)
```
kibertoad: This guard is vacuous. `runningPromisesSet` is empty both *before* the purge registers its promises and after they settle, so `waitAndRetry` returns on its first evaluation and the assertions below race the purge.

Use a condition that actually reflects the purge having landed, for example poll the persisted job until its payload is gone:

```ts
const persistedJob = await waitAndRetry(async () => {
  const j = await queueManager.getQueue('queue2').getJob(jobId)
  return j && !('value2' in j.data) ? j : undefined
})
```

Also note `10, 5` gives a 50ms budget total, which is tight for two Redis round trips on a loaded CI runner.

kibertoad (commented):
Feature and API shape look right, changeset and README are in order, CI is green. Two things I'd want addressed before merge (the `dispose()` / `runningPromises` gap, and the spy reordering), plus some test-synchronization fixes that are latent flakes rather than current failures. All details are inline.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -151,84 +171,104 @@ describe('AbstractBackgroundJobProcessorNew - success', () => {
...
-        resolve()
-      }
-    })
+    // When
+    const jobId = await queueManager.schedule('queue3', jobData)
+    await processorWithoutPurge.spy.waitForJobWithId(jobId, 'completed')
 
-    const jobId = await queueManager.schedule('queue2', jobData)
+    // Then - job data is preserved in full.
+    await waitAndRetry(() => processorWithoutPurge.runningPromisesSet.size === 0, 10, 5)
```
kibertoad: Same vacuous guard as above. Here it is less harmful since the expectation is that nothing changes, but that also means this test cannot distinguish "purge correctly skipped" from "purge just hasn't run yet". Worth asserting after an actual barrier (dispose, or a fixed wait) so the test can fail if the flag stops being honoured.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -151,84 +171,104 @@ describe('AbstractBackgroundJobProcessorNew - success', () => {
...
+    const reportSpy = vi.mocked(deps.errorReporter.report)
 
     // When
+    const jobId = await queueManager.schedule('queue2', jobData)
     await processorWithSuccessHook.spy.waitForJobWithId(jobId, 'completed')
 
-    // Then
-    await expect(purgePromise).resolves.not.toThrow()
+    // Then - the purge ran and swallowed the job-missing error.
+    await waitAndRetry(() => processorWithSuccessHook.runningPromisesSet.size === 0, 10, 5)
```
kibertoad: Same vacuous guard. `expect(reportSpy).not.toHaveBeenCalled()` here can pass simply because the purge has not run yet, so the test would not catch a regression where the job-missing error stops being silenced. Needs a real barrier before the assertion.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -29,16 +30,29 @@ const supportedQueues = [
...
+    }),
+  },
 ] as const satisfies QueueConfiguration[]
 
 type SupportedQueues = typeof supportedQueues
 
 describe('AbstractBackgroundJobProcessorNew - success', () => {
   let factory: TestDependencyFactory
-  let deps: BackgroundJobProcessorDependenciesNew<SupportedQueues, 'queue1' | 'queue2'>
+  let deps: BackgroundJobProcessorDependenciesNew<SupportedQueues, 'queue1' | 'queue2', any>
```
kibertoad, line 51: Nit: `any` for the `JobReturn` parameter drops type checking on the return value the tests below now assert on. `object | undefined` matches what `TestSuccessBackgroundJobProcessorNew` declares.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
```diff
@@ -29,16 +30,29 @@ const supportedQueues = [
       }),
     }),
   },
+  {
+    // Dedicated queue for asserting the opt-out behaviour of the flag itself.
+    queueId: 'queue3',
+    purgeJobDataOnSuccess: false,
```
kibertoad, line 36: Worth a line in the README: for the new processor the flag lives on the queue config only. `BackgroundJobProcessorConfigNew` carries no override, so multiple processors on the same queue cannot opt out independently. That is a reasonable design (purging is a property of the queue's data), it just needs to be explicit so consumers do not go looking for a per-processor knob.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
```diff
@@ -233,7 +233,7 @@ describe('AbstractBackgroundJobProcessor', () => {
 
       // @ts-expect-error executing protected method for testing
       const resolvedJob = await processor.queue.getJob(job.id)
-      expect(resolvedJob!.data).toMatchObject(jobData)
+      expect(resolvedJob!.data).toStrictEqual({ metadata: jobData.metadata })
```
kibertoad: Same race as the equivalent assertion in `AbstractBackgroundJobProcessorNew.success.spec.ts`: the spy resolves before the purge is issued, and the read goes through a different Redis connection than the write. Nothing here synchronizes the two.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
```diff
@@ -338,28 +338,35 @@ describe('AbstractBackgroundJobProcessor', () => {
...
       const jobId = await successBackgroundJobProcessor.schedule(jobData)
 
-      const job = await successBackgroundJobProcessor.spy.waitForJobWithId(jobId, 'completed')
-
-      // When
+      const completedJob = await successBackgroundJobProcessor.spy.waitForJobWithId(
+        jobId,
+        'completed',
+      )
+      // Disposing waits for the background purge running promises to settle.
```
kibertoad: This comment is not accurate as written. `dispose()` only awaits `runningPromises`, and the purge promises are not in that set yet at the moment `dispose()` is entered (see my comment on `AbstractBackgroundJobProcessor.ts`). What actually makes this pass is that `await this._worker?.close(...)` takes long enough for the purge to complete. Either fix the drain in `dispose()` so the comment becomes true, or reword it.

packages/app/background-jobs-common/README.md
```diff
@@ -99,6 +99,34 @@ To set up a queue configuration, you need to define a list of objects containing
...
+    jobPayloadSchema: z.object({
+      id: z.string(),
+      value: z.string(),
+      metadata: z.object({ correlationId: z.string() }),
+    }),
+  },
+] as const satisfies QueueConfiguration[]
+```
+
+Jobs configured with `removeOnComplete: true` (or `1`) are unaffected, since BullMQ already removes them on completion.
```
kibertoad: Two caveats worth documenting here, since the default is on:

1. Retrying a completed job (the Bull Board retry button, `job.retry()`) will re-run it with `{ metadata }` only and fail `jobPayloadSchema` validation. This is the most likely way a consumer gets surprised by the new default.
2. The purge adds two Redis round trips per completed job (`hset` for the data plus a log key delete) on every queue by default. Fine for most queues, worth flagging for high-throughput ones so they know `purgeJobDataOnSuccess: false` is also a throughput lever.

Also worth stating explicitly that the flag is queue-wide for the new processor, with no per-processor override.

packages/app/background-jobs-common/test/processors/TestSuccessBackgroundJobProcessorNew.ts
```diff
@@ -1,42 +1,76 @@
...
+  constructor(
+    dependencies: Omit<
+      BackgroundJobProcessorDependenciesNew<Q, T, object | undefined>,
+      'workerFactory' | 'transactionObservabilityManager'
+    >,
+    queueId: T,
+  ) {
+    super(
+      {
+        transactionObservabilityManager: {
```
kibertoad: Nit: this no-op `transactionObservabilityManager` stub is now duplicated verbatim in `TestSuccessBackgroundJobProcessor.ts` too. Both helpers used to inherit it from the `Fake*` processors. Worth extracting a shared stub under `test/`.

Related: dropping `FakeBackgroundJobProcessor` / `FakeBackgroundJobProcessorNew` as the base classes removes the incidental coverage these specs gave those two exported classes. Probably fine, just noting it.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
CarlosGamero: Good catch, fixed in 1eb86a18.

Both worker listeners now track the whole `internalOnSuccess` / `internalOnFailed` promise in `runningPromises` (registered synchronously when the event fires, before the awaited hook), so `dispose()` genuinely drains the hook + purge instead of seeing an empty set:

```ts
this._worker?.on('completed', (job) => {
  // @ts-expect-error
  const promise = this.internalOnSuccess(job, job.requestContext)
    .catch(() => undefined)
    .finally(() => this.runningPromises.delete(promise))
  this.runningPromises.add(promise)
})
```

As you noted, this makes the per-promise bookkeeping inside `purgeJobData` redundant, so I dropped it there. Same fix applied to the deprecated processor, where I also reordered `dispose()` to drain `runningPromises` **before** `_queue.close()` so the purge never writes through a closed queue connection.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: Fixed in the same commit (1eb86a18). The `completed` / `failed` listeners here now register the whole `internalOnSuccess` / `internalOnFailed` promise in `runningPromises`, and `dispose()` was reordered to drain them **before** closing the queue:

```ts
await this._worker?.close(this.config.isTest)
// Drain pending hook/purge work before closing the queue, so purges that
// write through the queue connection do not race a closed queue.
await Promise.allSettled(this.runningPromises)
await this._queue?.close()
```

So the purge is now genuinely drained instead of relying on `worker.close()` taking long enough. The per-promise bookkeeping in `purgeJobData` was dropped as redundant here too.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
CarlosGamero: You're right that the spy is immune to the purge either way (it deep-clones `job.data` on capture), so "spies capture the final unpurged data" was the wrong rationale — I've corrected the PR description.

The move is intentional but for a correctness reason, not data fidelity:

1. The spy used to run *before* the hook specifically as a workaround for consumers that called `purgeJobData` manually inside their own `onSuccess` hook, which corrupted what the spy observed. Now that purging is automatic and `purgeJobData` is `private`, that workaround is no longer needed.
2. Recording the spy *after* the hook means `waitForJobWithId(..., 'completed')` resolves only once the hook's side effects have settled — e.g. a DB row written in `onSuccess` is committed by the time a consumer test observes the completed job. That flow was effectively broken before, and some consumers rely on it.
3. It also makes the success path symmetric with `internalOnFailed`, which already records the spy after `onFailed`.

Re your concern about a slow/never-settling hook blocking spy waits: agreed that's the real trade-off. It's the same coupling the failure path already has, and a hook that never settles is a consumer bug that would also stall `dispose()`. I've made the rationale explicit in the description rather than the data-fidelity claim.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CarlosGamero: Same as on the new processor: the spy already deep-clones `job.data`, so the move isn't about data fidelity (PR description corrected). The reason it now runs after the hook is (1) the pre-hook position was a workaround for consumers calling `purgeJobData` manually inside their hook, which no longer applies, (2) `waitForJobWithId(..., 'completed')` now resolves only after the hook's side effects have settled, and (3) symmetry with `internalOnFailed`, which already records the spy after the hook here too.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
CarlosGamero: Good point. Extracted the check into an `isJobRemovedOnComplete(options)` util shared by both processors and widened it to also treat `removeOnComplete: 0` and `{ count: 0 }` as already-removed, so those queues no longer pay two doomed Redis calls plus a silenced job-missing error on every completed job. Added unit coverage for the util.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
CarlosGamero: You're right. Rather than synchronize this test with the purge, I removed the purged-data assertion from `runs the job with autogenerated id` entirely — that test is about id generation / the spy / the worker, not purging. The purge is asserted in `purges job data on success by default`, which now polls the persisted job until its payload is gone.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
CarlosGamero: Fixed. Replaced the `runningPromisesSet.size === 0` guard with polling the persisted job until its payload is gone, so the assertion waits for the purge to actually land instead of returning on the first tick.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
CarlosGamero: Fixed. This opt-out test now awaits `dispose()` as a real drain barrier before asserting the data is untouched (dispose() now drains the whole `internalOnSuccess`, purge included), so it can actually fail if the flag stops being honoured.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
CarlosGamero: Fixed. Now awaits `dispose()` before asserting `reportSpy` was not called, so the assertion runs after the purge has completed rather than before it starts.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
CarlosGamero: Same as on the new processor: removed the purged-data assertion from this `runs the job` test, since it isn't about purging. The purge is covered by the dedicated purge tests, which use `dispose()` as a barrier.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.spec.ts
CarlosGamero: Fixed both ways you suggested: `dispose()` now genuinely drains the purge (the `completed` listener registers the whole `internalOnSuccess` promise in `runningPromises`, and `dispose()` drains it before closing the queue), and I reworded the comment to describe exactly that.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
CarlosGamero: I looked into this and kept `any` here, because `object | undefined` doesn't type-check in this spot:

- `JobReturn` is invariant in `BackgroundJobProcessorDependenciesNew` (the `QueueManager` / `Queue` it carries use it in both input and output positions — e.g. `getWaiting()` returns `Job<..., JobReturn>` while other methods consume it). So the type parameter has to match exactly, not just be a supertype.
- This single `deps` is shared across processors with different returns: `FakeBackgroundJobProcessorNew` and `factory.createNew(...)` are `void`, while `TestSuccessBackgroundJobProcessorNew` is `object | undefined`. `any` is the only type assignable to all of them; `object | undefined` fails against the `void` ones (`Type 'void' is not assignable to type 'object | undefined'`).
- The `returnvalue` the tests assert on is actually typed through `queueManager.getQueue(...).getJob(...)`, not through this `deps` parameter, so widening it here wouldn't add type-safety to those assertions anyway.

Making it precise would mean turning `createNew` generic over `JobReturn` and casting at each processor construction, which felt disproportionate for a test-only nit. Happy to do that refactor if you'd prefer it over `any`.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.success.spec.ts
CarlosGamero: Added to the README's "Purging job data on success" section: it now states explicitly that for the new processor the flag is queue-wide (lives on `QueueConfiguration` only), that `BackgroundJobProcessorConfigNew` has no per-processor override, and that multiple processors on the same queue can't opt out independently — with a note that the deprecated processor carries the flag on its own config instead.

packages/app/background-jobs-common/README.md
CarlosGamero: Documented both caveats in the README under a new "Caveats" subsection:

1. Retrying a completed job re-runs it with `{ metadata }` only and fails `jobPayloadSchema` validation (Bull Board *Retry* / `job.retry()`), with a pointer to `purgeJobDataOnSuccess: false` for queues whose jobs may be retried.
2. The purge costs two Redis round trips per completed job (`hset` + log-key delete), and `purgeJobDataOnSuccess: false` is called out as a throughput lever for high-throughput queues.

Also reworded the `removeOnComplete` note to reflect the widened guard (`true`, a keep-count of `0` or `1`, or `{ count: 0 }`).

packages/app/background-jobs-common/test/processors/TestSuccessBackgroundJobProcessorNew.ts
CarlosGamero: Good catch. Instead of a bespoke shared stub, I replaced the duplicated inline no-op in both `TestSuccess*` helpers with node-core's canonical `NoopObservabilityManager` (`new NoopObservabilityManager()`), which implements `TransactionObservabilityManager` as a no-op — so the verbatim duplication is gone without adding a new test helper.

On the incidental-coverage note: agreed it's fine. `FakeBackgroundJobProcessor` / `FakeBackgroundJobProcessorNew` are still directly exercised by the other specs (e.g. the `AbstractBackgroundJobProcessor(New)` suites construct them directly), so they keep their coverage.

CarlosGamero:
@kibertoad all comments addressed, thank you so much for the in-depth review 🙏

kibertoad (approved):

PR #1133, opened 2026-07-29
chore: new lib -> text-measure 

CarlosGamero:
## Changes

New public package **`@lokalise/text-measure`** — text-measurement utilities for Lokalise content. This PR scaffolds the package only (empty `index`); the utilities land in follow-ups.

**What it will support**
- Character count (Unicode code points; excludes NTC marker tags, includes prefix/suffix) — the first use case.
- Word count (the "translatable words" logic that currently lives in TSS) to follow.
- Extensible to other counting **units** (e.g. bytes) in a future iteration — it starts as a simple count; the unit option will be added later in an additive, non-breaking way.
- Universal (FE + BE), no runtime deps — usable from the editor, TSS (usage/limit, QA) and, over time, Polyglot, so every layer counts identically.

**Why a new package instead of extending an existing one**
- **Single responsibility / discoverability** — "measure text (length, words)" is its own concern; a dedicated, well-named package is findable and evolves cleanly as counting units grow.
- **Not `@lokalise/non-translatable-markup`** — that lib owns *what is non-translatable* (tags), not *how long text is*; a unit-aware counter there would couple two concerns and bloat a markup lib. This package instead *depends on* it to strip NTC markers — reuse, not duplicate.
- **Not `@lokalise/polyglot-sdk`** — it's AI-oriented and counts in UTF-16 today; pulling it into the FE editor just to count is a heavy, semantically wrong dependency. Better that polyglot-sdk consumes this lib so AI counts the same way.
- **Consistency is the real driver** — one shared source of truth prevents editor / QA / AI counts from drifting.

## Checklist

## AI Assistance Tracking


packages/app/text-measure/src/index.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: Public API for @lokalise/text-measure — exports go here.
+export {}
```
CarlosGamero: if the new package proposal is approved, we will implement the lib in a separate PR

dariacm (approved):

drdaemos (approved):

packages/app/background-jobs-common/package.json
```diff
@@ -62,7 +62,7 @@
         "rimraf": "^6.1.3",
         "toad-scheduler": "^4.0.1",
         "typescript": "7.0.2",
-        "vitest": "^4.1.10",
+        "vitest": "^4.1.9",
```
kibertoad: why the downgrade?

packages/app/text-measure/LICENSE.md
```diff
@@ -0,0 +1,13 @@
+Copyright 2024 Lokalise, Inc.
```
kibertoad:
```suggestion
Copyright 2026 Lokalise, Inc.
```

kibertoad (approved):

packages/app/background-jobs-common/package.json
CarlosGamero: All the repo is using 4.1.9 and wanted to standarise it in the monorepo. Probably better to do an upgrade of the rest, will do it before merging

CarlosGamero: Done!

packages/app/text-measure/LICENSE.md
CarlosGamero: Good catch! thank you 🙏

PR #1134, opened 2026-07-30
feat: EXP-2186 Text measure lib implementation

CarlosGamero:
## Changes

Implements `@lokalise/text-measure` (EXP-2186) — the utilities on top of the package scaffolded in #1133.

**`countCharacters(text, options?)`**
Counts characters. NTC marker tags are excluded (the content they wrap is kept); everything else in the text, including whitespace, is counted. The algorithm is pluggable via `options.algorithm`:
- `utf16` (default) — UTF-16 code units (JS `String.length`).
- `codePoints` — Unicode code points.

**`countTranslatableWords(text, options?)`**
Counts translatable words, excluding NTC regions and tags, via `gmx-word-counter`; `options.locale` (BCP47) drives segmentation. Extracted from the logic that lives in TSS today so FE, BE and Polyglot can share a single implementation.

**Dependencies:** `@lokalise/non-translatable-markup` (workspace) and `gmx-word-counter`.

### Why `utf16` is the default

The first integration where we enforce character limits is Contentful, and we verified empirically how Contentful measures a field's length: it counts **UTF-16 code units** (equivalent to JS `String.length`). Measured directly in a Contentful field:

| Input | Contentful | `utf16` | `codePoints` |
|-------|:---------:|:-------:|:------------:|
| `€` | 1 | 1 | 1 |
| `👍` | 2 | 2 | 1 |
| `👨‍👩‍👧` | 8 | 8 | 5 |

_The published package stays integration-agnostic; the Contentful rationale for the default is kept here in the PR, not in the package docs._

## Checklist

## AI Assistance Tracking

### DIFF
--- a/packages/app/text-measure/package.json
+++ b/packages/app/text-measure/package.json
@@ -31,6 +31,10 @@
         "package-version": "echo $npm_package_version",
         "postversion": "biome check --write package.json"
     },
+    "dependencies": {
+        "@lokalise/non-translatable-markup": "workspace:^",
+        "gmx-word-counter": "^1.2.0"
+    },
     "devDependencies": {
         "@biomejs/biome": "^2.5.6",
         "@lokalise/biome-config": "workspace:^",
--- a/packages/app/text-measure/package.json
+++ b/packages/app/text-measure/package.json
@@ -6,6 +6,7 @@
     ],
     "license": "Apache-2.0",
     "type": "module",
+    "sideEffects": false,
     "main": "./dist/index.js",
     "homepage": "https://github.com/lokalise/shared-ts-libs",
     "repository": {
--- a/.changeset/text-measure-initial-release.md
+++ b/.changeset/text-measure-initial-release.md
@@ -1,0 +1,8 @@
+---
+"@lokalise/text-measure": major
+---
+
+Add `@lokalise/text-measure`, utilities to measure text:
+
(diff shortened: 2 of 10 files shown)


packages/app/text-measure/package.json
```diff
@@ -31,6 +31,10 @@
         "package-version": "echo $npm_package_version",
         "postversion": "biome check --write package.json"
     },
+    "dependencies": {
+        "@lokalise/non-translatable-markup": "workspace:^",
+        "gmx-word-counter": "^1.2.0"
+    },
```
diogomiguel, line 38: **suggestion:** 
With a single barrel entry and no "sideEffects" declaration, importing countCharacters also pulls in `countTranslatableWords` and `gmx-word-counter` with it. Both modules look side-effect free, so adding "sideEffects": false next to "type": "module" would let bundlers keep each consumer's cost to what it actually imports.

CarlosGamero: Good call — added `"sideEffects": false`. The modules are side-effect-free, so this lets bundlers drop `countTranslatableWords` (and `gmx-word-counter`) for consumers that only import `countCharacters`, which matters here since the editor (FE) is a consumer. Thanks!

diogomiguel (approved):

eduard-fa (approved):

kjamrog (approved):

PR #1155, opened 2026-08-10
docs(fastify-api-contracts): document SSE-aware error handling

CatchMe2:
## Summary

Documentation-only (follow-up to the Slack discussions on #1032 error handling): the earlier error-handling implementation in this PR was reverted — error handling stays with the regular Fastify chain, and the shared `@lokalise/fastify-extras` error handler is where SSE awareness will be implemented (separate follow-up).

This PR adds an **Error handling** section to the `buildFastifyApiRoute` docs describing the current behavior and the recommended global-handler setup:

- Errors from contract routes go through the regular Fastify error handling chain — contract routes behave exactly like any other route.
- On an `autoClose` SSE session, a handler error closes the stream without reaching the error handler — clients detect failure by the missing terminal event (e.g. `done`).
- On a `keepAlive` session, a handler rejection reaches the error handler with the stream still open, where only a terminal `error` event can be sent. The documented `setErrorHandler` pattern branches on `reply.sse?.isConnected && reply.raw.headersSent` (the plugin's `isConnected` flag alone is already true before the stream starts).

Changeset: patch (published README).

## AI Assistance Tracking


packages/app/fastify-api-contracts/src/api-contracts/index.ts
```diff
@@ -1,3 +1,9 @@
+export {
+  type ApiErrorResponse,
+  type FastifyApiContractsOptions,
+  fastifyApiContracts,
+  type ResolveApiErrorResponse,
+} from './apiErrorHandler.ts'
```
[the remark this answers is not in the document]
CatchMe2: I'd rather keep it internal

kibertoad (commented):
Code review of the error-handling changes. Five findings, all validated against the real `@fastify/sse@0.5.0` and `fastify@5.10.0` sources; the first three with runnable repros over real sockets. Details are inline.

Checked and confirmed as non-issues, since they look suspicious: a handler that throws before `sse.start()` is safe (`@fastify/sse` closes the never-started context in its own `catch` because `shouldKeepAlive` is false, so `isConnected` is already false); `preHandler`/validation errors are safe because `reply.sse` does not exist yet; `fastify.errorHandler` is a real public getter in Fastify 5; `keepAlive()` does not interact with the `heartbeat` route option; and `fastify-plugin@6.0.0` is a real published version.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -148,6 +173,17 @@ export function buildApiSSEContext(
   let sessionMode: SSESessionMode | undefined
   let closedByServer = false
 
+  const closeAsServer = () => {
+    closedByServer = true
+    reply.sse.close()
+  }
+
+  ;(reply as WithApiSseRuntime)[kApiSseRuntime] = {
+    isConnected: () => reply.sse.isConnected,
```
kibertoad: **High: `isConnected` is already true before `sse.start()`, so non-SSE errors get answered as an SSE stream.**

`@fastify/sse` sets `#isConnected = true` in the `SSEContext` constructor, which runs before the handler. The flag is therefore true on every SSE-capable (`kind: 'manual'`) route from the first line of the handler, long before `sse.start()` is called, and possibly on a request that never starts a stream at all. Any error raised while the handler is still on the stack takes the SSE branch of `apiRouteErrorHandler`.

Concrete case, verified over a real socket: an SSE-capable contract whose handler returns the JSON representation, with a body that fails the `fastify-type-provider-zod` response schema. Fastify calls the route `errorHandler` synchronously from inside `reply.send()`, `sse.isConnected()` returns true, and the client receives `HTTP 200` / `content-type: text/event-stream` with `event: error\ndata: {...}` instead of the resolver's status code and JSON payload.

`buildApiSSEContext` already tracks `started`; gating the runtime on it fixes this:

```ts
isConnected: () => started && reply.sse.isConnected,
```

The same premature-true flag also makes the `reply.sse?.isConnected` check the README recommends for a user `setErrorHandler` misfire in this case.

packages/app/fastify-api-contracts/src/api-contracts/apiErrorHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+      try {
+        return await request.server.errorHandler(error, request, reply)
+      } finally {
+        if (sse?.isConnected()) {
+          sse.close()
+        }
+      }
+    }
+
+    const { statusCode, payload, headers } = await configuredResolver(error, request, reply)
```
kibertoad: **High: a throwing resolver leaves the SSE stream open and takes the process down.**

This `await` is outside any `try`/`finally` that closes the stream. Because `start()` now always calls `reply.sse.keepAlive()`, `@fastify/sse` no longer closes the stream on a handler rejection, so this route `errorHandler` is the only thing that can close it.

If a user's `resolveErrorResponse` throws or rejects while a stream is live (the documented usage calls `errorReporter.report()` and an error-object resolver inside it, both of which can throw), the stream is never closed, the rejection reaches Fastify's `fallbackErrorHandler`, and `res.writeHead()` on committed headers throws `ERR_HTTP_HEADERS_SENT` from a promise rejection callback: an uncaught exception that kills the process, plus a hung connection. Reproduced.

The `sendErrorEvent` failure path below is already guarded this way; the resolver call needs the same guard.

packages/app/fastify-api-contracts/src/api-contracts/apiErrorHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+    error: FastifyError,
+    request: FastifyRequest,
+    reply: FastifyReply,
+  ): Promise<void> {
+    const configuredResolver = routeResolver ?? request.server[apiContractsErrorResolverKey]
+    const sse = getApiSseRuntime(reply)
+
+    if (!configuredResolver) {
+      try {
+        return await request.server.errorHandler(error, request, reply)
```
kibertoad: **Medium: the app-level error handler runs twice when it throws.**

Delegating with `await request.server.errorHandler(...)` bypasses Fastify's native chain. Fastify sets `reply[kReplyNextErrorHandler]` to the *parent* of the handler it just ran; here the handler that "just ran" is the built-in route handler, so its parent is the app handler, the one that already threw.

Measured against fastify@5.10.0: an app `setErrorHandler` that throws is called once natively and twice through this wrapper. Rethrowing to delegate to the parent scope's handler is a supported Fastify pattern, so apps using it get duplicated error reporting.

packages/app/fastify-api-contracts/src/api-contracts/apiErrorHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * Registers an app-wide error resolver for every contract route built with
+ * `buildFastifyApiRoute` — the single place to serialize (and report) errors for all
+ * contract endpoints. A route-level `resolveErrorResponse` option overrides it per route.
+ */
+export const fastifyApiContracts = fp<FastifyApiContractsOptions>(
+  (app: FastifyInstance, options, done) => {
+    app.decorate(apiContractsErrorResolverKey, options.resolveErrorResponse)
```
kibertoad: **Medium: registering the plugin twice crashes the app.**

`app.decorate` throws `FST_ERR_DEC_ALREADY_PRESENT` on a second registration. `fastify-plugin`'s `name` option does not deduplicate, and because `fp` breaks encapsulation every registration targets the root instance, so two independent sub-plugins (or a bootstrap plus a test helper) each registering `fastifyApiContracts` crashes at `ready()`. Reproduced.

Guard with `app.hasDecorator(apiContractsErrorResolverKey)`, and reassign instead of decorating on a repeat registration.

packages/app/fastify-api-contracts/src/api-contracts/apiHandlerTypes.ts
```diff
@@ -265,10 +266,16 @@ export type InferApiHandler<Contract extends ApiContract> = [
...
+ * the contract provides (`method`, `url`, `schema`, `handler`, `sse`, `errorHandler` —
+ * error handling is customized via `resolveErrorResponse` instead, since a raw
+ * `errorHandler` never sees mid-stream SSE errors and would only half-replace the built-in
+ * behavior), plus the SSE lifecycle options (`onConnect`/`onClose`/`onReconnect`, …) that
+ * apply only to SSE-capable contracts.
  */
-export type ApiRouteOptions = Omit<RouteOptions, 'method' | 'url' | 'schema' | 'handler' | 'sse'> &
+export type ApiRouteOptions = Omit<
+  RouteOptions,
+  'method' | 'url' | 'schema' | 'handler' | 'sse' | 'errorHandler'
```
kibertoad: **Low: this is a breaking change shipped as a `minor`.**

`errorHandler` was accepted by `buildFastifyApiRoute` before and is now removed from `ApiRouteOptions`, but `.changeset/contract-error-handling.md` marks the release `minor`. A consumer currently passing `errorHandler` gets a compile error on what looks like a non-breaking upgrade, and their handler is silently dropped at runtime.

Either bump the changeset to `major`, or call the removal and the `resolveErrorResponse` migration out explicitly in the summary.

CarlosGamero (approved):

PR #1174, opened 2026-08-13
fix(opentelemetry-fastify-bootstrap): bump otel peers past security patches

ppletscher:
## Changes

Bumps the OpenTelemetry peer (and matching dev) dependency ranges of `@lokalise/opentelemetry-fastify-bootstrap`:

| Peer | From | To |
|---|---|---|
| `@opentelemetry/sdk-node` | ^0.218.0 | ^0.221.0 |
| `@opentelemetry/exporter-trace-otlp-grpc` | ^0.218.0 | ^0.221.0 |
| `@opentelemetry/auto-instrumentations-node` | ^0.76.0 | ^0.79.0 |
| `@opentelemetry/sdk-trace-base` | ^2.7.1 | ^2.10.0 |

**Why:** `sdk-node@0.218` exact-pins two packages below their security patches — `@opentelemetry/core@2.7.1` (unbounded memory allocation in W3C Baggage propagation, fixed in 2.8.0) and `@opentelemetry/propagator-jaeger@2.7.1` (DoS via unhandled exception on a malformed header, fixed in 2.9.0). Since the peer range capped sdk-node below 0.219, consumers currently need pnpm overrides to get the patched versions (see [lokalise/translation-memory#862](https://github.com/lokalise/translation-memory/pull/862) and [this thread](https://github.com/lokalise/translation-memory/pull/862#discussion_r3773647843)). `sdk-node@0.221` pins both at 2.10.0, so after this release consumers can drop those overrides by just updating the bootstrap.

Released as a **major** (changeset included), consistent with 3.0.0 which was also a peer-range bump: consumers relying on pnpm `auto-install-peers` need no code changes, but anyone declaring the peers explicitly must move to the new ranges.

Build, lint, and all 56 tests of the package pass against the new versions.

## Checklist

## AI Assistance Tracking


.changeset/otel-security-peer-bump.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lokalise/opentelemetry-fastify-bootstrap": major
```
CarlosGamero: I’d say it’s a patch. I consider security-related bumps to be patches, but please feel free to disagree.

Just for context and knowledge sharing: having a major version bump in this package wouldn’t be a big deal. However, for fastify-extras, since it’s quite a core package, it would trigger updates in quite a few places 😓

CarlosGamero (approved):

ppletscher:
So, spoke to CC about this:

```
My assessment: keep major

The reason this is major isn't that it's a security bump — it's what kind of change it is. The PR narrows peer dependency ranges to ones that don't overlap with the old ones: ^0.218.0 means >=0.218.0 <0.219.0 for these 0.x OTel packages, so ^0.221.0 excludes every previously-valid version. A consumer currently on sdk-node@0.218 who picks up this release has unmet peers until they bump their own OTel deps. "You must change your own dependencies to consume this release" is the textbook definition of breaking.

The patch label matters practically because of what it signals to tooling: Renovate/Dependabot auto-merge patches. A patch that invalidates every existing consumer's peer set is exactly the kind of surprise semver is supposed to prevent — whereas a major makes the required OTel bump a conscious, visible step, which is also what you want here since the whole point is pushing consumers past the vulnerable versions.

There's also direct precedent in this exact package: v3.0.0 was a major for the identical change ("Bump OpenTelemetry peer dependencies to latest… ^0.218.0…"), with no API change either.

The genuine middle ground would be widening the range (>=0.218.0 <0.222.0) and shipping a patch — that breaks nobody — but it defeats the purpose of the PR, since it keeps blessing the vulnerable versions.
```

No strong opinion and I trust you guys, so whatever you say. But given the above, seems we should keep `major`? @kibertoad opinions?

PR #1175, opened 2026-08-13
feat: EXP-2244 Text measure skip ntc

CarlosGamero:
## Checklist

## AI Assistance Tracking

### DIFF
--- a/.changeset/text-measure-exclude-ntc.md
+++ b/.changeset/text-measure-exclude-ntc.md
@@ -1,0 +1,5 @@
+---
+"@lokalise/text-measure": minor
+---
+
+Add `excludeNtc` option to `countCharacters` to exclude non-translatable content (text wrapped between NTC tags) from the count. Defaults to `false`, preserving the current behavior of counting the wrapped content.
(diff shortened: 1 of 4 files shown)


.changeset/text-measure-exclude-ntc.md
kjamrog, line 1: Do we use those? I don't see other changeset md files in the repo

CarlosGamero: it is the new release method, we are using changeset files to trigger release. The library is reading the file, releasing the new version + updating changelog with the md description and then removing it.

kjamrog: Ok so it's the first attempt of using it?

CarlosGamero: Not really, we have been using it a couple of months already haha

kjamrog (approved):

.changeset/text-measure-exclude-ntc.md
kjamrog: Oh I see, I didn't commit here for a while :D All good then

eduard-fa (approved):

PR #1179, opened 2026-08-13
chore: pin OpenTelemetry dependency versions

CarlosGamero:
## What

Pin OpenTelemetry dependency versions across the OTel packages so all services track a single validated set instead of resolving their own.

- **`opentelemetry-fastify-bootstrap`**: move the OpenTelemetry packages from `peerDependencies` to pinned (exact) `dependencies`. `fastify` stays a peer (it's the host framework).
- **`context-fastify-plugins`**: align the OpenTelemetry versions with the bootstrap release train — `@opentelemetry/api` `1.9.1`, experimental packages `0.221.0`, stable packages `2.10.0`, `@opentelemetry/semantic-conventions` `1.43.0`.

## Why

Currently each service can independently resolve/override the OpenTelemetry libraries. Since metrics aren't exercised locally / in PRE and there's no dedicated DataDog test env, a clean install or a well-intentioned bump can silently break the integration. Centralizing and pinning the versions means services stick to a known-good set; we bump the shared packages only after validating a newer version.

Follow-up to the discussion on #1100.

## Changeset

- `minor` for `@lokalise/opentelemetry-fastify-bootstrap`
- `major` for `@lokalise/context-fastify-plugins`

## AI Assistance Tracking


packages/app/opentelemetry-fastify-bootstrap/package.json
```diff
@@ -42,25 +42,21 @@
...
+        "@opentelemetry/sdk-trace-base": "2.10.0"
+    },
+    "peerDependencies": {
         "fastify": "^5.0.0"
     },
     "devDependencies": {
         "@biomejs/biome": "^2.5.7",
         "@lokalise/biome-config": "workspace:^",
         "@lokalise/tsconfig": "workspace:^",
         "@opentelemetry/api": "^1.9.1",
```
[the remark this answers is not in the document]
CarlosGamero: Good catch, thanks. We addressed the underlying concern (`^1.9.1` allowing an incompatible `1.10.0`) by **pinning `@opentelemetry/api` to exact `1.9.1`** and moving it into `dependencies` (removed from `devDependencies`).

We went with an exact pin in `dependencies` instead of a `>=1.9.0 <1.10.0` range in `peerDependencies` on purpose:

- Consistency with the rest of this PR — all OpenTelemetry packages are now pinned and declared as `dependencies` so every service tracks a single validated set instead of resolving its own.
- Putting it in `dependencies` (not `peerDependencies`) guarantees the compatible `api` version for consumers too, without the friction of an exact peer requirement.

.changeset/pin-otel-dependencies.md
```diff
@@ -0,0 +1,9 @@
+---
+"@lokalise/opentelemetry-fastify-bootstrap": minor
+"@lokalise/context-fastify-plugins": major
```
ppletscher, line 3: seems to contradict the PR body?

> minor for both packages (dependency contract change).

packages/app/opentelemetry-fastify-bootstrap/package.json
```diff
@@ -42,25 +42,21 @@
         "package-version": "echo $npm_package_version",
         "postversion": "biome check --write package.json"
     },
-    "dependencies": {},
-    "peerDependencies": {
+    "dependencies": {
```
ppletscher, line 45: FYI, CC raised these:

- Pinning @opentelemetry/api as an exact regular dependency is the one debatable call. OTel's own guidance is that libraries should take the API as a peer, because it's meant to be a process-wide singleton. In practice it's mostly harmless — the API registers itself into a global (Symbol.for('opentelemetry.js.api.1')), so duplicate copies still share context — but if a service ends up with api@1.9.0 from dd-trace/Prisma and api@1.9.1 pinned here, pnpm will install both copies. 
- Exact pins defeat deduplication by design. If a service still has its own @opentelemetry/sdk-trace-base@^2.x alongside this library, they'll get two copies, and cross-copy instanceof checks (span processors, samplers) can fail in confusing ways. The implicit contract is now "services must drop their direct OTel deps and let these packages own them" — that migration expectation would be good to state in the changeset/README.

The second point makes me think this should maybe be `major`?

ppletscher (approved):
LGTM, but take it with a grain of salt and see the comments below.

.changeset/pin-otel-dependencies.md
CarlosGamero: My bad, they were both minor before and forgot to update PR description, thank you!

packages/app/opentelemetry-fastify-bootstrap/package.json
CarlosGamero: > Pinning @opentelemetry/api as an exact regular dependency is the one debatable call. OTel's own guidance is that libraries should take the API as a peer, because it's meant to be a process-wide singleton. In practice it's mostly harmless — the API registers itself into a global (Symbol.for('opentelemetry.js.api.1')), so duplicate copies still share context — but if a service ends up with api@1.9.0 from dd-trace/Prisma and api@1.9.1 pinned here, pnpm will install both copies.

Good call — moved @opentelemetry/api to peerDependencies (>=1.9.0 <1.10.0, the range the SDK graph already imposes) so it stays a single shared singleton instead of a second pinned copy. Kept it in devDependencies for our own build/tests. The SDK packages  stay pinned as dependencies — those aren't singletons. Also dropped it from context-fastify-plugins entirely, since it never imports api directly (it was only a transitive pin).

> Exact pins defeat deduplication by design. If a service still has its own @opentelemetry/sdk-trace-base@^2.x alongside this library, they'll get two copies, and cross-copy instanceof checks (span processors, samplers) can fail in confusing ways. The implicit contract is now "services must drop their direct OTel deps and let these packages own them" — that migration expectation would be good to state in the changeset/README.

Good point Added a Migration section to the changeset and a "This package owns the OpenTelemetry SDK versions" note to both package READMEs, stating that services should drop their own direct @opentelemetry/* SDK deps and let these packages own them

I don’t consider this a major change. Having duplicates isn’t a breaking change, although I agree it’s debatable. Let’s see what @kibertoad thinks :D

CatchMe2 (approved):

PR #1182, opened 2026-08-18
feat: EXP-2240 api contracts visibility

CarlosGamero:
## Context

[EXP-2240](https://lokalise.atlassian.net/browse/EXP-2240): our services auto-generate their OpenAPI docs from every registered route, so internal / backend-for-frontend (BFF) endpoints — consumed only by our own frontends, where we want the freedom to introduce breaking changes — end up in the published API document.

This PR is **step 1 of 3** and covers `@lokalise/api-contracts` only: it introduces the route `visibility` concept on API contracts. Follow-up PRs will make `@lokalise/fastify-api-contracts` honor it during OpenAPI generation (`schema.hide`), and then bring `opinionated-machine` in line.

## Changes

**New `visibility` field (`'public' | 'internal'`, exported `RouteVisibility` type)**

- Added to every contract type: `CommonRouteDefinition` (legacy REST), `SSEContractDefinition` / `AnySSEContractDefinition`, `DualModeContractDefinition` / `AnyDualModeContractDefinition`, and `CommonApiContract` (new `defineApiContract` system).
- Optional in every builder input, **required on every built contract**: all builders (`buildGetRoute` / `buildPayloadRoute` / `buildDeleteRoute`, `buildRestContract`, `buildContract`, `buildSseContract`, `defineApiContract`) stamp it, defaulting to `'public'`. `'internal'` will be excluded from generated OpenAPI docs by the generators (follow-up PR); it has no runtime effect.

**Contract types tightened: built contracts always carry their parsing fields**

- `isEmptyResponseExpected` / `isNonJSONResponseExpected` are now required on built REST contract types (builders always stamped their defaults at runtime; the types now reflect it). Builder inputs keep them optional via a new `MayOmit` type util (mirrors `@lokalise/universal-ts-utils`; kept local so the package stays dependency-free).
- Dual-mode contracts now declare the flags too: `isEmptyResponseExpected` is configurable (default `false`) — enabling dual-mode routes whose sync branch responds 204/no-body, previously inexpressible — and `isNonJSONResponseExpected` is always `false` (the sync response is JSON by construction). This preserves dual-mode → REST structural assignability. Pure SSE contracts carry `visibility` but no flags (they don't apply to event streams).
- `defineApiContract` now returns a shallow copy with `visibility` stamped instead of the input reference; its input type was renamed to `ApiContractConfig` accordingly.

**Docs & release**

- README: new "Route visibility" section + `visibility` in the all-fields reference.
- Changeset: `minor`. Contracts built via builders need no changes; hand-written contract literals must add the newly-required fields (documented in the changeset).

## Testing

- TDD throughout: runtime + type-level (`expectTypeOf`) tests per builder, extended within the existing spec structure; snapshots updated; new type tests for all `typeUtils`.
- Full verification: 867 tests in `api-contracts`, plus the complete suites of all workspace consumers against the new types: `universal-testing-utils` (161), `fastify-api-contracts` (440), `backend-http-client` (486), `frontend-http-client` (302).
- An 8-angle code review was run over the full diff; the one confirmed cross-package regression it found (dual-mode contracts losing REST assignability) is fixed in this PR via the dual-mode flags above.

## Checklist

## AI Assistance Tracking

[EXP-2240]: https://lokalise.atlassian.net/browse/EXP-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/api-contracts/src/sse/dualModeContracts.ts
```diff
@@ -83,4 +93,10 @@ export type AnyDualModeContractDefinition = {
   description?: string
   summary?: string
   tags?: readonly string[]
+  // 'internal' excludes the route from generated OpenAPI docs; builders default it to 'public'
+  visibility: RouteVisibility
+  // Whether the sync (JSON) response may be empty (204); builder defaults it to false
+  isEmptyResponseExpected: boolean
+  // The sync response of a dual-mode route is always JSON; carried for REST-contract compatibility
+  isNonJSONResponseExpected: false
```
CarlosGamero, line 103: I found this was missing here, and decided to already fix it and align it with `CommonRouteDefinition`

packages/app/api-contracts/src/sse/dualModeContracts.ts
```diff
@@ -59,6 +63,12 @@ export type DualModeContractDefinition<
   description?: string
   summary?: string
   tags?: readonly string[]
+  // 'internal' excludes the route from generated OpenAPI docs; builders default it to 'public'
+  visibility: RouteVisibility
+  // Whether the sync (JSON) response may be empty (204); builder defaults it to false
+  isEmptyResponseExpected: boolean
+  // The sync response of a dual-mode route is always JSON; carried for REST-contract compatibility
```
kjamrog: 🟢 This comment is a bit confusing, maybe:

```suggestion
  // Whether the sync response of a dual-mode route may have non-JSON format; carried for REST-contract compatibility
```

kjamrog (approved):

.changeset/add-route-visibility-to-api-contracts.md
```diff
@@ -0,0 +1,12 @@
+---
+"@lokalise/api-contracts": minor
```
kibertoad, line 2: **medium — this should be `major`, not `minor`.**

The change makes previously-optional properties required on exported types: `visibility` on `CommonRouteDefinition`, `SSEContractDefinition`, `AnySSEContractDefinition`, `DualModeContractDefinition`, `AnyDualModeContractDefinition` and `CommonApiContract`, plus `isEmptyResponseExpected` / `isNonJSONResponseExpected` on the REST and dual-mode shapes.

A consumer with

```ts
const route: GetRouteDefinition<typeof schema> = {
  method: 'get',
  pathResolver,
  successResponseBodySchema: schema,
}
```

or a wrapper function annotated to return one of these types, stops compiling after a `^7` minor update. The closing note in this changeset describes exactly that break, which is the definition of a major bump.

packages/app/api-contracts/src/sse/dualModeContracts.ts
```diff
@@ -59,6 +63,12 @@ export type DualModeContractDefinition<
   description?: string
   summary?: string
   tags?: readonly string[]
+  // 'internal' excludes the route from generated OpenAPI docs; builders default it to 'public'
+  visibility: RouteVisibility
+  // Whether the sync (JSON) response may be empty (204); builder defaults it to false
+  isEmptyResponseExpected: boolean
```
kibertoad: **medium — `boolean` here widens the response body type of every dual-mode contract used with a REST client helper.**

The REST types thread `IsEmptyResponseExpected` through as a generic parameter; this hardcodes `boolean`. `RequestResultDefinitiveEither` in `backend-http-client` branches on a naked `IsEmptyResponseExpected extends true ? ... : ...`, so a `boolean` distributes over `true | false` and the result becomes `RequestResult<T | null> | RequestResult<T>`.

Verified concretely. On `main` this compiles:

```ts
const x: { status: string } = (
  await sendByGetRoute(client, dualModeContract, {}, { requestLabel: 'x' })
).result.body
```

On this branch the same code fails with `Type '{ status: string; } | null' is not assignable`. So every dual-mode contract passed to a REST client helper now yields a nullable body, including ones that never return 204.

Threading `IsEmptyResponseExpected` through as a generic (the way `CommonRouteDefinition` does) keeps the precise type.

packages/app/api-contracts/src/sse/dualModeContracts.ts
```diff
@@ -83,4 +93,10 @@ export type AnyDualModeContractDefinition = {
...
```
kibertoad, line 101: Same as the comment on `DualModeContractDefinition` above: a fixed `boolean` for `isEmptyResponseExpected` distributes over `true | false` in `RequestResultDefinitiveEither`, widening the client-side body type to `T | null`. Both declarations need the generic parameter.

packages/app/api-contracts/src/sse/sseContractBuilders.ts
```diff
@@ -461,6 +478,10 @@ export function buildSseContract(
       responseBodySchemasByStatusCode: (config as { responseBodySchemasByStatusCode?: unknown })
         .responseBodySchemasByStatusCode,
       isDualMode: true,
+      isEmptyResponseExpected:
```
kibertoad, line 489: **low — the changeset claim about 204 dual-mode routes does not hold as written.**

The changeset says this enables "dual-mode routes whose sync branch responds 204/no-body, previously inexpressible", but `buildSseContract` only takes the dual-mode branch when `successResponseBodySchema !== undefined`. Omitting the schema still produces an SSE-only contract with no such field, so the flag can only ever be set alongside a declared sync response schema.

That combination (what the new `reflects explicit isEmptyResponseExpected value` test in `sseContractBuilders.spec.ts` constructs) yields a contract telling clients to expect an empty body while carrying a schema the client never applies, since `backend-http-client` short-circuits at `result.statusCode === 204 && isEmptyResponseExpected`.

Either let the config produce a body-less dual-mode contract, or drop the claim from the changeset.

kibertoad (commented):
Code review: the runtime change is additive and correct (every builder stamps `visibility`, `buildRestContract` still defaults `isEmptyResponseExpected` to `method === 'delete'`, and `fastify-api-contracts` maps contract fields explicitly so `visibility` cannot leak into generated schemas). `tsc --noEmit` is clean on `api-contracts` and all four workspace consumers, and the `api-contracts` / `fastify-api-contracts` / `universal-testing-utils` suites pass. Three type-level findings inline.

.changeset/add-route-visibility-to-api-contracts.md
CarlosGamero: This is true, but according to our docs, the way to create contracts is by using our builder methods. So if someone is doing what you shared, they’re going against the suggested way of using the lib.

So, while you’re right that the code you shared won’t compile, it’s also not using the provided buildert. That’s why I was debating between a major and a minor bump.

I ended up choosing a minor bump to avoid the chain of library updates we’d otherwise need to perform.

That being said, if you still think my argument does not makes sense, I am happy to have it as a major and address all the required updates across our libs

VSevostianov (approved):

packages/app/api-contracts/src/sse/dualModeContracts.ts
CarlosGamero: Good catch, thanks — fixed in bbe7a16b39f4df5753e1f6046fe305cdac528044 by threading `IsEmptyResponseExpected extends boolean = false` through `DualModeContractDefinition`, both dual-mode configs and the `buildSseContract` overloads, mirroring how `CommonRouteDefinition` does it. Omitted → literal `false` (your repro types back to `{ status: string }`, verified against `sendByGetRoute`), explicit `true` → literal `true`. Type tests added for both cases.

One deliberate deviation from the suggestion: `AnyDualModeContractDefinition` keeps `boolean` — it's the type-erased wildcard (its schemas are plain `ZodTypeAny` too) used in generic constraints, while inference into the client helpers flows through the generic definition. Happy to thread it there as well if you see an inference path going through `Any*`.

packages/app/api-contracts/src/sse/dualModeContracts.ts
CarlosGamero: Addressed together with the comment above (bbe7a16b39f4df5753e1f6046fe305cdac528044): the generic is threaded through `DualModeContractDefinition` and the builder overloads. `AnyDualModeContractDefinition` intentionally stays `boolean` as the type-erased wildcard — see the reasoning in the other thread.

packages/app/api-contracts/src/sse/sseContractBuilders.ts
CarlosGamero: Fair point — the claim overstated it. Reworded in d773ec6d190a99b8cb620ab910e951a914e7c4bd: a body-less dual-mode contract is indeed still not expressible (the dual-mode branch requires `successResponseBodySchema`), and we're not planning to allow one in this PR. What the flag covers is the mixed case: a sync branch that may respond 204 alongside its regular JSON response — the schema applies to responses with a body, and `isEmptyResponseExpected: true` lets clients accept the 204 short-circuit. The changeset now describes exactly that.

kosmos2uh (approved):

packages/app/api-contracts/src/sse/dualModeContracts.ts
CarlosGamero: Fixed, thank youuu

CatchMe2 (approved):

kibertoad (approved):

PR #1184, opened 2026-08-19
feat: EXP-2240 Fastify api contracts - visibility

CarlosGamero:
## Context

[EXP-2240](https://lokalise.atlassian.net/browse/EXP-2240): auto-generated OpenAPI docs include every endpoint, including internal / backend-for-frontend routes that must not be part of the published API surface. #1182 introduced the `visibility` field (`'public' | 'internal'`) on API contracts; this PR is **step 2 of 3** and makes `@lokalise/fastify-api-contracts` act on it. Step 3 will bring `opinionated-machine` in line for its SSE/dual-mode builders.

## Changes

**`visibility` → `schema.hide` derivation in every route builder**

- `buildFastifyRoute`, the deprecated `buildFastifyNoPayloadRoute` / `buildFastifyPayloadRoute`, and `buildFastifyApiSchema` (which `buildFastifyApiRoute` delegates to) stamp `hide: contract.visibility === 'internal'` on the route schema. `schema.hide` is the convention natively honored by `@fastify/swagger` and `fastify-type-provider-zod`'s transform, so internal routes disappear from the generated OpenAPI document with **zero per-service changes**, while still being registered and served.
- `hide` is always an explicit boolean (`false` for public routes, previously the key didn't exist). Generated docs are byte-identical for public routes (`hide` is stripped by the generators); downstream tests snapshotting route schemas will see the new key.
- Design note: `hide` is a derived rendering instruction only. Consumers needing the semantic value (auth hooks, gateways, doc tooling) read `config.apiContract.visibility` from the route — documented in the new README section. Contract-less legacy routes can use fastify-swagger's native `schema: { hide: true }`.

**Supporting changes**

- `@lokalise/api-contracts` peer floor raised to `>=7.2.0` — the first version whose builders stamp `visibility`. With older versions the derivation would silently evaluate to `false` and leak internal routes into docs.
- Unified the duplicated `ExtendedFastifySchema` type (the local copy in `buildFastifyApiSchema.ts` had already drifted from `types.ts`); only the `types.ts` declaration remains and it is the only one that was ever exported from the package barrel.
- Removed the dead `describe` field from `ExtendedFastifySchema` and the builders: a day-one typo for `description` (the real OpenAPI field, set alongside it since); nothing ever consumed it.
- README: new "Route visibility" section documenting the derivation and where to read the semantic value.
- Changeset: `minor`.

## Testing

- Per-builder unit tests inside the existing spec structure: internal → `hide: true` and public → `hide: false` for all four builders (both branches on each deprecated builder). Mutation-checked: removing a stamp fails exactly the corresponding tests.
- Full package suite green (456 tests, typecheck, lint).

## Checklist

## AI Assistance Tracking

[EXP-2240]: https://lokalise.atlassian.net/browse/EXP-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/fastify-api-contracts/src/types.ts
```diff
@@ -23,10 +23,11 @@ interface FastifyContractRouteInterface<ReplyType, BodyType, ParamsType, QueryTy
  * Default fastify fields + fastify-swagger fields
  */
 export type ExtendedFastifySchema = FastifySchema & {
-  describe?: string
```
CarlosGamero, line 26: We already have `description` which is the one for OpenAPI, looks like this one is unused so cleaning it up. Please let me know if I am missing something

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
```diff
@@ -98,18 +98,15 @@ function buildResponseSchemas(contract: ApiContract): Record<string, FastifyResp
   return schemas
 }
 
-export type ExtendedFastifySchema = FastifySchema & {
-  description?: string
-  summary?: string
-  tags?: readonly string[]
-}
```
CarlosGamero, line 105: Duplicated on types.ts

packages/app/fastify-api-contracts/src/types.ts
```diff
@@ -23,10 +23,11 @@ interface FastifyContractRouteInterface<ReplyType, BodyType, ParamsType, QueryTy
...
   description?: string
   summary?: string
   tags?: readonly string[]
+  /** When true, the route will not be added to the OpenAPI docs */
+  hide?: boolean
```
CarlosGamero, line 30: Fastify swager is automatically hiding routes from OpenAPI based on this

.changeset/derive-openapi-hide-from-visibility.md
```diff
@@ -0,0 +1,10 @@
+---
+"@lokalise/fastify-api-contracts": minor
```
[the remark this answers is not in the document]
CarlosGamero: it is a compatible bump, minor is ok

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
CatchMe2: The duplication was intentional, because the shared type had `describe` which doesn't provide any value for the new api-contracts, and I didn't want to remove it from shared types to avoid major breaking change.

Also, duplicating such types will make removing the old contracts easier (once we migrate everything to the new approach)

CarlosGamero: yeah, I was trying to find an answer to why describe is there, I think it is not used so decided to remove it. 
if removing it is ok, then we can reuse same type, I don't think having a single type will make the future clean-up more complex tbh

CatchMe2 (approved):

packages/app/fastify-api-contracts/src/fastifyRouteBuilder.ts
```diff
@@ -289,10 +289,10 @@ export function buildFastifyRoute(
       params: apiContract.requestPathParamsSchema,
       querystring: apiContract.requestQuerySchema,
       headers: apiContract.requestHeaderSchema,
-      describe: apiContract.description,
       description: apiContract.description,
       summary: apiContract.summary,
       tags: apiContract.tags,
+      hide: apiContract.visibility === 'internal',
```
kibertoad, line 295: This derivation fails open when `visibility` is missing.

Peer ranges are advisory: npm and yarn only warn on an unmet peer, and pnpm's `strict-peer-dependencies` defaults to false. A service that resolves `@lokalise/api-contracts` 7.1.x gets contracts with no `visibility` field, so `apiContract.visibility` is `undefined`, every route lands with `hide: false`, and an endpoint authored as internal is published in the OpenAPI document with no error, warning, or failing test. The implementation signature takes `apiContract: any`, so nothing catches the version skew at runtime either.

Since the feature exists to keep internal endpoints off the published surface, throwing (or at least logging) when `visibility` is absent would turn a silent leak into a loud failure.

Same code at `fastifyApiContracts.ts:160`, `fastifyApiContracts.ts:285`, and `api-contracts/buildFastifyApiSchema.ts:108`.

packages/app/fastify-api-contracts/src/fastifyRouteBuilder.spec.ts
```diff
@@ -575,5 +575,32 @@ describe('buildFastifyRoute', () => {
 
       expect(route.schema).not.toHaveProperty('tags')
     })
+
+    it('maps internal visibility to schema.hide true', () => {
```
kibertoad, line 579: All four new tests assert only the intermediate `schema.hide` value; none registers a route on a Fastify instance and checks the generated document.

The behavior being shipped depends on a cross-library convention: `fastify-type-provider-zod`'s transform honoring `schema.hide`, then `@fastify/swagger`. The installed 7.0.0 does honor it (it destructures `hide` out and returns `{ hide: true }` early when truthy), so this is a coverage gap rather than a defect. But if a future transform version stops handling `hide`, or the peer floor moves, internal routes silently reappear in published docs and every test in this package still passes.

One integration test asserting the internal path is absent from the generated spec would pin the contract that actually matters.

packages/app/fastify-api-contracts/README.md
```diff
@@ -390,6 +391,20 @@ const routes = [
 ]
 ```
 
+### Route visibility
+
+Every route builder derives the fastify-swagger `schema.hide` flag from the contract's `visibility` field (`'public' | 'internal'`, always stamped by the contract builders): routes from `visibility: 'internal'` contracts get `hide: true` and are excluded from the generated OpenAPI document, while still being registered and served as usual. Public routes carry an explicit `hide: false`.
+
+```ts
+const contract = buildGetRoute({
```
kibertoad: The only example in this section uses `buildGetRoute`, which this same README marks deprecated at lines 59 and 307.

The section sits at `###` under `## Builders`, as a sibling of `buildFastifyApiRoute` and `buildFastifyRoute`, so it reads as cross-builder guidance. A reader on the current `defineApiContract` API copies a deprecated builder out of a section that is supposed to apply to theirs. Switching the example to `defineApiContract`, as the neighboring sections do, fixes it.

kibertoad (commented):
Code review: 3 low-severity findings, nothing blocking. The rest checked out clean: all four schema-construction sites updated, the `hide` convention verified against the installed `fastify-type-provider-zod@7.0.0`, the `>=7.2.0` peer floor correct, the deleted duplicate `ExtendedFastifySchema` never reachable from the barrel, and 456 tests green with no type errors.

packages/app/fastify-api-contracts/src/fastifyRouteBuilder.ts
CarlosGamero: True, but I think fail-open is fine here. In the skew you describe (api-contracts 7.1 + latest fastify-api-contracts), visibility doesn't exist as a concept yet (7.1 contracts can't even declare a route as internal) so there's nothing to leak. Defaulting hide to false reproduces exactly what 7.1 semantics were: everything documented. The moment a team can actually write visibility: 'internal', they're on 7.2+ types and the field is required, so the guard is already there at compile time.

CarlosGamero: True, but I think fail-open is fine here. In the skew you describe (api-contracts 7.1 + latest fastify-api-contracts), `visibility` doesn't exist as a concept yet — 7.1 contracts can't even declare a route as internal, so there's nothing to leak. Defaulting `hide` to `false` reproduces exactly what 7.1 semantics were: everything documented. The moment a team can actually write `visibility: 'internal'`, they're on 7.2+ types and the field is required, so the guard is already there at compile time.

packages/app/fastify-api-contracts/src/fastifyRouteBuilder.spec.ts
CarlosGamero: Added in 7b278ebe — `openApiVisibility.spec.ts` registers a public and an internal route on a real Fastify app with `@fastify/swagger` + the zod transform, and asserts the internal path is absent from the generated document (public one present).

packages/app/fastify-api-contracts/README.md
CarlosGamero: Fixed in 11fee291 — the example now uses `defineApiContract`.

kosmos2uh (approved):

PR #1202, opened 2026-08-20
feat: EXP-2240 Contracts mandatory visibility

CarlosGamero:
## Checklist

## AI Assistance Tracking

### DIFF
--- a/packages/app/api-contracts/src/apiContracts.spec.ts
+++ b/packages/app/api-contracts/src/apiContracts.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, expectTypeOf, it } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import { z } from 'zod/v4'
 import {
   buildDeleteRoute,
@@ -6,7 +6,6 @@ import {
   buildPayloadRoute,
   describeContract,
   mapRouteToPath,
-  type RouteVisibility,
 } from './apiContracts.ts'
 
 const BODY_SCHEMA = z.object({})
@@ -31,6 +30,7 @@ describe('apiContracts', () => {
   describe('buildPayloadRoute', () => {
     it('sets default change route values', () => {
       const contract = buildPayloadRoute({
+        visibility: 'internal',
         successResponseBodySchema: BODY_SCHEMA,
         requestBodySchema: BODY_SCHEMA,
         method: 'post',
@@ -44,24 +44,12 @@ describe('apiContracts', () => {
 
       expect(contract).toMatchSnapshot()
     })
-
-    it('propagates visibility to the contract', () => {
-      const contract = buildPayloadRoute({
-        successResponseBodySchema: BODY_SCHEMA,
-        requestBodySchema: BODY_SCHEMA,
-        method: 'post',
-        pathResolver: () => '/',
-        visibility: 'internal',
-      })
-
-      expectTypeOf(contract.visibility).toEqualTypeOf<RouteVisibility>()
-      expect(contract.visibility).toBe('internal')
-    })
   })
 
   describe('buildGetRoute', () => {
     it('sets default get route values', () => {
       const contract = buildGetRoute({
+        visibility: 'public',
         successResponseBodySchema: BODY_SCHEMA,
         pathResolver: () => '/',
         responseSchemasByStatusCode: {
@@ -77,6 +65,7 @@ describe('apiContracts', () => {
 
     it('resolves path params', () => {
       const contract = buildGetRoute({
+        visibility: 'public',
         successResponseBodySchema: BODY_SCHEMA,
         requestPathParamsSchema: PATH_PARAMS_SCHEMA,
         pathResolver: (pathParams) => `/users/${pathParams.userId}`,
@@ -84,21 +73,12 @@ describe('apiContracts', () => {
 
       expect(contract).toMatchSnapshot()
     })
-
-    it('propagates visibility to the contract', () => {
-      const contract = buildGetRoute({
-        successResponseBodySchema: BODY_SCHEMA,
-        pathResolver: () => '/',
-        visibility: 'public',
-      })
-
-      expect(contract.visibility).toBe('public')
-    })
   })
 
   describe('buildDeleteRoute', () => {
     it('sets default delete route values', () => {
       const contract = buildDeleteRoute({
+        visibility: 'public',
         successResponseBodySchema: BODY_SCHEMA,
         pathResolver: () => '/',
         description: 'some description',
@@ -106,21 +86,12 @@ describe('apiContracts', () => {
 
       expect(contract).toMatchSnapshot()
     })
-
-    it('propagates visibility to the contract', () => {
-      const contract = buildDeleteRoute({
-        successResponseBodySchema: BODY_SCHEMA,
-        pathResolver: () => '/',
-        visibility: 'internal',
-      })
-
-      expect(contract.visibility).toBe('internal')
-    })
   })
 
   describe('mapRouteToPath', () => {
     it('returns path without params', () => {
       const contract = buildGetRoute({
+        visibility: 'public',
         successResponseBodySchema: BODY_SCHEMA,
         pathResolver: () => '/',
       })
@@ -131,6 +102,7 @@ describe('apiContracts', () => {
 
     it('returns path with one param', () => {
       const contract = buildGetRoute({
+        visibility: 'public',
         successResponseBodySchema: BODY_SCHEMA,
         requestPathParamsSchema: PATH_PARAMS_SCHEMA,
         pathResolver: (pathParams) => `/users/${pathParams.userId}`,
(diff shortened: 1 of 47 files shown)


packages/app/api-contracts/src/apiContracts.spec.ts
```diff
@@ -44,24 +44,12 @@ describe('apiContracts', () => {
...
-
-      expectTypeOf(contract.visibility).toEqualTypeOf<RouteVisibility>()
-      expect(contract.visibility).toBe('internal')
-    })
   })
 
   describe('buildGetRoute', () => {
     it('sets default get route values', () => {
       const contract = buildGetRoute({
+        visibility: 'public',
```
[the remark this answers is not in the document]
CarlosGamero: it is covered by snapshots, same for internal as we have tests for them

packages/app/api-contracts/src/apiContracts.types.spec.ts
```diff
@@ -91,6 +94,7 @@ describe('legacy builders type inference', () => {
 
     it('buildGetRoute reflects explicit true value in type', () => {
       const contract = buildGetRoute({
+        visibility: 'public',
```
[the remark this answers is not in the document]
CarlosGamero: We do not have such tests for other props, following lib standards

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
```diff
@@ -105,7 +105,7 @@ function buildResponseSchemas(contract: ApiContract): Record<string, FastifyResp
 export function buildFastifyApiSchema(contract: ApiContract): ExtendedFastifySchema {
   const schema: ExtendedFastifySchema = {
     summary: contract.summary,
-    hide: contract.visibility === 'internal',
+    hide: contract.visibility !== 'public',
```
kibertoad, line 108: `hide` flipped from `=== 'internal'` to `!== 'public'` here (and at the matching spot in `fastifyRouteBuilder.ts:295`), while every builder in api-contracts simultaneously lost its `?? 'public'` fallback. The combination means a contract that arrives with `visibility` undefined at runtime is now excluded from the generated OpenAPI document instead of documented.

Reachable cases: a plain-JS consumer; a contract object hand-written and cast to the contract type (the 7.2.0 CHANGELOG documents hand-typed literals as a supported shape); or `buildSseContract`, whose `buildBaseFields(config: any, ...)` reads `config.visibility` off an untyped value and no longer has a guard behind it. In each of those the route silently vanishes from the published docs, with nothing logged and nothing failing.

Either keep `=== 'internal'` so an absent value still fails open, or keep `!== 'public'` and say so in the fastify-api-contracts changeset, which currently mentions only the peer bump.

packages/app/api-contracts/src/new/defineApiContract.ts
```diff
@@ -76,10 +74,7 @@ export const defineApiContract = <
   contract: Exactly<TContract, ApiContractConfig<TPathParamsSchema>> & {
     requestPathParamsSchema?: TPathParamsSchema
   },
-): TContract & { visibility: RouteVisibility } =>
-  ({ ...contract, visibility: contract.visibility ?? 'public' }) as TContract & {
-    visibility: RouteVisibility
-  }
+): TContract => contract as TContract
```
kibertoad: Two behaviour changes ride on this line, and neither is in the changeset.

**1. `visibility` now infers as a literal, breaking consumer comparisons.** The return type went from `TContract & { visibility: RouteVisibility }` to bare `TContract`, and `TContract` is a `const` type parameter, so the literal is preserved:

```ts
const c = defineApiContract({ visibility: 'public', summary: 's', method: 'get', ... })
c.visibility === 'internal'
// error TS2367: This comparison appears to be unintentional because
// the types '"public"' and '"internal"' have no overlap.
```

I reproduced that against this branch; it compiled before. The fastify-api-contracts README tells consumers to read `visibility` off the contract for auth hooks, gateways and custom doc tooling, so the path has users. Keeping the explicit `TContract & { visibility: RouteVisibility }` return annotation preserves the widening while `visibility` stays required in the input.

**2. The contract is now the caller's object, not a copy.** `{ ...contract, visibility }` became `contract as TContract`. That reverts behaviour the 7.2.0 CHANGELOG called out by name: "defineApiContract now returns a shallow copy with visibility stamped instead of returning the input object by reference". A caller that holds a reference to the config object and mutates it later now mutates the contract too.

packages/app/fastify-api-contracts/package.json
```diff
@@ -48,7 +48,7 @@
     },
     "peerDependencies": {
         "@fastify/sse": ">=0.5.0",
-        "@lokalise/api-contracts": ">=7.2.0",
+        "@lokalise/api-contracts": ">=8.0.0",
```
kibertoad, line 51: The code does not require 8.x. Everything in `src/` only reads `contract.visibility`, which every api-contracts builder has stamped since 7.2.0, so `>=7.2.0` still expresses the real runtime requirement.

With the floor at `>=8.0.0`, a consumer has to adopt the api-contracts 8 breaking change (adding `visibility` to every contract definition in their codebase) before they can pick up any future fastify-api-contracts release, including unrelated patches.

packages/app/api-contracts/src/apiContracts.types.spec.ts
```diff
@@ -99,45 +103,4 @@ describe('legacy builders type inference', () => {
       expectTypeOf(contract.isNonJSONResponseExpected).toEqualTypeOf<true>()
     })
   })
```
kibertoad, line 105: This deletion drops the `visibility types` block, and the equivalent blocks go away in `restContractBuilder.types.spec.ts` and `sseContractBuilders.types.spec.ts`. Nothing replaces them, so the suite no longer pins the invariant this PR exists to establish: a change that makes `visibility` optional again in any of the six builders would pass CI.

I verified the invariant does hold today (omitting `visibility` is a compile error in all six builders), so this is only about keeping it that way. One `@ts-expect-error` case per builder covers it. One thing to know while writing them: on the GET overloads of `buildRestContract`, `buildContract` and `buildSseContract`, overload resolution reports the omission as `Type '"get"' is not assignable to type '"patch" | "post" | "put"'` rather than naming `visibility`.

kibertoad:
One finding from the review that cannot be anchored inline, because the file is not part of this diff:

`packages/app/fastify-api-contracts/README.md:396` still says the `visibility` field is `'public' | 'internal'`, "always stamped by the contract builders". After this PR the builders stamp nothing and the caller supplies it. The api-contracts README was updated for this change; the fastify one was not.

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiSchema.ts
CarlosGamero: Keeping the fail-closed direction intentionally: an unclassified route (visibility `undefined` at runtime) should stay out of the published OpenAPI document rather than be exposed by accident. Now documented in the `fastify-api-contracts` changeset (`fastify-hide-fail-closed.md`) covering exactly the reachable cases you list (plain-JS consumers, hand-written literals, pre-7.2 contract packages), and pinned with two runtime tests that strip `visibility` from a contract and assert `hide: true` — see 8945db1d.

packages/app/fastify-api-contracts/package.json
CarlosGamero: You are right, but as it will be used to generate public doc I wanted to be explicit is this lib so whenever we want to start exposing a service our check list should have api-contracts >= 8  fastify-api-contracts >= 7 are used

packages/app/api-contracts/src/new/defineApiContract.ts
CarlosGamero: Both points addressed in 2da83c27 (refined in b0b47a79):

1. The return type is now `Omit<TContract, 'visibility'> & { visibility: RouteVisibility }`, which widens the `const`-inferred literal back to the union so consumer comparisons compile again (a plain `TContract & { visibility: RouteVisibility }` intersection doesn't work — `'public' & RouteVisibility` reduces back to the literal). Pinned with a test that asserts `route.visibility === 'internal'` compiles and that `route.visibility` equals `RouteVisibility`.
2. The shallow copy is back (`({ ...contract })`), now without a cast so the compiler verifies the assignment. Also pinned with a test that mutates the caller's config after the call and asserts the contract is unaffected.

`visibility` stays required in the input — only the output typing and copy semantics were restored.

kibertoad (approved):

packages/app/api-contracts/src/apiContracts.types.spec.ts
CarlosGamero: Restored in 4f868c94: each types spec now pins the invariant with one `@ts-expect-error` guard per builder (`buildGetRoute`/`buildPayloadRoute`/`buildDeleteRoute` in `apiContracts.types.spec.ts`, plus `buildContract`, `buildRestContract` and `buildSseContract` in their respective specs), nested under each file's main describe. If `visibility` ever becomes optional again, the directives turn unused and CI fails.

Implementation note on the overload caveat you flagged: with an inline object literal the TS2769 elaboration lands on the property lines rather than the call line, escaping the directive — so the configs are extracted to a variable (`method: 'get' as const`) and the directive sits on the single-line call.

PR #1203, opened 2026-08-20
fix: run job execution within its transaction's observability context

kibertoad:
## Changes

Follow-up to lokalise/fastify-extras#456, which fixed background transactions being swallowed by an unrelated trace. This fixes the other half of the problem: the transaction started for a job was never made the *active* one, so nothing the job did was traced as part of it.

`TransactionObservabilityManager.start`/`stop` are two separate calls, which cannot express a scope — so `start()` could only create the transaction, never activate it. Consequences:

- every span produced while the job ran (database queries, outgoing HTTP calls, nested spans) became a **detached root** instead of a child of the job transaction, leaving each job as a single childless span with no usable trace;
- `addCustomAttribute` / `setUserID` / `setControllerName`, which read the active span, were silently no-ops during job execution.

Job execution is now wrapped in the manager's `runInSpanContext(uniqueTransactionKey, fn)` when it implements one — the OpenTelemetry manager in `@lokalise/fastify-extras` does. `runInSpanContext` is an *optional* member of the `TransactionObservabilityManager` contract, so managers without it just run the job as before and a job never fails because of its observability tooling. Applies to both processors and to periodic jobs.

This requires `@lokalise/node-core` >= 14.9.1, which declares the optional method and ships the `runInTransactionContext` helper that applies it (preferring it over an inline `manager.runInSpanContext?.(key, fn) ?? fn()`, which would run the job twice whenever it legitimately resolves to `undefined` or `null`). The dependency is bumped here and the helper this PR originally hand-rolled in `src/observability/transactionContext.ts` is dropped in favour of the published one.

Second fix: the outcome is now passed to `stop()`. Previously every transaction was recorded as successful, including failed ones, so trace error rates for background work were always zero. BullMQ control-flow errors (`DelayedError`, `WaitingChildrenError`, `RateLimitError`) are cooperative deferrals rather than failures, so those still count as successful — matching how `jobAttemptError` already treats them. The thrown value reaches `jobEnd` wrapped in a `JobFailure`, so its *presence* marks the failure and its contents only classify it: a job that throws a falsy value (`throw undefined`, `Promise.reject()`) is still recorded as failed.

The existing `isSuccess` field on the "Finished job" log line is left alone (it reports `job.progress === 100`), to avoid changing a log shape consumers may parse.

#### Verification

`tsc` and `biome check` clean against node-core 14.9.1. The monitor, barrier, error and stalled specs pass locally after the review fixes. Test suite (last full local run, before the 14.9.1 bump): 244 passed, 1 skipped, plus one failure in `ignores missing job error during automatic purging` — that test is flaky on `main` independently of this change (it races `dispose()` closing the Redis connection against the purge, and fails with `Connection is closed.`). I confirmed it fails identically with `main`'s version of the package checked out. CI validates the suite on this revision.

## Checklist

## AI Assistance Tracking


packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -257,34 +257,39 @@ export abstract class AbstractBackgroundJobProcessorNew<
...
+        await job.updateProgress(100)
+        return result
+      })
     } catch (error) {
+      jobError = error
       this.monitor.jobAttemptError(job, error, requestContext)
       throw error
     } finally {
-      this.monitor.jobEnd(job, requestContext)
+      this.monitor.jobEnd(job, requestContext, jobError)
```
CarlosGamero: I am actually wondering if we could wrap the whole `processInternal` with the r`unInJobContext`, it might make this logic a bit simpler. Not sure just thinking out load

CarlosGamero: process internal can potentially return `Either`

packages/app/background-jobs-common/src/background-job-processor/monitoring/BackgroundJobProcessorMonitor.ts
```diff
@@ -112,15 +113,29 @@ export class BackgroundJobProcessorMonitor<
...
       },
       `Finished job ${job.name}`,
     )
-    this.transactionObservabilityManager.stop(resolveJobId(job))
+    /**
+     * BullMQ control-flow errors are cooperative deferrals (delayed, waiting for children,
+     * rate limited) rather than failures, so the transaction did not actually fail.
+     */
+    const wasSuccessful = !error || isBullmqControlFlowError(error)
+    this.transactionObservabilityManager.stop(resolveJobId(job), wasSuccessful)
```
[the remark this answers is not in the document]
kibertoad: Valid, fixed in c00f5392 — with a wrapper instead of a `didThrow` flag, so the two can't disagree:

```ts
export type JobFailure = { error: unknown }

public jobEnd(job: JobType, requestContext: RequestContext, failure?: JobFailure): void {
  const wasSuccessful = !failure || isBullmqControlFlowError(failure.error)
```

The presence of the wrapper marks the failure, its contents only classify it, so `throw undefined`, `Promise.reject()` and `Promise.reject(0)` are recorded as failures while control-flow deferrals still count as successful. Both processors now build it in the `catch` block (`jobFailure = { error }`) and pass it from `finally`, which keeps `stop()` on a single code path. Covered by a parametrized test over `undefined`, `null`, `0` and `''`.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
kibertoad: Good call — done in c00f5392, as far as the span allows. `runInJobContext` can only activate a transaction that already exists, so `jobStart` has to stay outside; everything after it is now inside a single wrap:

```ts
private async processInternal(job: JobType): Promise<JobReturn> {
  const requestContext = this.monitor.getRequestContext(job)
  this.monitor.jobStart(job, requestContext)

  // everything below is traced as part of the transaction started above
  return await this.monitor.runInJobContext(job, async () => {
    let jobFailure: JobFailure | undefined
    try { ... } catch (error) { ... } finally { this.monitor.jobEnd(job, requestContext, jobFailure) }
  })
}
```

Two things fall out of it: one nesting level instead of a wrap inside a `try`, and `jobAttemptError`/`jobEnd` now log from inside the span context too, so those lines are trace-correlated like the rest of the job. `jobStart` leaving the `try` also means a manager that throws on `start()` no longer leads to a `stop()` for a transaction that was never started.

On `Either`: right, `JobReturn` is unconstrained, so a job can report failure in its return value. `runInJobContext<T>` is generic and passes the value straight through, so that keeps working. I deliberately did not inspect it for the transaction outcome — a returned failure `Either` is a *completed* job as far as BullMQ is concerned (no retry, `onSuccess` hook, job data purged), and having the transaction disagree with the job state seemed worse than the missing signal. Jobs that want the failure traced can throw, or add their own attribute via `addCustomAttributes` — now that the context propagates, that actually reaches the job span.

CarlosGamero (approved):

PR #1211, opened 2026-08-24
feat: support bullmq 6 and ioredis 6

kibertoad:
## What

BullMQ 6 and ioredis 6 support across the three packages that touch them.

| Package | Change |
| --- | --- |
| `background-jobs-common` | `bullmq` peer `^5.28.2` -> `^5.28.2 \|\| ^6.0.0`, `ioredis` peer `^5.4.1` -> `^5.4.1 \|\| ^6.0.0` |
| `healthcheck-utils` | `ioredis` peer `^5.4.1` -> `^5.4.1 \|\| ^6.0.0` |
| `fastify-bullboard-plugin` | `@bull-board` 8 -> 9 (which supports bullmq 6), devDependency on bullmq 6 |

Dev dependencies move to bullmq 6.2.0 / ioredis 6.0.0, so CI exercises the new majors.

## BullMQ 6 changes that touched `background-jobs-common`

- `paused` was removed from `JobType` (v6 pauses through queue metadata instead of a separate list). The
  pending-job type list moved to a shared `PENDING_JOB_TYPES` constant. `getJobCount` still asks for the
  `paused` count: on v6 that key is never written, so it contributes 0, and on v5 the behaviour is unchanged.
- `debounce` and `repeat` were dropped from `JobsOptions`. `stripChildOnlyFields` deletes them through a
  record, so it keeps guarding against a JS caller smuggling them onto a flow child under either major.
- `FlowChildJob` was renamed to `FlowJobNode`. The type is now derived from `FlowJob['children']`, which
  resolves to the right thing on both.
- Legacy repeatable jobs are gone. The two repeatable tests now create the schedule through
  `queue.upsertJobScheduler`, which exists in both majors.

## ioredis 6 change

`zrange`'s stop argument narrowed to `string | Buffer`, so `backgroundJobProcessorGetActiveQueueIds` and its
tests pass `'-1'` instead of `-1`.

## `fastify-bullboard-plugin`

`@bull-board` 9 peers `bullmq ^5.79.2 || ^6.0.0`. One source change: `QueueConstructor` no longer declares the
third constructor parameter. BullMQ 6 replaced it (`Connection`) with a `BackendFactory`, and the plugin never
passed it, so the type now mirrors `QueueProConstructor` and accepts BullMQ's `Queue` on both majors. Marked
`major` in the changeset because @bull-board 9 ships a redesigned dashboard UI and the exported type narrowed.

## Is v5 still genuinely supported?

Yes. Every change is behaviour-identical on v5: the same six job types are passed, the same four job options are
deleted, `NonNullable<FlowJob['children']>[number]` resolves to exactly `FlowChildJob`, `zrange` sends the same
wire command, and `upsertJobScheduler` has existed since bullmq 5.16.

Caveat worth naming: CI runs one version, so the 5.x side is a point-in-time check rather than a standing
guarantee. A version matrix can be added if that should be locked down.

## Verification

- `tsc --noEmit` clean for all three packages against bullmq 6.2.0 + ioredis 6.0.0 **and** against
  bullmq 5.81.3 + ioredis 5.11.1.
- Test suites pass on both lines. The two `ignores missing job error during automatic purging` failures seen
  locally reproduce on `main` unchanged and are unrelated to this PR.
- `pnpm run build` green across the workspace.

Runtime dependencies (`@lokalise/node-core`, `redis-semaphore`, `ts-deepmerge`) were already at their latest
releases, so nothing to bump there.

## AI Assistance Tracking

### DIFF
--- a/.changeset/bullmq-6-support.md
+++ b/.changeset/bullmq-6-support.md
@@ -1,8 +1,11 @@
 ---
 "@lokalise/background-jobs-common": minor
+"@lokalise/healthcheck-utils": minor
 ---
 
-Support BullMQ 6 and ioredis 6 alongside 5: peer ranges widen to `^5.28.2 || ^6.0.0` and `^5.4.1 || ^6.0.0`.
+Support BullMQ 6 and ioredis 6 alongside 5. `background-jobs-common` widens its peer ranges to
+`^5.28.2 || ^6.0.0` (bullmq) and `^5.4.1 || ^6.0.0` (ioredis); `healthcheck-utils` widens its ioredis peer
+to `^5.4.1 || ^6.0.0`.
 
 BullMQ 6 removes `paused` from `JobType`, drops `debounce` and `repeat` from `JobsOptions`, and renames the
 `FlowChildJob` type to `FlowJobNode`. The library now derives the flow child type from `FlowJob` and deletes the
--- a/.changeset/bullmq-6-support.md
+++ b/.changeset/bullmq-6-support.md
@@ -1,0 +1,10 @@
+---
+"@lokalise/background-jobs-common": minor
+---
+
+Support BullMQ 6 and ioredis 6 alongside 5: peer ranges widen to `^5.28.2 || ^6.0.0` and `^5.4.1 || ^6.0.0`.
+
+BullMQ 6 removes `paused` from `JobType`, drops `debounce` and `repeat` from `JobsOptions`, and renames the
+`FlowChildJob` type to `FlowJobNode`. The library now derives the flow child type from `FlowJob` and deletes the
+removed job options through a record, so one source tree compiles against both majors. `getJobCount` still counts
+the `paused` state; on BullMQ 6 that key is never written and contributes 0.
--- a/packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
+++ b/packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
@@ -1,5 +1,4 @@
 import type {
-  FlowChildJob,
   FlowJob,
   FlowOpts,
   FlowProducer,
@@ -29,13 +28,22 @@ import type {
   SupportedQueueIds,
 } from './types.ts'
 
+/**
+ * A node below the flow root. BullMQ v5 exports this as `FlowChildJob`, v6 as
+ * `FlowJobNode`; reading it off `FlowJob` gives the same shape on both majors.
+ */
+type BullmqFlowChildJob = NonNullable<FlowJob['children']>[number]
+
 const stripChildOnlyFields = (opts: JobsOptions): JobsOptions => {
-  const next: JobsOptions = { ...opts }
+  // `debounce` and `repeat` were dropped from JobsOptions in BullMQ v6, so
+  // delete through a record: TypeScript no longer knows those keys, but a JS
+  // caller can still pass them in.
+  const next = { ...opts } as Record<string, unknown>
   delete next.deduplication
   delete next.debounce
   delete next.parent
   delete next.repeat
-  return next
+  return next as JobsOptions
 }
 
 /**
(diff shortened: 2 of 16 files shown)


.changeset/bullmq-6-support.md
```diff
@@ -0,0 +1,10 @@
+---
+"@lokalise/background-jobs-common": minor
+---
```
CarlosGamero, line 4: Probably worth doing together with `healthcheck-utils` and `fastify-bullboard-plugin`

kibertoad: Done, both are in now.

`healthcheck-utils`: ioredis peer widened to `^5.4.1 || ^6.0.0`, devDependency on 6. No source change needed.

`fastify-bullboard-plugin`: moved to `@bull-board` 9 (peers `bullmq ^5.79.2 || ^6.0.0`) and bullmq 6. One source change: `QueueConstructor` dropped its third constructor parameter. BullMQ 6 replaced `Connection` with a `BackendFactory`, and the plugin never passed that argument, so the type now mirrors `QueueProConstructor` right above it and accepts BullMQ's `Queue` on both majors. Marked `major` in the changeset since @bull-board 9 ships a redesigned dashboard UI and the exported type narrowed.

Verified all three packages: `tsc --noEmit` and tests green against bullmq 6.2.0 + ioredis 6.0.0 *and* against bullmq 5.81.3 + ioredis 5.11.1.

Worth flagging: CI only exercises one version (6), so the 5.x side is a point-in-time check rather than a standing guarantee. Happy to add a version matrix if you want that locked down, or we can stop advertising v5 support

packages/app/background-jobs-common/src/background-job-processor/managers/FlowManager.ts
```diff
@@ -29,13 +28,22 @@ import type {
...
+ * `FlowJobNode`; reading it off `FlowJob` gives the same shape on both majors.
+ */
+type BullmqFlowChildJob = NonNullable<FlowJob['children']>[number]
+
 const stripChildOnlyFields = (opts: JobsOptions): JobsOptions => {
-  const next: JobsOptions = { ...opts }
+  // `debounce` and `repeat` were dropped from JobsOptions in BullMQ v6, so
+  // delete through a record: TypeScript no longer knows those keys, but a JS
+  // caller can still pass them in.
+  const next = { ...opts } as Record<string, unknown>
```
CarlosGamero: I am not sure about this, probably I would avoid the casting as callers are already going to need to adjust input right?

kibertoad: Casts are gone. The local is now widened by the two keys BullMQ 6 removed, so one `delete` list still serves both majors and `deduplication`/`parent` keep their real types:

```ts
type RemovedInBullmq6JobOptions = { debounce?: unknown; repeat?: unknown }

const stripChildOnlyFields = (opts: JobsOptions): JobsOptions => {
  const next: JobsOptions & RemovedInBullmq6JobOptions = { ...opts }
  delete next.deduplication
  delete next.debounce
  delete next.parent
  delete next.repeat
  return next
}
```

On the deletes themselves, I'd keep them: it isn't only about callers adjusting their input. `stripChildOnlyFields` runs over the queue-level defaults too, not just the per-call options:

```ts
const defaultOptions = stripChildOnlyFields((rawDefaults ?? {}) as JobsOptions)
const callerOptions = stripChildOnlyFields((options ?? {}) as JobsOptions)
```

`jobOptions` on `QueueConfiguration` is the full `JobOptionsType`, so a queue config can legitimately carry `deduplication` (and on v5 `repeat`/`debounce`) for `QueueManager.schedule`. When a flow child is built from that same config, those have to be dropped or BullMQ rejects the child. That's the library reusing its own config, not a caller passing something wrong.

There was no test covering it, so I added one: a child on `dedup_queue`, which has a queue-level `deduplication` default, must come back without it. Confirmed it fails if the `deduplication` delete is removed.

`debounce`/`repeat` are the weaker half of the list, agreed. On v6 they're unreachable and the deletes are no-ops; they only still matter for a v5 config. Happy to drop those two lines if you'd rather keep it to what v6 can express.

Rechecked on both lines: `tsc --noEmit` clean and full suite green against bullmq 6.2.0 + ioredis 6.0.0 and bullmq 5.81.3 + ioredis 5.11.1.

CarlosGamero (approved):

PR #1215, opened 2026-08-25
feat(backend-http-client): add transport-error detection helpers

Drodevbar:
## Changes

- Add `isTransportError` / `getTransportErrorCode` to `@lokalise/backend-http-client`: detect transport-level request failures without an HTTP response (undici `UND_ERR_*` timeouts and socket errors, `ECONNREFUSED`, `ECONNRESET`, `ETIMEDOUT`, `EPIPE`, `EAI_AGAIN`)
- Detection walks the `cause` chain (depth-capped, cycle-safe), so errors wrapped by `InternalRequestError` or `fetch` are recognized too

## Checklist

## AI Assistance Tracking


packages/app/backend-http-client/src/errors/transportError.ts
```diff
@@ -0,0 +1,39 @@
...
+])
+
+const MAX_CAUSE_DEPTH = 5
+
+/**
+ * Returns the transport-error code of the given error (or of an error in its
+ * `cause` chain, e.g. when wrapped by InternalRequestError or fetch), or
+ * undefined when the error is not a transport-level failure.
+ */
+export function getTransportErrorCode(error: unknown): string | undefined {
```
CarlosGamero: Should we maybe narrow down type to the possible it might return? Not sure if really usefult so please feel free to challenge

CarlosGamero (approved):

packages/app/backend-http-client/src/errors/transportError.ts
Drodevbar: makes sense, narrowed the type

jmatom:
nit (just a business / scope question for myself): should those transport errors cause a retry at the method `withRetry`? or are we delegating it into every project using the library? thanks @Drodevbar

packages/app/backend-http-client/src/errors/transportError.ts
```diff
@@ -0,0 +1,45 @@
...
+const MAX_CAUSE_DEPTH = 5
+
+/**
+ * Returns the transport-error code of the given error (or of an error in its
+ * `cause` chain, e.g. when wrapped by InternalRequestError or fetch), or
+ * undefined when the error is not a transport-level failure.
+ */
+export function getTransportErrorCode(error: unknown): TransportErrorCode | undefined {
+  let current: unknown = error
+  for (let depth = 0; depth <= MAX_CAUSE_DEPTH; depth++) {
```
jmatom: nit: MAX CAUSE DEPTH is 5 but loop is running 6, could we change the logic to `<` and/or adjust the constant?

packages/app/backend-http-client/src/index.ts
```diff
@@ -48,3 +48,5 @@ export {
   isResponseStatusError,
   ResponseStatusError,
 } from './errors/ResponseStatusError.ts'
+export type { TransportErrorCode } from './errors/transportError.ts'
+export { getTransportErrorCode, isTransportError } from './errors/transportError.ts'
```
kibertoad, line 52: can we also add documentation for the new things?

packages/app/backend-http-client/src/errors/transportError.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Transport-level failures without an HTTP response: undici timeouts/socket
+ * errors and node connection errors. Typically temporary, so callers usually
+ * want to retry them.
+ */
+const TRANSPORT_ERROR_CODES = [
```
kibertoad, line 6: I wonder if we should export them - we don't need them for direct client configuration?

packages/app/backend-http-client/src/errors/transportError.ts
Drodevbar: good catch, fixed

packages/app/backend-http-client/src/index.ts
Drodevbar: added

packages/app/backend-http-client/src/errors/transportError.ts
Drodevbar: I would say no. They are mostly for catch-site in consumers. The client already retries transport errors inline (`retryOnNetworkError`), but when retries are exhausted (or disabled, e.g. `retryOnTimeout: false`).

packages/app/backend-http-client/README.md
```diff
@@ -405,6 +405,25 @@ retry: {
...
+When retries are exhausted (or disabled), the transport error escapes to the caller. `isTransportError` and `getTransportErrorCode` let catch-sites tell "the server never responded" (typically temporary, worth retrying at a higher level) apart from other failures:
+
+```ts
+import { getTransportErrorCode, isTransportError, type TransportErrorCode } from '@lokalise/backend-http-client'
+
+try {
+  await sendGet(client, { path: '/items', requestLabel: 'List items' })
+} catch (err) {
+  if (isTransportError(err)) {
+    // undici timeout/socket error or node connection failure - safe to retry later
```
kibertoad: where does this retry logic live in this example?

packages/app/backend-http-client/src/errors/transportError.ts
jmatom, line 6: i would add `EHOSTUNREACH`

packages/app/backend-http-client/README.md
Drodevbar: In that example it's not there, but in a specific example of composed-connector - error will be classified -> returned to the caller (CTENG) and caller could retry it by sending another request.

Do you want me to adjust the example?

kibertoad: at least mention in the example that this is for usage somewhere down the line inside the app code itself

kibertoad (approved):

packages/app/backend-http-client/src/errors/transportError.ts
jmatom: Missed codes for the undice version we are using: 8.10.0
Add — 7 codes:
EHOSTUNREACH                    in undici's own retry set; the masking code I measured
ENETUNREACH                     in undici's set; no-IPv6-egress
ENETDOWN                        in undici's set
EHOSTDOWN                       in undici's set
EADDRNOTAVAIL                   port exhaustion; NOT in undici's set — stands on merit
ERR_SOCKET_CONNECTION_TIMEOUT   Node, internalConnectMultiple, no errno
UND_ERR_PRX_CONN                confirmed present in 8.10.0

Drodevbar:
@jmatom 
> nit (just a business / scope question for myself): should those transport errors cause a retry at the method withRetry? or are we delegating it into every project using the library? thanks @Drodevbar

I am initially gonna delegate it into every project using the library.

packages/app/backend-http-client/README.md
Drodevbar: done

packages/app/backend-http-client/src/errors/transportError.ts
Drodevbar: Extended the list, as you suggested

jmatom (approved):
approved in advance, maybe some extra codes can be added

PR #1217, opened 2026-08-25
Add errors package

CatchMe2:
## Summary

Introduces `@lokalise/errors`, base error classes with three guarantees:

- **Nominal typing.** Every error carries a string-literal `code`, so structurally identical error classes stop being assignable to each other. Constructors are private and the `from()` factories are the only creation path: a hand-written subclass omitting `readonly` on the `code` override would silently widen it to `string` and turn the check off.
- **Reliable `isInstance`.** Instances are stamped with `Symbol.for` symbols derived from their class-name inheritance path, and the static `isInstance()` guard checks the stamps instead of the prototype chain, so it works across realms and duplicated package copies. Plain `instanceof` also works via a `Symbol.hasInstance` override but is documented as deprecated. Class names are part of the identity; the README documents the naming contract (renames are breaking, names must survive minification, unnamed classes throw).
- **Type-safe error → API response flow.** `toPayload()` returns only the client-safe `{ message, code, errorCode, details? }` (never `stack`/`cause`/`name`) and always satisfies the definition's companion Zod `schema`. `mergeErrorSchemasByStatusCode()` turns definitions into `{ [statusCode]: payloadSchema }` for `@lokalise/api-contracts`, using `z.discriminatedUnion('code', ...)` when several errors share a status code.

## API

```ts
// Internal errors (never surfaced to clients): code only, details typed via a type argument
class DatabaseQueryError extends InternalError.from('DATABASE_QUERY_ERROR')<{ query: string }> {}

// Public errors: a definition (reused by API contracts) bound to a class
const projectNotFoundErrorDefinition = definePublicError({
  code: 'PROJECT_NOT_FOUND',
  type: ErrorType.NOT_FOUND,
  detailsSchema: z.object({ id: z.string() }),
})
class ProjectNotFoundError extends PublicError.from(projectNotFoundErrorDefinition) {}

const error = new ProjectNotFoundError({ message: 'Project abc not found', details: { id: 'abc' } })
reply.status(error.httpStatusCode).send(error.toPayload()) // 404, derived from ErrorType
```

To ease the `@lokalise/node-core` migration, every error also exposes a deprecated `errorCode` getter mirroring `code`; it appears in `toPayload()` and the companion schema (flagged deprecated in OpenAPI output) and will be removed in a future major.

## Supporting changes

* **CI:** registered `packages/app/errors` in the changed-package map.
* **Changeset:** added an intentionally package-less changeset; the first release will be published manually.
* **Package setup:** mirrors the other shared libraries, including Biome, Vitest with coverage, shared `tsconfig`, and Zod v4 as a peer dependency.

## Checklist

## AI Assistance Tracking


packages/app/errors/src/EnhancedError.ts
```diff
@@ -0,0 +1,147 @@
...
+ *   instanceof across realms and package copies.
+ *
+ * Since the symbols are derived from class names, every class in the chain
+ * must have one. The constructor throws when it encounters an unnamed class
+ * (e.g. a class expression returned from a factory). Name such classes
+ * explicitly with `Object.defineProperty(TheClass, 'name', { value: '...' })`.
+ * Avoid '.' in such custom names — it is the path delimiter, so a name like
+ * 'Foo.Bar' produces the same identity as a `Foo` → `Bar` chain.
+ */
+export abstract class EnhancedError<TDetails = undefined> extends Error {
```
kibertoad, line 136: personally I'm not a fan of supporting instanceof only for just this one group of errors, it's very magical and gotcha'y

can we also expose a type-guard and have it supported as a first-class option too?

CatchMe2: This follows the approach already used in node-core errors, so it's the established pattern for our error classes rather than something new to this package.

We could additionally expose a static `isInstance()` class method (like the AI SDK does: https://github.com/vercel/ai/blob/main/packages/ai/src/error/invalid-stream-part-error.ts), but it wouldn't work for subclasses created in services (I mean to check for subclass instanceof), each service would have to reimplement the same machinery for every custom error class, exactly like the AI SDK does per-class. That complicates creating custom internal errors quite a bit, so I'm not in favour of it. I'm also not sure we really need two ways of doing the same check.

packages/app/errors/README.md
```diff
@@ -0,0 +1,261 @@
...
+factory** rather than extending and overriding `code`/`type` manually — the
+factory bakes literal types in from the definition automatically, avoiding the
+footgun of accidentally omitting `readonly` on an override.
+
+Details are typed via an optional Zod schema, which also enables OpenAPI schema
+generation on the contract layer.
+
+```ts
+import { z } from 'zod/v4'
+import { PublicError, ErrorType, definePublicError } from '@lokalise/errors'
```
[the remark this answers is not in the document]
CatchMe2: fixed

packages/app/errors/README.md
```diff
@@ -0,0 +1,261 @@
...
+class. TypeScript narrows on literal types, so cross-error assignments become
+compile errors.
+
+## Error hierarchy
+
+```
+EnhancedError (abstract, not exported) — shared base: code, details, cause, cross-realm instanceof
+├── InternalError (abstract)  — non-public operational errors
+└── PublicError   (abstract)  — client-facing errors; has type + httpStatusCode
+```
```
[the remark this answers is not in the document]
CatchMe2: Fixed.

packages/app/errors/src/EnhancedError.ts
CatchMe2: The `isInstance` approach was added, and `instanceof` was deprecated.

packages/app/errors/src/PublicError.ts
```diff
@@ -0,0 +1,195 @@
...
+   * the definition declares a `detailsSchema`. Excludes non-public fields
+   * (`stack`, `cause`, `name`) and always satisfies the definition's `schema`.
+   */
+  toPayload(): PublicErrorPayload<T> {
+    return {
+      message: this.message,
+      code: this.code,
+      ...(this.details !== undefined && { details: this.details }),
+    } as PublicErrorPayload<T>
+  }
```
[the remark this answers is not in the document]
CatchMe2: It should be handled on the response serialization (fastify-zod-provider is executing parse on response object)

packages/app/errors/src/EnhancedError.ts
```diff
@@ -0,0 +1,171 @@
...
+   * Type guard for this error class — the preferred way to check errors.
+   *
+   * Inherited by every subclass and narrows to the subclass it is called on.
+   */
+  static isInstance<T>(
+    this: { prototype: T; [Symbol.hasInstance](value: unknown): boolean },
+    value: unknown,
+  ): value is T {
+    // biome-ignore lint/complexity/noThisInStatic: intentional to support subclasses
+    return value instanceof this
```
kibertoad: **`isInstance` throws when detached from its class.**

It is `value instanceof this`, so passing it as a callback breaks at runtime:

```ts
errors.filter(DatabaseQueryError.isInstance)
```

That typechecks cleanly (the `this` parameter is not enforced through `Array.prototype.filter`) and throws `TypeError: Right-hand side of 'instanceof' is not an object`, because `this` is `undefined` in a module.

The README presents `isInstance` as the supported guard and it has exactly the shape of a predicate, so this is an easy trap. Binding it in `from()`, or declaring it as a `const` arrow static, removes it.

kibertoad (commented):
Review of the new `@lokalise/errors` package. Verified by building with the repo's TS 7.0.2 setup, compiling a separate consumer project against the emitted `dist/*.d.ts`, running the 69 tests plus the 100% coverage thresholds, and runtime probes/microbenchmarks.

7 inline comments below, most to least important. Only the first one is a real bug; the rest are hardening, typing, docs, and perf.

Clean: the anonymous-class identity guard (correct for leaf and mid-chain unnamed classes), `cause` handling, `httpStatusByErrorType`, the status-code grouping logic and its type-level key preservation, declaration emit for the `from()` factories, the CI changed-package map entry, and the changeset format.

packages/app/errors/src/mergeErrorSchemasByStatusCode.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const result: Record<number, z.ZodType> = {}
+
+  for (const [statusCode, schemas] of schemasByStatusCode) {
+    const [firstSchema] = schemas
+    result[statusCode] =
+      schemas.length === 1 && firstSchema
+        ? firstSchema
+        : z.discriminatedUnion('code', schemas as [z.ZodObject, ...z.ZodObject[]])
```
kibertoad, line 120: **Duplicate `code` within a status code fails at first parse, not at merge time.**

zod v4 builds the discriminator map lazily, so two definitions sharing a status code and a `code` merge without complaint and the first `safeParse` throws `Duplicate discriminator value "X"` (reproduced). The documented uniqueness constraint then surfaces during request/response validation instead of at contract-definition time.

A duplicate-code check in the grouping loop would move the failure to startup.

packages/app/errors/src/PublicError.ts
```diff
@@ -0,0 +1,216 @@
...
+   *
+   * Also includes the deprecated `errorCode` alias of `code` for
+   * `@lokalise/node-core` compatibility; it will be dropped in a future major.
+   */
+  toPayload(): PublicErrorPayload<T> {
+    return {
+      message: this.message,
+      code: this.code,
+      errorCode: this.code,
+      ...(this.details !== undefined && { details: this.details }),
```
kibertoad: **`toPayload()` puts the live `details` object into the payload by reference.**

Every call returns the same object, shared with the error instance. A response-redaction hook or a caller adding a field to `payload.details` mutates the error that is later logged, and mutating the error mutates payloads already handed out.

`...(this.details !== undefined && { details: { ...this.details } })` decouples them.

packages/app/errors/src/PublicError.ts
```diff
@@ -0,0 +1,216 @@
...
+ * definition declares a `detailsSchema`.
+ */
+export type PublicErrorPayload<T extends PublicErrorDefinition> = {
+  message: string
+  code: T['code']
+  /** @deprecated Use {@link code} instead — node-core compatibility alias. */
+  errorCode: T['code']
+} & (T['detailsSchema'] extends z.ZodObject
+  ? { details: InferPublicErrorDetails<T> }
+  : { details?: never })
```
kibertoad: **The generic payload type resolves `details` to `never`.**

For the wide base type, `T['detailsSchema']` is `z.ZodObject | undefined`, which does not extend `z.ZodObject`, so `PublicErrorPayload<PublicErrorDefinition>` lands on the `{ details?: never }` branch.

The README's own generic handler hits this:

```ts
if (PublicError.isInstance(err)) reply.send(err.toPayload())
```

The payload types as never carrying details while the runtime payload does carry them, so reading `payload.details` types as `undefined`, and assigning the payload to a contract response type that requires `details` fails to compile.

packages/app/errors/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+Override it in a subclass only if you need custom matching logic; the default
+inherited guard covers the standard cases.
+
+### How `isInstance` stays reliable across realms and package copies
+
+On instantiation each error tags itself with `Symbol.for` symbols derived from
+its inheritance path (e.g.
+`'@lokalise/errors.EnhancedError.InternalError.TranslatorTimeoutError'`), and
```
kibertoad: **This identity path is not what the code stamps, and the consequence is undocumented.**

`InternalError.from` renames the bound class to `InternalError<CODE>` (`InternalError.ts:84`), so a live instance carries `'@lokalise/errors.EnhancedError.InternalError.InternalError<TRANSLATOR_TIMEOUT>.TranslatorTimeoutError'` (verified at runtime), not the path shown here.

That means the error **code string is part of the cross-realm identity**: changing a `code` breaks `isInstance` across realms and duplicated copies exactly like renaming the class. The "Class naming rules" section only warns about renames, so this should be called out there too.

packages/app/errors/src/EnhancedError.ts
```diff
@@ -0,0 +1,171 @@
...
+    super(options.message, options.cause !== undefined ? { cause: options.cause } : undefined)
+
+    // Set the error's name to the name of the class that was instantiated
+    this.name = new.target.name
+    // Cast needed because the conditional type is not narrowed by the compiler here.
+    this.details = options.details as TDetails
+
+    const prototypeNames = getConstructorNamesPostError(new.target)
+
+    // An unnamed class anywhere in the chain truncates the walk, which would
```
kibertoad: **The prototype-chain walk is redone on every construction and every check.**

The constructor walks the chain, reverses/slices it, and calls `Symbol.for` + `Object.defineProperty` once per ancestor for each instance; `Symbol.hasInstance` repeats the walk plus a global-registry lookup on every check. Measured ~4x the cost of `new Error()` (2.48s vs 0.62s for 200k instantiations) and ~0.75µs per `isInstance`.

Both results are pure functions of the constructor, so a module-level `WeakMap<Function, ...>` would memoize them. Worth doing if these errors are constructed on hot paths.

packages/app/errors/package.json
```diff
@@ -0,0 +1,45 @@
...
+        "README.md"
+    ],
+    "type": "module",
+    "main": "./dist/index.js",
+    "exports": {
+        ".": "./dist/index.js",
+        "./package.json": "./package.json"
+    },
+    "homepage": "https://github.com/lokalise/shared-ts-libs",
+    "repository": {
```
kibertoad, line 17: **`repository` is missing `"directory"`.**

Every other package in the monorepo sets `"directory": "packages/app/errors"`. Without it the npm page's repository link points at the monorepo root instead of this package.

packages/app/errors/src/PublicError.ts
CatchMe2: Even the suggested approach wouldn’t fully fix this issue, since nested details fields might also be objects, and those object references would remain.

We could use `structuredClone`, or, since we already have a Zod schema for details, use `z.parse()`. The latter would ensure that we don’t expose any additional properties in the payload and would also give us a deep clone (it was raised by CodeRabbit https://github.com/lokalise/shared-ts-libs/pull/1217#discussion_r3852653909). However, I assume a similar check would already happen during response serialization.

WDYT?

packages/app/errors/src/EnhancedError.ts
CatchMe2: fixed

packages/app/errors/src/mergeErrorSchemasByStatusCode.ts
CatchMe2: Good catch, improved.

packages/app/errors/src/PublicError.ts
CatchMe2: Good catch, fixed

packages/app/errors/package.json
CatchMe2: fixed

packages/app/errors/src/PublicError.ts
kibertoad: `structuredClone` sounds reasonable, it is also widely supported by browsers.

CatchMe2: Sure, I've added cloning details field with structuredClone

packages/app/errors/README.md
CatchMe2: Improved the docs

.changeset/add-errors-package.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+Add the @lokalise/errors package. Its first release is published manually, so this changeset intentionally lists no packages.
```
CarlosGamero: What is the purpose of an empty changeset? just curious

packages/app/errors/src/EnhancedError.ts
```diff
@@ -0,0 +1,191 @@
...
+   * Compatibility alias for {@link code}, matching the error shape of
+   * `@lokalise/node-core`.
+   *
+   * @deprecated Use {@link code} instead. This alias only exists to ease the
+   * migration from `@lokalise/node-core` errors and will be removed in a
+   * future major version.
+   */
+  get errorCode(): string {
+    return this.code
+  }
```
CarlosGamero, line 155: 🟡 `errorCode` (and `PublicError.httpStatusCode`) are prototype getters, invisible to `JSON.stringify`, object spread, and own-property log serializers — the main consumption paths of node-core migrators.

packages/app/errors/src/InternalError.ts
```diff
@@ -0,0 +1,92 @@
...
+   * `string`), so cross-error assignments stay compile errors — without
+   * relying on subclass authors remembering `override readonly`.
+   *
+   * The bound class is named `InternalError<CODE>`, making the code part of
+   * the cross-realm identity path (see {@link EnhancedError}). Changing a
+   * code therefore breaks `isInstance` across realms and package copies,
+   * exactly like renaming a class.
+   */
+  static from<const TCode extends string>(code: TCode) {
+    class BoundInternalError<
```
CarlosGamero, line 131: I’m not a big fan of defining classes inline, as I find it makes the code harder to read. Because of TCode, I’m not sure if it’s possible to extract it into a private class within the file, but just mentioning it.

packages/app/errors/src/mergeErrorSchemasByStatusCode.ts
```diff
@@ -0,0 +1,103 @@
...
+ * keys, each mapping to a schema whose output is the union of the payloads of
+ * the definitions sharing that status code.
+ */
+export type MergedErrorSchemasByStatusCode<
+  TDefinitions extends readonly PublicErrorDefinitionWithSchema[],
+> = {
+  [TStatusCode in StatusCodeOf<TDefinitions[number]>]: z.ZodType<
+    z.output<DefinitionsForStatusCode<TDefinitions[number], TStatusCode>['schema']>
+  >
+}
```
CarlosGamero: CC finding, please check if it is a valid concern

MergedErrorSchemasByStatusCode declares every derivable status-code key as required, which is unsound when the definitions array is not a literal.
Verified with tsc: with `defs: PublicErrorDefinitionWithSchema[]` (e.g. definitions collected in a loop or registry), the result is typed as having all nine status codes non-optionally, so `responses[404].safeParse(body)` compiles — and throws `TypeError: Cannot read properties of undefined` at runtime when no not-found definition was passed. Make the mapped keys optional (Partial) for non-literal inputs.

packages/app/errors/src/EnhancedError.ts
CatchMe2: Good idea. Added the `WeakMap`

packages/app/errors/src/mergeErrorSchemasByStatusCode.ts
```diff
@@ -0,0 +1,103 @@
...
+ *   // ...
+ *   responsesByStatusCode: {
+ *     200: projectSchema,
+ *     ...mergeErrorSchemasByStatusCode([
+ *       projectNotFoundErrorDefinition,
+ *       projectNameAlreadyExistsErrorDefinition,
+ *     ]),
+ *     // → { 404: <not-found schema>, 409: <conflict schema> }
+ *   },
+ * })
```
CarlosGamero, line 90: This means we’ll define public errors in our API schema packages. I actually love the idea and I am in favour of it, but just wanted to highlight it in case anyone has any concerns.
cc: @kibertoad

.changeset/add-errors-package.md
CatchMe2: Ohh, I thought we require every change to have a changeset (we have such CI check in Polyglot). More info here: https://changesets.dev/guide/automating#blocking

I guess we can keep this file, it will be removed on merge automatically anyway.

packages/app/errors/src/mergeErrorSchemasByStatusCode.ts
CatchMe2: This is actually what triggered me to implement this package. In Polyglot, we had to duplicate public errors, define them in the contract and then implement the corresponding error class in the service.

With this package, you can define the public error in the contract (which becomes the source of truth) and then set up the service side public error class directly from that contract definition.

This gives us end-to-end type safety and ensures that errors are fully documented in OpenAPI, without having to maintain the same error definition in two places.

packages/app/errors/src/InternalError.ts
CatchMe2: Unfortunately, this is the only way to enforce literal types on `code`. Assuming that devs will always use `readonly` modifier on the `code` is is too fragile, easy to forget.

Using as const makes the literal type explicit and guarantees that the error code remains type-safe.

packages/app/errors/src/EnhancedError.ts
CatchMe2: This is true, but IMO we don't need backward compatibility for logs. I think we only need to preserve it for service logic that may rely on a specific field (e.g. error.errorCode) or when returning errors in HTTP responses, which is handled by the toPayload method.
I'd keep it this way. cc @kibertoad

packages/app/errors/src/PublicError.ts
```diff
@@ -0,0 +1,228 @@
...
+   *
+   * Also includes the deprecated `errorCode` alias of `code` for
+   * `@lokalise/node-core` compatibility; it will be dropped in a future major.
+   */
+  toPayload(): PublicErrorPayload<T> {
+    return {
+      message: this.message,
+      code: this.code,
+      errorCode: this.code,
+      ...(this.details !== undefined && { details: structuredClone(this.details) }),
```
CarlosGamero, line 138: 🟡 details is not validated against the schema, so the documented guarantee that the payload 'always satisfies the definition's schema' is false for schemas where zod input and output are different, right? Maybe we should use zod parse on error building phase

.changeset/add-errors-package.md
CarlosGamero: I think it is not required here, we might want to skip release (eg: dev dep bump) but it does not hurt, all good if we want to keep it

packages/app/errors/src/EnhancedError.ts
CarlosGamero: Mmm, okay, logs are not that problematic, all good from side :D

packages/app/errors/src/InternalError.ts
CarlosGamero: Okay, all good then. The type-safe is way more important tha my concern so I agree with the approach

packages/app/errors/src/PublicError.ts
kibertoad: do we really want to throw errors on error details mismatch?

CarlosGamero: It’s a good question. However, if we don’t run `parse` (or `safeParse`), our API docs might not be accurate (for example, a property could accept a number as input but return a string as output).

Maybe we could use `safeParse` and, in case of a mismatch, ignore the error and fall back to the input details.

packages/app/errors/src/PublicError.ts
```diff
@@ -0,0 +1,228 @@
...
+ * For the wide {@link PublicErrorDefinition} type (where `detailsSchema` may or
+ * may not be present), details resolve to `Record<string, unknown> | undefined`
+ * so generic error handlers can still inspect them.
+ */
+export type InferPublicErrorDetails<TDef extends PublicErrorDefinition> =
+  TDef['detailsSchema'] extends z.ZodObject
+    ? z.infer<TDef['detailsSchema']>
+    : [Extract<TDef['detailsSchema'], z.ZodObject>] extends [never]
+      ? undefined
+      : z.infer<Extract<TDef['detailsSchema'], z.ZodObject>> | undefined
```
CarlosGamero: 🟢 Not blocking, feel free to ignore. I noticed this three-branch conditional (with the `[Extract<...>] extends [never]` trick) appears three times: here, in `PublicErrorPayload`, and in the local `Schema` type inside `definePublicError` (where the base fields are also spelled out in both branches). When `errorCode` gets dropped in the planned major, or details semantics change, the three copies need to move in lockstep, and missing one compiles fine while quietly desynchronizing constructor options, payload type, and companion schema.

What do you think about keeping the schema extraction in a single place? For example:

```ts
export type InferPublicErrorDetails<TDef extends PublicErrorDefinition> =
  TDef['detailsSchema'] extends infer TSchema
    ? TSchema extends z.ZodObject
      ? z.infer<TSchema>
      : undefined
    : never
```

The distribution over `z.ZodObject | undefined` covers the wide-definition case naturally. Then `PublicErrorPayload` can derive its `details` slot from the resolved type instead of re-extracting the schema:

```ts
type PayloadDetailsField<TDetails> = [TDetails] extends [undefined]
  ? { details?: never }
  : undefined extends TDetails
    ? { details?: TDetails }
    : { details: TDetails }

export type PublicErrorPayload<T extends PublicErrorDefinition> = {
  message: string
  code: T['code']
  /** @deprecated Use {@link code} instead — node-core compatibility alias. */
  errorCode: T['code']
} & PayloadDetailsField<InferPublicErrorDetails<T>>
```

And the `Schema` type in `definePublicError` only needs to branch on the field that varies:

```ts
type BaseSchemaShape = {
  message: z.ZodString
  code: z.ZodLiteral<T['code']>
  errorCode: z.ZodLiteral<T['code']>
}
type Schema = z.ZodObject<
  T['detailsSchema'] extends z.ZodObject
    ? BaseSchemaShape & { details: T['detailsSchema'] }
    : BaseSchemaShape
>
```

I verified this compiles with TS 7.0.2 strict and keeps the spec's type assertions passing: literal `code`, required/absent/optional `details` (the tuple check `[TDetails] extends [undefined]` preserves `details?: never`, so the `@ts-expect-error` on `noDetailsPayload.details?.name` still fires), and the wide-handler `Record<string, unknown> | undefined` case. Runtime is untouched.

packages/app/errors/src/PublicError.ts
CatchMe2: We've discussed it partially here: https://github.com/lokalise/shared-ts-libs/pull/1217#discussion_r3871405143, but we didn't cover input/output schemas.

Thanks for raising it, this topic is not that simple and worth discussing (I guess a similar issue affects our `api-contracts` as well).

Imagine we have a Zod schema that uses `transform` and, transforms a `number` (input) into a `string` (output). If we use `zod.parse` on `details`, `toPayload()` would return details with the transformed `string` value.

The problem is that we also pass this Zod schema to the contract, which sets it as route schema, and uses it again during response serialization (via zod-type-provider). At that point, serialization would fail because we're passing the transformed string, while the schema expects a number as its input.

There seems to be a similar issue with api-contracts as well:
1. Fastify runs zod.parse before returning the response, so we're sending the schema's output type in the HTTP response body.
2. On the client side, we then apply zod.parse to the response body. However, at this point we're parsing the output payload using the schema's input type, which can fail for schemas that use transform.

WDYT guys cc @kibertoad

packages/app/errors/src/mergeErrorSchemasByStatusCode.ts
CatchMe2: Yeah, good point. I added logic to cover that, and added info in README to avoid non-literal inputs there.

> [!IMPORTANT]
> Always pass the definitions as an inline array literal, as in the example
> above. The type-level guarantees only hold when TypeScript sees exactly
> which definitions you passed. With a non-literal array (e.g. definitions
> collected dynamically into a `PublicErrorDefinitionWithSchema[]`), the type
> system cannot know which status codes are actually present at runtime:
> every status code key becomes optional and payload types degrade.

packages/app/errors/src/PublicError.ts
CarlosGamero: This is a good point. If we actually run parse when building the error, it will make it fail at the fastify-type-provider-zod level 😓. So OpenAPI will be correct, but we’ll always return a 500 if details doesn’t match the schema.

We could run parse on error building but still save details input  ... the only advantage of validating details is that we can detect these issues in lower-level tests (use cases, services, repos) without being forced to verify them in e2e tests. I’m not sure how much value we’d get from that, what do you think?.

What I do think is important, though, is to properly type details at this point. So, if I have MyError.details.prop, prop should be typed according to the Zod input type, ensuring we don’t get runtime issues.

CatchMe2: I'm not sure runtime validation is useful in lower-level tests. I'd leave it up to the test author to decide how they want to handle it. We still have type-level enforcement for `details`, so you'd have to explicitly bypass it with something like `details as any`.

In the end, we should have at least one test covering every possible endpoint result, including non-successful outcomes like 4xx errors. Those cases would be covered by the Zod parsing provided by `zod-type-provider`.

As for the `details` types, yeah, agreed. Let's use `z.input` for `details`, including the `details` type used by the `toPayload` method result.

CarlosGamero: yeah, I agree with you on that, let's avoid the validation in that point then :D

packages/app/errors/src/PublicError.ts
CatchMe2: Good idea. Improved

packages/app/errors/src/PublicError.ts
CatchMe2: Done

CarlosGamero (approved):

kibertoad (approved):

CarlosGamero (approved):

packages/app/errors/src/EnhancedError.ts
```diff
@@ -0,0 +1,221 @@
...
+ * 'Foo.Bar' produces the same identity as a `Foo` → `Bar` chain.
+ */
+export abstract class EnhancedError<TDetails = undefined> extends Error {
+  /**
+   * Stable, unique string identifier for this error class.
+   * Must be declared `readonly` in every subclass to enable TS narrowing.
+   */
+  abstract readonly code: string
+
+  readonly details: TDetails
```
kibertoad, line 143: `details` is assigned as an own **enumerable** property (here and at line 163), so any JSON-based sink republishes it. For an `InternalError` that is the internal data this class documents as never reaching clients:

```ts
class DbError extends InternalError.from('DB')<{ query: string; secret: string }> {}
const err = new DbError({ message: 'boom', details: { query: 'select 1', secret: 'hunter2' } })

JSON.stringify(err)
// {"details":{"query":"select 1","secret":"hunter2"},"name":"DbError","code":"DB"}
Object.keys(err)
// ['details', 'name', 'code']
```

`JSON.stringify(new Error('x'))` is `'{}'` for comparison. Any `reply.send(err)`, `{ ...err }`, or JSON-based log shipper publishes details verbatim, which is the leak `toPayload()` and the doc comment at `InternalError.ts:28-32` exist to prevent. (`this.name = new.target.name` on line 161 is what puts `name` on that list too; plain `Error` keeps it non-enumerable on the prototype.)

### Careful with the obvious fix

Making the property non-enumerable breaks logging. `@lokalise/node-core`'s `resolveGlobalErrorLogObject` serializes through pino's `stdSerializers.err`, and that serializer copies properties with `for (const key in err)` (`pino-std-serializers/lib/err.js:29`). Enumerability is precisely what lands `details` in the logs today, so flipping the attribute empties them silently.

Measured on four shapes of the same class:

| vector | own enumerable (current) | own non-enumerable | getter on prototype |
| --- | --- | --- | --- |
| `JSON.stringify(err)` | leaks details | clean | clean |
| `{ ...err }`, `Object.assign` | leaks details | clean | clean |
| `Object.keys(err)` | `['name','details']` | `['name']` | `['name']` |
| `stdSerializers.err(err)` | details logged | **details lost** | details logged |
| `err.details` | works | works | works |

What separates the two behaviors is where the property lives, not only its attributes: `JSON.stringify`, object spread and `Object.keys` read own properties only, while `for...in` walks the prototype chain.

### Suggested shape

```ts
export abstract class EnhancedError<TDetails = undefined> extends Error {
  #details: TDetails

  /**
   * Error details. Backed by a private field and exposed through a prototype
   * accessor, so `JSON.stringify(error)`, `{ ...error }` and `Object.keys`
   * (own properties only) cannot carry details into a response body, while
   * log serializers that walk the prototype chain with `for..in` still
   * record them.
   */
  get details(): TDetails {
    return this.#details
  }

  static {
    // Class accessors are non-enumerable by default, which would hide details
    // from log serializers. A partial descriptor flips only that attribute.
    Object.defineProperty(EnhancedError.prototype, 'details', { enumerable: true })
  }

  constructor(options: EnhancedErrorOptions<TDetails>) {
    // ...
    this.#details = options.details as TDetails
  }
}
```

Verified against the real pino serializer: `JSON.stringify(err)` becomes `{"name":"DbError","code":"DB"}` while `stdSerializers.err(err).details` still holds both fields. The partial `Object.defineProperty` preserves the existing getter.

Two things to watch:

- Never reintroduce an own `details` slot. A bare field declaration on a subclass (what TS emits under `useDefineForClassFields` for `details: T` without `declare`) shadows the accessor with an own enumerable `undefined`: reads and logging both break. Nothing declares `details` outside this class today, so it is a note for later, not an edit.
- `Object.hasOwn(err, 'details')` flips to `false`. The identity-symbol checks use symbol keys, so they are unaffected.

A lighter option, if the storage change is unwelcome: a prototype `toJSON()` returning `{ name, code, message }`, optionally `this.toPayload()` on `PublicError`. Class methods are non-enumerable, so pino keeps seeing details and `JSON.stringify` gets the safe shape, but `{ ...err }` and `Object.keys` keep leaking. If you do override in `PublicError`, the base hook needs a loose return type (`Record<string, unknown>`), since `PublicErrorPayload<T>` has no `name`.

Either way, please pin both directions in a single test: `JSON.stringify(err)` carries no details, **and** the pino error serializer still reports them. Without that second assertion, a later "make details non-enumerable for safety" commit passes review and quietly empties the logs.

CatchMe2: I wouldn't change it. The existing behavior makes it backward-compatible with node-core errors.

Also, logging all error details by default is preferable here. The logs are internal anyway, and having the details available can make debugging significantly easier.

As for InternalErrors, they aren't meant to be returned in the response in the first place. For those, we return a hardcoded code/message instead: https://github.com/lokalise/fastify-extras/blob/main/lib/errors/errorHandler.ts#L140

So I don't think we need to optimize the error shape to prevent details from being serialized, especially since doing so could also change the existing logging behavior.

PR #1230, opened 2026-08-27
fix(fastify-api-contracts): keep committed SSE stream open when handler errors

CatchMe2:
## Summary

When an error escaped an SSE-capable route handler after `sse.start()`, `@fastify/sse` closed the committed stream before rethrowing (its handler wrapper calls `context.close()` for non-keepAlive sessions on the error path). By the time the error reached the global error handler, the connection was already gone, so an SSE-aware error handler (fastify-extras) could not serialize the error into a terminal SSE event.

## Changes

- `buildApiSSEContext` now exposes a `preserveStreamOnError()` probe: once the stream is started, it flags the session keep-alive (making the plugin's error-path close a no-op) and marks the eventual close as server-initiated for `onClose`. It is a no-op before `sse.start()`.
- `handleApiRoute` wraps handler execution (both the imperative `sse.start()` path and the async-iterable body path) in a try/catch that calls the probe and rethrows, so the error propagates to the global error handler with the stream still connected. The error handler is then responsible for sending a terminal error event and closing the stream.
- Errors thrown before the stream starts keep the regular HTTP error response path.

## Tests

- Error after `sse.start()` reaches a global error handler with the stream still connected; the error event goes out on the open stream.
- Same for an async-iterable body that throws mid-stream.
- `onClose` reports `'server'` as the initiator when the error handler closes the errored stream.
- A throw before `sse.start()` still returns a regular JSON 500.

## AI Assistance Tracking

### DIFF
--- a/packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
+++ b/packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
@@ -1090,8 +1090,8 @@ describe('buildFastifyApiRoute — runtime', () => {
 
   it('keeps the stream open for the global error handler when the handler throws after sse.start()', async () => {
     app = await buildApp()
-    // Mimics an SSE-aware global error handler (fastify-extras): the stream is still
-    // connected when the error arrives, so it goes out as a terminal SSE event.
+    // Mimics an SSE-aware global error handler: the stream is still connected when the
+    // error arrives, so it goes out as a terminal SSE event.
     app.setErrorHandler<Error>(async (error, _request, reply) => {
       expect(reply.sse.isConnected).toBe(true)
       await reply.sse.send({ event: 'error', data: { message: error.message } })
--- a/packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
+++ b/packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
@@ -1088,6 +1088,117 @@ describe('buildFastifyApiRoute — runtime', () => {
     await vi.waitFor(() => expect(closeInitiator).toBe('server'))
   })
 
+  it('keeps the stream open for the global error handler when the handler throws after sse.start()', async () => {
+    app = await buildApp()
+    // Mimics an SSE-aware global error handler (fastify-extras): the stream is still
+    // connected when the error arrives, so it goes out as a terminal SSE event.
+    app.setErrorHandler<Error>(async (error, _request, reply) => {
+      expect(reply.sse.isConnected).toBe(true)
+      await reply.sse.send({ event: 'error', data: { message: error.message } })
+      reply.sse.close()
+    })
+    app.route(
+      buildFastifyApiRoute(sseOnlyContract, async (_request, _reply, { sse }) => {
+        const session = sse.start('autoClose')
+        await session.send('update', { value: 1 })
+        throw new Error('stream blew up')
+      }),
+    )
+    await app.ready()
+
+    const response = await app.inject({
+      method: 'GET',
+      url: '/stream',
+      headers: { accept: 'text/event-stream' },
+    })
+    expect(response.statusCode).toBe(200)
+    expect(response.body).toContain('event: update')
+    expect(response.body).toContain('event: error')
+    expect(response.body).toContain('stream blew up')
+  })
+
+  it('keeps the stream open for the global error handler when an async-iterable body throws mid-stream', async () => {
+    app = await buildApp()
+    app.setErrorHandler<Error>(async (error, _request, reply) => {
+      expect(reply.sse.isConnected).toBe(true)
+      await reply.sse.send({ event: 'error', data: { message: error.message } })
+      reply.sse.close()
+    })
+    app.route(
+      buildFastifyApiRoute(sseOnlyContract, (_request, _reply) => ({
+        status: 200,
+        // biome-ignore lint/suspicious/useAwait: async is required to satisfy AsyncIterable
+        body: (async function* (): AsyncGenerator<SSEStreamMessage<typeof sseEventsSchema>> {
+          yield { event: 'update', data: { value: 1 } }
+          throw new Error('source failed')
+        })(),
+      })),
+    )
+    await app.ready()
+
+    const response = await app.inject({
+      method: 'GET',
+      url: '/stream',
+      headers: { accept: 'text/event-stream' },
+    })
+    expect(response.statusCode).toBe(200)
+    expect(response.body).toContain('event: update')
+    expect(response.body).toContain('event: error')
+    expect(response.body).toContain('source failed')
+  })
+
+  it("reports 'server' as the close initiator when the global error handler closes an errored stream", async () => {
+    let closeInitiator: string | undefined
+    app = await buildApp()
+    app.setErrorHandler<Error>(async (error, _request, reply) => {
+      await reply.sse.send({ event: 'error', data: { message: error.message } })
+      reply.sse.close()
+    })
+    app.route(
+      buildFastifyApiRoute(
+        sseOnlyContract,
+        async (_request, _reply, { sse }) => {
+          const session = sse.start('autoClose')
+          await session.send('update', { value: 1 })
+          throw new Error('stream blew up')
+        },
+        {
+          onClose: (_session, initiator) => {
+            closeInitiator = initiator
+          },
+        },
+      ),
+    )
+    await app.ready()
+
+    const response = await app.inject({
+      method: 'GET',
+      url: '/stream',
+      headers: { accept: 'text/event-stream' },
+    })
+    expect(response.statusCode).toBe(200)
+    await vi.waitFor(() => expect(closeInitiator).toBe('server'))
+  })
+
+  it('returns a regular 500 when an SSE-capable handler throws before sse.start()', async () => {
+    app = await buildApp()
+    app.route(
+      buildFastifyApiRoute(sseOnlyContract, () => {
+        throw new Error('failed before streaming')
+      }),
+    )
+    await app.ready()
+
+    const response = await app.inject({
+      method: 'GET',
+      url: '/stream',
+      headers: { accept: 'text/event-stream' },
+    })
+    expect(response.statusCode).toBe(500)
+    expect(response.headers['content-type']).toContain('application/json')
+    expect(response.json()).toMatchObject({ message: 'failed before streaming' })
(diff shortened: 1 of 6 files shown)


packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
```diff
@@ -1088,6 +1088,117 @@ describe('buildFastifyApiRoute — runtime', () => {
     await vi.waitFor(() => expect(closeInitiator).toBe('server'))
   })
 
+  it('keeps the stream open for the global error handler when the handler throws after sse.start()', async () => {
+    app = await buildApp()
+    // Mimics an SSE-aware global error handler (fastify-extras): the stream is still
+    // connected when the error arrives, so it goes out as a terminal SSE event.
```
CarlosGamero: Nit: I would avoid mentioning fastify-extras, it is tailored to our use case but this library should work independently to fasitfy-extras 
Edit: I am talking about the comment only, the test LGTM. Feel free to ignore

CarlosGamero (approved):

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
CatchMe2: Yeah, good point. I've removed `fastify-extra` mention.

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -279,5 +280,19 @@ export function buildApiSSEContext(
...
+    // keep-alive hands the still-open stream to the error handler, which owns serializing
+    // the error into a terminal SSE event and closing the connection (a server-initiated
+    // close). A no-op before `start()` — an uncommitted reply goes through the regular
+    // HTTP error path.
+    preserveStreamOnError: () => {
+      if (!started) {
+        return
+      }
+      closedByServer = true
+      reply.sse.keepAlive()
```
kibertoad: **High — with a non-SSE-aware error handler this never ends the response and crashes the process.**

Flagging keep-alive unconditionally means nothing closes the committed stream unless the app's global error handler happens to be SSE-aware. The plugin wrapper in `@fastify/sse@0.6.0` is:

```js
} catch (error) {
  if (!context.shouldKeepAlive) { context.close() }
  throw error
}
```

`context.close()` was the thing calling `reply.raw.end()`. Once `keepAlive()` suppresses it, `reply.sent` stays `false` (it's `kReplyHijacked || raw.writableEnded`, and the plugin only ever used `raw.write()`), so Fastify's error path continues into `onSendEnd` → `safeWriteHead` and throws `ERR_HTTP_HEADERS_SENT`. The retry inside `fallbackErrorHandler` (`fastify/lib/error-handler.js:40`) throws again outside any try/catch → unhandled rejection. The client never sees the response end, and the 30s heartbeat timer stays armed.

Reproduced against `fastify@5.12.1` + `@fastify/sse@0.6.0`, on real sockets and via `app.inject`, with both the default Fastify error handler and a plain `reply.status(500).send(...)` one. Removing the `keepAlive()` call makes the same request end cleanly (pre-PR behaviour).

Since this ships as a `patch`, consumers who haven't adopted the README's `reply.sse?.isConnected && reply.raw.headersSent` branch get this silently. Suggest either a safety net after the rethrow (if `reply.sse.isConnected && !reply.raw.writableEnded`, close it) or making the preserve behaviour opt-in, rather than assuming every error handler is SSE-aware.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
```diff
@@ -279,5 +280,19 @@ export function buildApiSSEContext(
...
+    // the error would reach the global error handler on an already-ended stream. Flagging
+    // keep-alive hands the still-open stream to the error handler, which owns serializing
+    // the error into a terminal SSE event and closing the connection (a server-initiated
+    // close). A no-op before `start()` — an uncommitted reply goes through the regular
+    // HTTP error path.
+    preserveStreamOnError: () => {
+      if (!started) {
+        return
+      }
+      closedByServer = true
```
kibertoad: **Medium — `closedByServer` is set before it's known who actually closes the stream.**

At this point the error has only been observed; whether the server ends the connection depends on what the global error handler does. For a `keepAlive` session that errors (the stream stayed open both before and after this PR), a later client-initiated disconnect is now reported to `onClose` as `'server'` where it was correctly `'client'` before. Same for `autoClose` when the error handler only logs.

Setting the flag in the close path — the way `session.close()` already does — keeps the initiator accurate, and initiator-keyed cleanup/metrics on the consumer side keep working.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
```diff
@@ -1088,6 +1088,117 @@ describe('buildFastifyApiRoute — runtime', () => {
     await vi.waitFor(() => expect(closeInitiator).toBe('server'))
   })
 
+  it('keeps the stream open for the global error handler when the handler throws after sse.start()', async () => {
```
kibertoad, line 1091: All four new tests install an SSE-aware error handler (`reply.sse.send(...)` + `reply.sse.close()`), which is exactly the configuration that masks the `ERR_HTTP_HEADERS_SENT` problem noted on `sseUtils.ts`. A case using the default Fastify error handler — or a plain `reply.status(500).send(...)` one — would fail today, and is the coverage that would keep the regression from coming back.

---
_Generated by [Claude Code](https://claude.ai/code)_

kibertoad (commented):
Reviewed the change against the real `fastify@5.12.1` and `@fastify/sse@0.6.0` sources and reproduced the main finding end-to-end.

The change itself is coherent: the try/catch placement is right, `preserveStreamOnError()` is correctly a no-op before `start()`, and the pre-`start()` JSON 500 path is preserved. The problem is what happens when the app's global error handler is *not* SSE-aware — see the inline comments on `sseUtils.ts`.

Findings, most severe first:

1. **High** — `sseUtils.ts:290-295`: flagging keep-alive on the error path means nothing closes the committed stream unless the error handler is SSE-aware; with the default handler Fastify raises an uncaught `ERR_HTTP_HEADERS_SENT` and the response never ends. Reproduced.
2. **Medium** — `sseUtils.ts:294`: `closedByServer = true` is set at error time, before it's known who ends the connection, so a later client disconnect is misreported to `onClose` as `'server'`.
3. **Medium** — `README.md:285` (not in this diff): the "Error handling" section still documents the pre-PR `autoClose` behaviour ("a handler error closes the stream without reaching the error handler") — the exact behaviour removed here — and confines the SSE-aware `setErrorHandler` pattern to `keepAlive`. That's the section consumers follow, so as written it walks them into finding 1.

The changeset is present and correctly scoped, though `patch` is arguable given this changes observable error-path behaviour for existing consumers.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: This is intentional. An SSE-aware global error handler is a documented requirement of this library, not an optional extra, see the Error handling section of the README: once a stream is committed, the only correct error signal is a terminal error event over the stream, and the documented setErrorHandler branches on `reply.sse?.isConnected && reply.raw.headersSent` to do exactly that.

Before this PR that requirement effectively only applied to keepAlive sessions. An autoClose error closed the stream silently and never reached the error handler, invisible to error reporting, and indistinguishable on the client from a dropped connection. The point of this change is to route those errors through the same documented path.

kibertoad: This requirement needs to be documented WAY louder than it currently is. even though I was intentionally looking for it now, quick skim did not revealed where that is mentioned.

we should also probably directly mention https://github.com/lokalise/fastify-extras/blob/main/lib/errors/errorHandler.ts as an example of SSE-aware global error handler

CatchMe2: I've added the `important` section to the README to make it more visually appearing.

This is how it is described now:

#### Error handling

Errors from contract routes go through the regular Fastify error handling chain (`fastify.setErrorHandler` or the default handler) — contract routes behave exactly like any other route. The exception is a **live SSE stream**: once the stream started, the status line and headers are on the wire, so a standard `reply.status().send()` can no longer work.

An error escaping the handler after the stream started — whether the session is `autoClose` or `keepAlive`, imperative (`sse.start()`) or declarative (`{ status, body: asyncIterable }`) — reaches the global error handler with the stream **still open**. Errors thrown before the stream starts take the regular HTTP error path.

> [!IMPORTANT]
> A Fastify instance serving SSE-capable routes **must** set an SSE-aware `setErrorHandler`. The only correct error signal on a committed stream is a terminal `error` event sent over the stream before closing it; an error handler that unconditionally calls `reply.status().send()` runs into already-sent headers, leaving the stream open and the error unreported to the client. The [`errorHandler` in `fastify-extras`](https://github.com/lokalise/fastify-extras/blob/main/lib/errors/errorHandler.ts) is a ready-made SSE-aware implementation.

An SSE-aware error handler branches on the stream state:

```ts
app.setErrorHandler(async (error, request, reply) => {
    const { statusCode, payload } = errorObjectResolver(error)
    errorReporter.report({ error, request })

    // The stream is live: headers are committed, only a terminal event can be sent.
    // (`isConnected` alone is not enough — @fastify/sse sets it before the handler runs.)
    if (reply.sse?.isConnected && reply.raw.headersSent) {
        await reply.sse.send({ event: 'error', data: payload })
        reply.sse.close()
        return
    }

    return reply.status(statusCode).send(payload)
})
```

packages/app/fastify-api-contracts/src/api-contracts/sseUtils.ts
CatchMe2: Good catch, improved

kibertoad (approved):

packages/app/fastify-api-contracts/src/api-contracts/buildFastifyApiRoute.spec.ts
CatchMe2: It is covered by improved README about the need for SSE aware global error handler.

PR #1233, opened 2026-08-31
feat: EXP-2307 Translation quality checks

CarlosGamero:
## Changes

`@lokalise/translation-quality-checks 1.0.0`

## Checklist

## AI Assistance Tracking


packages/app/translation-quality-checks/src/checks/mismatch/doubleWhitespaceMismatchCheck.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Mismatch check: reports the double-whitespace runs on which the text and the reference text
+ * disagree. Runs are paired as a multiset (position-independent, duplicates and run lengths
+ * counted, since positions naturally shift between languages): unpaired reference runs are
+ * reported as `missing`, unpaired text runs as `added`. Content inside NTC regions is not
+ * checked.
+ */
+export const doubleWhitespaceMismatchCheck: MismatchCheck = (
+  text: string,
+  compareWith: string,
+): DoubleWhitespaceMismatchIssue | undefined => {
```
kjamrog: 🟢 I'm wondering if `null` would be more explicit than `undefined` here. It ensures we always return value explicitly and not having `undefined` by mistake. We could also use `Either` to be even more verbose

packages/app/translation-quality-checks/README.md
```diff
@@ -1,3 +1,74 @@
-# @lokalise/translation-quality-checks
+# Translation quality checks
 
-> 🚧 **Work in progress.**
+Deterministic, code-based quality checks for translations — the rule-based counterpart to AI
+scoring. Pure and dependency-light: same input, same issues, no I/O — runs in the browser, in a
+backend, or inside Polyglot with identical results.
```
kjamrog: 🟢 This library will be public, while Polyglot is internal service - maybe let's not mention it here? At the end it's also a backend, so we can just say it works on both BE and client side

kjamrog (approved):

packages/app/translation-quality-checks/README.md
CarlosGamero: Oh right, CC was too specific 😮‍💨, thank you!

packages/app/translation-quality-checks/src/checks/mismatch/doubleWhitespaceMismatchCheck.ts
CarlosGamero: I would avoid using Either, as it would either mean duplicating the type here or adding a dependency on another package, which I’d prefer to avoid for an internal implementation detail since this is not part of the public API.

Regarding null vs. undefined, null seems to fit better here since it represents an explicit domain value.

Already adjusted

packages/app/translation-quality-checks/src/detectQualityIssues.ts
```diff
@@ -0,0 +1,89 @@
...
+    : Object.values(QualityIssueErrorEnum)
+  const checksToExclude = new Set<QualityIssueError>(options?.checksToExclude ?? [])
+
+  if (options && 'skipSingleTextChecks' in options && options.skipSingleTextChecks) {
+    for (const error of Object.keys(singleTextChecksByError)) {
+      checksToExclude.add(error as SingleTextQualityIssue['error'])
+    }
+  }
+
+  return checksToInclude.filter((check) => !checksToExclude.has(check))
```
diogomiguel: **suggestion,question** 
We are allowing duplicate codes in `checksToInclude` (a `new Set` before filtering would resolve it). Additionally, a custom checksToInclude (`options.checksToInclude`) the result order follows the caller's array rather than the enum order the default path uses. Is this intentional?

diogomiguel (approved):

packages/app/translation-quality-checks/src/detectQualityIssues.ts
CarlosGamero: Dedupe fixed with a Set. Order is intentionally unspecified — callers shouldn't rely on it, and the tests now assert order-agnostically.

packages/app/translation-quality-checks/src/checks/singleText/doubleWhitespaceCheck.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type DoubleWhitespaceIssue = QualityIssueShape<{
+  error: 'DOUBLE_WHITESPACE'
+  details: {
+    /** Every double-whitespace run found in the text, e.g. `['  ', ' \t']`. */
+    whitespaces: string[]
+  }
+}>
+
+const doubleWhitespaceRegexp = /\s{2,}/g
```
kibertoad: **Finding 1 of 15.** `/\s{2,}/g` treats line breaks as whitespace, so every CRLF newline and every blank line in a multi-line translation is reported as a `DOUBLE_WHITESPACE` issue.

**Failure scenario**

`\s` matches `\r` and `\n`, so:

```ts
doubleWhitespaceCheck('Line one\r\nLine two')
// { error: 'DOUBLE_WHITESPACE', details: { whitespaces: ['\r\n'] } }
doubleWhitespaceCheck('a\n\nb')  // whitespaces: ['\n\n']  (paragraph break)
doubleWhitespaceCheck('a\n  b')  // whitespaces: ['\n  ']  (indented continuation)
```

Any Windows-authored or multi-paragraph translation gets flagged on every line ending, and the reported `whitespaces` entry is a newline rather than a double space, which is meaningless in a translator-facing UI.

**Suggested fix**

The intent is horizontal whitespace, so exclude line terminators:

```ts
const doubleWhitespaceRegexp = /[^\S\r\n]{2,}/g
```

Use `/ {2,}/g` instead if only literal spaces should count.

kibertoad (commented):
## Code review: 15 findings

Recall-oriented pass over the whole diff. Findings are ordered most severe first and posted as
inline comments, each numbered so the ordering survives GitHub grouping them by file.

1. `doubleWhitespaceCheck.ts:12`: `/\s{2,}/g` flags CRLF newlines and blank lines as double whitespace.
2. `doubleWhitespaceMismatchCheck.ts:15`: the same pattern makes a CRLF source and an LF target a spurious mismatch.
3. `detectQualityIssues.ts:73`: `checksToInclude: []` runs every check instead of none.
4. `detectQualityIssues.ts:54`: options are silently dropped when the reference argument is `undefined`.
5. `utils.ts:68`: `ByError` does not tie a registry key to the check that reports that error code.
6. `index.ts:1`: `SingleTextQualityIssue` and `MismatchQualityIssue` are in the public signature but not exported.
7. `detectQualityIssues.ts:27`: `DetectQualityIssuesOptions` is exported but assignable to neither overload.
8. `nonTranslatableTagsMismatchCheck.ts:17`: `slice(1, -1)` hardcodes the NTC marker length.
9. `README.md:3`: 20 em-dashes were added, which the user-level CLAUDE.md bans (exact hits listed inline).
10. `doubleWhitespaceMismatchCheck.ts:17`: the double-whitespace extraction is duplicated across two check files.
11. `detectQualityIssues.ts:58`: the dispatch loop does a guaranteed-miss lookup per code and re-splits `text`.
12. `doubleWhitespaceCheck.ts:23`: `keepHtml: true` flags whitespace inside HTML attributes.
13. `doubleWhitespaceCheck.ts:16`: a whitespace-only NTC region hides a real run.
14. `detectQualityIssues.spec.ts:17`: the "never runs the mismatch checks" test is vacuous.
15. `package.json:29`: `--typecheck` was added to `test` only, and `types.spec.ts` is not a typecheck test file.

Items 1 through 8 are correctness or API-contract defects; 9 is a conventions violation; 10 through
13 are cleanup; 14 and 15 concern how the new checks are enforced. Nice work on the multiset diff and
on the exhaustiveness `satisfies` around `QualityIssueErrorEnum`, both of which read well.

packages/app/translation-quality-checks/src/checks/mismatch/doubleWhitespaceMismatchCheck.ts
```diff
@@ -0,0 +1,44 @@
...
+  error: 'DOUBLE_WHITESPACE_MISMATCH'
+  details: {
+    /** Double-whitespace runs present in the reference text but missing from the text. */
+    missing: string[]
+    /** Double-whitespace runs present in the text but absent from the reference text. */
+    added: string[]
+  }
+}>
+
+const doubleWhitespaceRegexp = /\s{2,}/g
```
kibertoad: **Finding 2 of 15.** The mismatch check carries the same newline-matching `/\s{2,}/g`, so a source and a target that differ only in line-ending style produce a spurious `DOUBLE_WHITESPACE_MISMATCH`.

**Failure scenario**

Source imported from a Windows-authored file, target typed in the web editor:

```ts
doubleWhitespaceMismatchCheck('Hola\nmundo', 'Hello\r\nworld')
// { error: 'DOUBLE_WHITESPACE_MISMATCH', details: { missing: ['\r\n'], added: [] } }
```

The translator is told they dropped a double whitespace that never existed. Same for a paragraph break: `'a\n\nb'` against `'x\ny'` reports a mismatch for what is a blank line, not double whitespace.

**Suggested fix**

Same pattern change as in `doubleWhitespaceCheck.ts`, which is also why the shared helper suggested further down matters:

```ts
const doubleWhitespaceRegexp = /[^\S\r\n]{2,}/g
```

packages/app/translation-quality-checks/src/detectQualityIssues.ts
```diff
@@ -0,0 +1,91 @@
...
+    const mismatchIssue = mismatchChecks[check]?.(text, compareWith)
+    if (mismatchIssue) issues.push(mismatchIssue)
+  }
+
+  return issues
+}
+
+const resolveChecksToRun = (options?: DetectQualityIssuesOptions): QualityIssueError[] => {
+  const checksToInclude = new Set(
+    options?.checksToInclude?.length
```
kibertoad: **Finding 3 of 15.** `options?.checksToInclude?.length` makes an explicitly empty `checksToInclude: []` fall back to running every check, the exact opposite of what the caller asked for.

**Failure scenario**

A caller derives the list from user settings:

```ts
detectQualityIssues(text, ref, { checksToInclude: enabledChecks })
```

When the user has disabled everything, `enabledChecks` is `[]`, the falsy-length test picks `Object.values(QualityIssueErrorEnum)`, and all seven checks run and report issues. The README documents `checksToInclude` as "checks to run; defaults to all of them", with no hint that `[]` means "all of them".

**Suggested fix**

Distinguish "not provided" from "provided empty":

```ts
const checksToInclude = new Set(
  options?.checksToInclude ?? Object.values(QualityIssueErrorEnum),
)
```

If the current behaviour is deliberate, state it in the README and in the JSDoc on `checksToInclude`, because `detectQualityIssues.spec.ts` is currently the only place it is written down.

packages/app/translation-quality-checks/src/detectQualityIssues.ts
```diff
@@ -0,0 +1,91 @@
...
+  compareWith: string,
+  options?: PairDetectQualityIssuesOptions,
+): QualityIssue[]
+export function detectQualityIssues(
+  text: string,
+  second?: string | DetectQualityIssuesOptions,
+  third?: DetectQualityIssuesOptions,
+): QualityIssue[] {
+  const compareWith = typeof second === 'string' ? second : undefined
+  const checksToRun = resolveChecksToRun(typeof second === 'string' ? third : second)
```
kibertoad: **Finding 4 of 15.** When the second argument is `undefined`, the third argument (the options object) is silently discarded, so every check runs instead of the requested subset.

**Failure scenario**

```ts
detectQualityIssues(text, maybeSource, { checksToExclude: ['DOUBLE_WHITESPACE'] })
```

with `maybeSource === undefined` at runtime (a nullable reference field, or any JavaScript consumer): `typeof second === 'string'` is false, so `resolveChecksToRun(second)` receives `undefined` and `third` is never read. The exclusion is dropped and all single-text checks run.

**Suggested fix**

Fall through to `third` when `second` is not an options object:

```ts
const compareWith = typeof second === 'string' ? second : undefined
const options = typeof second === 'string' ? third : (second ?? third)
const checksToRun = resolveChecksToRun(options)
```

packages/app/translation-quality-checks/src/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Contract of a single-text check.
+ */
+export type SingleTextCheck = (text: string) => SingleTextQualityIssue | null
+
+/**
+ * Contract of a mismatch check.
+ */
+export type MismatchCheck = (text: string, compareWith: string) => MismatchQualityIssue | null
+
+type ByError<Issue extends QualityIssueBase, Value> = { [K in Issue['error']]: Value }
```
kibertoad: **Finding 5 of 15.** `ByError` constrains only the key set, not that a key maps to the check that actually reports that error code. Annotating each check with the wide `SingleTextCheck` / `MismatchCheck` type throws away the narrow return type each check declares.

**Failure scenario**

```ts
export const doubleWhitespaceCheck: SingleTextCheck = (
  text: string,
): DoubleWhitespaceIssue | null => { ... }
```

The declared type of the binding is `SingleTextCheck`, whose return type is the wide union, so the `: DoubleWhitespaceIssue | null` annotation is invisible to every reader of the registry. As a result this type-checks cleanly:

```ts
export const mismatchChecksByError = {
  LEADING_WHITESPACE_MISMATCH: leadingWhitespaceMismatchCheck,
  TRAILING_WHITESPACE_MISMATCH: trailingWhitespaceMismatchCheck,
  DOUBLE_WHITESPACE_MISMATCH: leadingWhitespaceMismatchCheck, // wrong check, no error
  NON_TRANSLATABLE_TAGS_MISMATCH: nonTranslatableTagsMismatchCheck,
} as const satisfies ByError<MismatchQualityIssue, MismatchCheck>
```

and `checksToExclude: ['DOUBLE_WHITESPACE_MISMATCH']` would then silently disable the leading-whitespace check while the double-whitespace one keeps firing. The whole include/exclude mechanism keys off these registries, so the key to error correspondence is load-bearing.

**Suggested fix**

Tie the value's return type to the key:

```ts
type ByError<Issue extends QualityIssueBase, Args extends unknown[]> = {
  [K in Issue['error']]: (...args: Args) => Extract<Issue, { error: K }> | null
}

// ... as const satisfies ByError<SingleTextQualityIssue, [text: string]>
// ... as const satisfies ByError<MismatchQualityIssue, [text: string, compareWith: string]>
```

and drop the `: SingleTextCheck` / `: MismatchCheck` annotation on each check so its narrow return type survives.

packages/app/translation-quality-checks/src/index.ts
```diff
@@ -1,2 +1,2 @@
-// TODO: EXP-2307 To be implemented
-export const INIT = 'HELLO WORLD'
+export * from './detectQualityIssues.ts'
```
kibertoad, line 1: **Finding 6 of 15.** `SingleTextQualityIssue` and `MismatchQualityIssue` appear in the public overload signatures but live in `utils.ts`, which is not re-exported here and is unreachable through the `exports` map.

**Failure scenario**

Overload 1 returns `SingleTextQualityIssue[]` (`detectQualityIssues.ts:37`). A consumer writing `function render(issues: SingleTextQualityIssue[])` cannot import that name: `index.ts` re-exports only `./detectQualityIssues.ts` and `./types.ts`, and `package.json` `exports` exposes only `.` and `./package.json`, so a deep import of `dist/utils.js` is blocked. The only workaround is to rebuild the union by hand:

```ts
type SingleTextQualityIssue = Extract<
  QualityIssue,
  { error: 'LEADING_WHITESPACE' | 'TRAILING_WHITESPACE' | 'DOUBLE_WHITESPACE' }
>
```

**Suggested fix**

Re-export both unions from `types.ts`, which already imports them:

```ts
export type { MismatchQualityIssue, SingleTextQualityIssue } from './utils.ts'
```

packages/app/translation-quality-checks/src/detectQualityIssues.ts
```diff
@@ -0,0 +1,91 @@
...
+type PairDetectQualityIssuesOptions = {
+  /** Checks to run; defaults to all. */
+  checksToInclude?: QualityIssueError[]
+  /** Checks to skip; wins over `checksToInclude` on overlap. */
+  checksToExclude?: QualityIssueError[]
+  /** Shortcut for excluding every single-text check: only the mismatch checks run. */
+  skipSingleTextChecks?: boolean
+}
+
+export type DetectQualityIssuesOptions =
```
kibertoad, line 27: **Finding 7 of 15.** `DetectQualityIssuesOptions` is exported and advertised in the README, but a value of that union type is assignable to neither overload's options parameter.

**Failure scenario**

```ts
const opts: DetectQualityIssuesOptions = { checksToExclude: ['DOUBLE_WHITESPACE'] }
detectQualityIssues(text, opts)       // error
detectQualityIssues(text, ref, opts)  // error
```

The `PairDetectQualityIssuesOptions` member has `checksToInclude?: QualityIssueError[]`, wider than the `SingleTextQualityIssue['error'][]` that overload 1 requires, and overload 2 needs a `string` in second position. The README's closing line ("The relevant types are exported: ... and `DetectQualityIssuesOptions`") points consumers at a type they cannot pass anywhere.

**Suggested fix**

Export the two member types, so a caller can annotate options for the call shape they are actually using:

```ts
export type SingleTextDetectQualityIssuesOptions = { ... }
export type PairDetectQualityIssuesOptions = { ... }
```

packages/app/translation-quality-checks/src/checks/mismatch/nonTranslatableTagsMismatchCheck.ts
```diff
@@ -0,0 +1,40 @@
...
+    /** Non-translatable tokens present in the reference text but missing from the text. */
+    missing: string[]
+    /** Non-translatable tokens present in the text but absent from the reference text. */
+    added: string[]
+  }
+}>
+
+// The NTC markers are single characters, so slicing them off yields the wrapped token.
+const nonTranslatableTokens = (text: string): string[] =>
+  extractNTCTagsWithContent(text).map((region) => region.slice(1, -1))
```
kibertoad: **Finding 8 of 15.** `region.slice(1, -1)` hardcodes the NTC marker length instead of deriving it from the exported `NON_TRANSLATABLE_START_TAG` / `NON_TRANSLATABLE_END_TAG` constants.

**Failure scenario**

The comment on line 15 states the assumption but nothing enforces it. `@lokalise/non-translatable-markup` ships the markers as exported constants precisely so callers do not hardcode them, and this file's own spec imports them. If they ever become multi-character sequences, every token silently keeps marker fragments, the `missing` / `added` diffs compare mangled tokens against clean ones and report a mismatch for identical texts, and no test in this package fails.

**Suggested fix**

```ts
import {
  extractNTCTagsWithContent,
  NON_TRANSLATABLE_END_TAG,
  NON_TRANSLATABLE_START_TAG,
} from '@lokalise/non-translatable-markup'

const nonTranslatableTokens = (text: string): string[] =>
  extractNTCTagsWithContent(text).map((region) =>
    region.slice(NON_TRANSLATABLE_START_TAG.length, -NON_TRANSLATABLE_END_TAG.length),
  )
```

Better still: have the NTM package export a token extractor built on its existing `nonTranslatableTagsRegexpG` capture group, so the markers never have to leave that package.

packages/app/translation-quality-checks/README.md
```diff
@@ -1,3 +1,74 @@
-# @lokalise/translation-quality-checks
+# Translation quality checks
 
-> 🚧 **Work in progress.**
+Deterministic, code-based quality checks for translations — the rule-based counterpart to AI
```
kibertoad: **Finding 9 of 15.** Twenty em-dashes (U+2014) were added across this README and three source doc comments, which the user-level `~/.claude/CLAUDE.md` bans outright.

**Rule**

> Never emit U+2014 (em dash) or U+2013 (en dash) as punctuation. Never.

scoped to "code comments, docs, READMEs, `CLAUDE.md`/`AGENTS.md`, commit messages, PR titles, PR descriptions, review comments, ...". The same rule also rules out swapping in a spaced hyphen, a double hyphen, or an ellipsis for the same pause.

**Exact hits**

- `packages/app/translation-quality-checks/README.md`: lines 3, 4, 11, 17, 44, 45, 46, 56, 61, 62, 63, 64, 65, 66, 67, 70
- `packages/app/non-translatable-markup/src/methods/extractNTCTagsWithContent.ts`: line 4 (two occurrences on that one line)
- `packages/app/translation-quality-checks/src/checks/mismatch/nonTranslatableTagsMismatchCheck.ts`: line 20
- `packages/app/translation-quality-checks/src/detectQualityIssues.ts`: line 40

**Suggested fix**

Replace each with the punctuation the sentence needs: a colon before an elaboration, a comma before a clause, parentheses around an aside, a semicolon between two independent clauses, or a full stop and a second sentence. Then verify:

```bash
git diff main...HEAD | grep -nP '^\+.*\xe2\x80[\x93\x94]'
```

packages/app/translation-quality-checks/src/checks/mismatch/doubleWhitespaceMismatchCheck.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const doubleWhitespaceRuns = (text: string): string[] =>
```
kibertoad: **Finding 10 of 15.** `doubleWhitespaceRuns` and the `doubleWhitespaceRegexp` above it duplicate the extraction inlined in `doubleWhitespaceCheck.ts`, so the pattern and the NTC splitting options have to be kept in sync by hand across two files.

**Failure scenario**

The same regex plus the same `extractTextBetweenTags(text, { keepHtml: true, preserveSpacing: true }).flatMap((piece) => piece.match(...) ?? [])` appears at `doubleWhitespaceCheck.ts:12` and `:22-25`, and again here at `:15` and `:17-20`. The newline bug reported above has to be fixed twice, and any future divergence (say `keepHtml` flipped in one place only) makes the single-text check and the mismatch check disagree about the same text, with nothing in the suite catching it.

**Suggested fix**

One helper next to `multisetDiff.ts`, called from both checks:

```ts
// src/checks/utils/doubleWhitespaceRuns.ts
const doubleWhitespaceRegexp = /[^\S\r\n]{2,}/g

export const doubleWhitespaceRuns = (text: string): string[] =>
  extractTextBetweenTags(text, { keepHtml: true, preserveSpacing: true }).flatMap(
    (piece) => piece.match(doubleWhitespaceRegexp) ?? [],
  )
```

packages/app/translation-quality-checks/src/detectQualityIssues.ts
```diff
@@ -0,0 +1,91 @@
...
+  text: string,
+  second?: string | DetectQualityIssuesOptions,
+  third?: DetectQualityIssuesOptions,
+): QualityIssue[] {
+  const compareWith = typeof second === 'string' ? second : undefined
+  const checksToRun = resolveChecksToRun(typeof second === 'string' ? third : second)
+
+  const issues: QualityIssue[] = []
+
+  for (const check of checksToRun) {
```
kibertoad, line 59: **Finding 11 of 15.** The dispatch loop probes both registries for every error code even though their key sets are disjoint, and it re-runs the NTC split of `text` once per double-whitespace check.

**Failure scenario**

For the seven-code default set the loop performs fourteen lookups, seven of which can never hit, and the shape of the code implies a code could live in both registries (it cannot: `ByError` partitions them by construction). Separately, a pair call runs `extractTextBetweenTags(text, ...)` twice on the same string, once inside `doubleWhitespaceCheck` and once inside `doubleWhitespaceMismatchCheck`, plus once on `compareWith`: three regex splits and three array allocations where two would do.

**Suggested fix**

A code is in exactly one registry, so short-circuit after the first hit. This keeps the current interleaved issue order that `detectQualityIssues.spec.ts` asserts:

```ts
for (const check of checksToRun) {
  const singleTextCheck = singleTextChecks[check]
  if (singleTextCheck) {
    const issue = singleTextCheck(text)
    if (issue) issues.push(issue)
    continue
  }

  if (compareWith === undefined) continue

  const issue = mismatchChecks[check]?.(text, compareWith)
  if (issue) issues.push(issue)
}
```

Then hoist the double-whitespace piece extraction (see the shared-helper comment) so `text` is split once per call rather than once per check.

packages/app/translation-quality-checks/src/checks/singleText/doubleWhitespaceCheck.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Single-text check: reports every run of two or more consecutive whitespace characters found in
+ * the given text. Content inside NTC regions is not checked: the text is split by NTC region and
+ * every piece is evaluated on its own, so a run can never span a region boundary.
+ */
+export const doubleWhitespaceCheck: SingleTextCheck = (
+  text: string,
+): DoubleWhitespaceIssue | null => {
+  const whitespaces = extractTextBetweenTags(text, {
+    keepHtml: true,
```
kibertoad: **Finding 12 of 15.** `keepHtml: true` leaves HTML tag markup in the checked pieces, so whitespace inside attributes and tag syntax is reported as a translator-facing double-whitespace issue.

**Failure scenario**

```ts
doubleWhitespaceCheck('<span class="foo  bar">Hola</span>')
// { error: 'DOUBLE_WHITESPACE', details: { whitespaces: ['  '] } }
```

The double space is in the class list, markup the translator neither wrote nor can fix. Same for `` or `<span  id="x">`.

**Suggested fix**

`keepHtml: true` is the right call for spacing, since dropping the tags would merge runs across them, but tag interiors should be masked before matching. Blank out tag spans while keeping their length, so runs on either side stay separate:

```ts
const withoutTagInteriors = text.replace(/<[^<>]*>/g, (tag) => 'x'.repeat(tag.length))
```

then run the existing NTC split and match over that string.

packages/app/translation-quality-checks/src/checks/singleText/doubleWhitespaceCheck.ts
```diff
@@ -0,0 +1,33 @@
...
+    /** Every double-whitespace run found in the text, e.g. `['  ', ' \t']`. */
+    whitespaces: string[]
+  }
+}>
+
+const doubleWhitespaceRegexp = /\s{2,}/g
+
+/**
+ * Single-text check: reports every run of two or more consecutive whitespace characters found in
+ * the given text. Content inside NTC regions is not checked: the text is split by NTC region and
```
kibertoad: **Finding 13 of 15.** Splitting on NTC regions hides a real double-whitespace run whenever the run is interrupted by an NTC region whose content is itself whitespace.

**Failure scenario**

Take a text of `'a '` + `NTC_START` + `' '` + `NTC_END` + `' b'`. It renders as `a`, three spaces, then `b`: a genuine triple space. `extractTextBetweenTags` yields pieces `['a ', ' b']`, neither piece contains a run, and the check returns `null`.

Related edge case: `NTC_START` immediately followed by `NTC_END` is not a region at all, because the NTM pattern is `START.+?END` and needs at least one character between the markers. An empty NTC pair present on one side and absent on the other is therefore invisible to `nonTranslatableTagsMismatchCheck` too.

The doc comment on this line asserts "a run can never span a region boundary" as though that were always the right answer. It is a false negative when the region contributes whitespace.

**Suggested fix**

Mask each NTC region rather than splitting on it: replace a whitespace-only region with its whitespace, replace every other region with a single non-whitespace sentinel, then match over the masked string. If that is more machinery than the case is worth, at least record the limitation in the doc comment instead of stating the invariant as unconditional.

packages/app/translation-quality-checks/src/detectQualityIssues.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    it('aggregates every single-text issue found', () => {
+      expect(detectQualityIssues(' Hola  mundo ')).toEqual([
+        { error: 'LEADING_WHITESPACE', details: undefined },
+        { error: 'TRAILING_WHITESPACE', details: undefined },
+        { error: 'DOUBLE_WHITESPACE', details: { whitespaces: ['  '] } },
+      ])
+    })
+
+    it('never runs the mismatch checks', () => {
```
kibertoad, line 17: **Finding 14 of 15.** The test named "never runs the mismatch checks" is vacuous: the input has no defect of any kind, so `[]` comes back regardless of which checks execute.

**Failure scenario**

```ts
expect(detectQualityIssues('Hola', { checksToInclude: ['LEADING_WHITESPACE'] })).toEqual([])
```

`'Hola'` has no leading, trailing or double whitespace and no NTC tokens, so the assertion holds even if every mismatch check ran. In the single-text form there is also no `compareWith` for them to compare against, so by construction they cannot report anything. Nothing here would catch a regression where the single-text form starts comparing `text` against itself.

**Suggested fix**

Use an input that would trip a mismatch check if one ran, and assert on the error codes:

```ts
it('never runs the mismatch checks', () => {
  const issues = detectQualityIssues(' Hola  mundo ')

  expect(issues.map((issue) => issue.error)).toEqual([
    'LEADING_WHITESPACE',
    'TRAILING_WHITESPACE',
    'DOUBLE_WHITESPACE',
  ])
})
```

packages/app/translation-quality-checks/package.json
```diff
@@ -26,12 +26,15 @@
         "clean": "rimraf dist",
         "lint": "biome check . && tsc",
         "lint:fix": "biome check --write",
-        "test": "vitest run",
+        "test": "vitest run --typecheck",
```
kibertoad, line 29: **Finding 15 of 15.** `--typecheck` was added to `test` but not to `test:ci`, and `src/types.spec.ts` does not match vitest's default `typecheck.include` glob, so the new type-contract assertions never run as typecheck tests in CI.

**Failure scenario**

`.github/workflows/ci.package.yml:48` runs `pnpm run test:ci`, which is `vitest run --coverage` with no typecheck flag, so the flag added here affects local runs only. Vitest's default `typecheck.include` is `['**/*.{test,spec}-d.?(c|m)[jt]s?(x)']`, which `src/types.spec.ts` does not match, so even locally it is not collected as a typecheck test. At runtime that file's two `it()` blocks execute nothing but `expectTypeOf` calls, which are no-ops, so they always pass. The assertions survive today only because `lint` happens to run bare `tsc` over `src/**/*`.

**Suggested fix**

Pick one and make it consistent. Either commit to vitest typecheck tests:

```json
"test": "vitest run --typecheck",
"test:ci": "vitest run --coverage --typecheck",
```

and rename `src/types.spec.ts` to `src/types.spec-d.ts` so vitest collects it; or revert this line and rely on `tsc` in `lint`, which is what actually enforces the contract today.

packages/app/translation-quality-checks/src/checks/singleText/doubleWhitespaceCheck.ts
CarlosGamero: Good catch, fixed in 6f87830b: the pattern is now `/[^\S\r\n]{2,}/g` (horizontal whitespace only), with CRLF and paragraph-break test cases. Kept tabs/mixed runs reportable rather than narrowing to literal spaces.

packages/app/translation-quality-checks/src/checks/mismatch/doubleWhitespaceMismatchCheck.ts
CarlosGamero: Fixed in 6f87830b, same pattern change as the single-text check, with a differing-line-endings test case (`'Hola\nmundo'` vs `'Hello\r\nworld'` no longer reports a mismatch).

packages/app/translation-quality-checks/src/detectQualityIssues.ts
CarlosGamero: Agreed, changed in 3090ca9b: `?? ` instead of the length check, so an explicitly empty `checksToInclude` now runs no checks. Documented in the JSDoc and README, with a spec pinning the behaviour.

packages/app/translation-quality-checks/src/detectQualityIssues.ts
CarlosGamero: Fixed in 3090ca9b: falls through to `third` when `second` is not a string (`second ?? third`), with a spec covering the runtime-undefined `compareWith` case.

packages/app/translation-quality-checks/src/utils.ts
CarlosGamero: Applied as suggested in 478f9379: `ByError` now ties each key to a check returning `Extract<Issue, { error: K }> | null`, and the wide annotations are gone so the narrow return types survive. Verified the guarantee by miswiring `DOUBLE_WHITESPACE_MISMATCH` on purpose: tsc rejects it now.

packages/app/translation-quality-checks/src/index.ts
CarlosGamero: Intentional for now: `QualityIssue` (the superset) is the only issue type we commit to, and inference gives callers the narrow type at the call site. Exporting the two sub-unions later is additive, so we'll do it as soon as a consumer actually needs to name them.

packages/app/translation-quality-checks/src/detectQualityIssues.ts
CarlosGamero: Fixed in 199658d1: `SingleTextDetectQualityIssuesOptions` and `PairDetectQualityIssuesOptions` are now exported, and the README points at them instead of the union.

packages/app/translation-quality-checks/src/checks/mismatch/nonTranslatableTagsMismatchCheck.ts
CarlosGamero: Fixed in c7540dc2: the slice is now derived from `NON_TRANSLATABLE_START_TAG.length` / `NON_TRANSLATABLE_END_TAG.length`. Agreed on the longer-term point: a token extractor in the NTM package is the natural next step, noted for a follow-up there.

packages/app/translation-quality-checks/README.md
CarlosGamero: Swept in 5f59804e: every em dash in the README and the three doc comments replaced with the punctuation the sentence needed (colons, parentheses, semicolons, or a sentence split). Verified with the grep you suggested: zero hits.

packages/app/translation-quality-checks/src/checks/mismatch/doubleWhitespaceMismatchCheck.ts
CarlosGamero: Extracted in fd9120ca: both checks now share `checks/utils/doubleWhitespaceRuns.ts` (pattern + NTC splitting in one place), with its own spec covering the horizontal-whitespace and NTC cases.

kibertoad (approved):

packages/app/translation-quality-checks/src/checks/singleText/doubleWhitespaceCheck.ts
CarlosGamero: Intentional: HTML that is not wrapped in NTC markers is regular translatable content from the library's point of view, so whitespace inside it should carry the issue. When tag markup must be protected from checks, the ingestion side wraps it as NTC, and then it is already excluded.

packages/app/translation-quality-checks/src/detectQualityIssues.ts
CarlosGamero: Short-circuit applied in 7a8c88c3 (a code lives in exactly one registry, so the dispatch stops after the single-text hit; interleaved order preserved). Declined the hoisting part: passing precomputed pieces would change every check's signature, and three regex splits per pair call don't justify that coupling.

packages/app/translation-quality-checks/src/checks/singleText/doubleWhitespaceCheck.ts
CarlosGamero: Intended behaviour rather than a limitation, and now documented as such in 7a8c88c3: an NTC region is a content token regardless of what it wraps, so whitespace on either side of it is not consecutive whitespace, same as the spaces around `%{x}` in `a %{x} b`. Only whitespace the author can see and edit as adjacent counts.

packages/app/translation-quality-checks/src/detectQualityIssues.spec.ts
CarlosGamero: Applied as suggested in 1795dd62: the spec now feeds a defect-carrying input and asserts the exact error codes.

packages/app/translation-quality-checks/package.json
CarlosGamero: Made consistent in 0c6cfc18 by deriving one script from the other: `test:ci` is now `pnpm run test --coverage`, so CI runs the same `--typecheck` as local. The `expectTypeOf` assertions in `types.spec.ts` are enforced by the bare `tsc` in `lint` (which CI also runs); we kept the file on the regular spec glob on purpose.

PR #1234, opened 2026-08-31
feat: TRAP-1579 add Expert's missing languages

tomaszhanc:
## Changes

[TRAP-1579](https://lokalise.atlassian.net/browse/TRAP-1579)

The Expert TM migration drops around 648k translation memory records because this package rejects 21 language codes that Expert's language table uses. 13 of them are real languages with valid IANA-registered codes, so this PR adds them. 

The other 8 turned out to be bad data (invented codes like `fl` and `kh`, made-up regions like `tlh-KL`) and are written off with reasons in [a ticket comment](https://lokalise.atlassian.net/browse/TRAP-1579?focusedCommentId=285283). We need to decide what to do with these.

## Checklist

## AI Assistance Tracking

[TRAP-1579]: https://lokalise.atlassian.net/browse/TRAP-1579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomaszhanc:
@coderabbitai review

packages/app/supported-languages/src/constants/languages.ts
CarlosGamero, line 1: 🟡 Some of the new new languages are Arabic-script RTL languages, but rtlLanguages was not updated (so the exported `getLocaleDirection()` now returns a confidently wrong 'ltr' for them)

CarlosGamero, line 1: 🔴 CC finding, please check it:
languages.ts documents itself as 'List of all languages supported by CLDR43' (README repeats it), but the 13 added codes are not in any CLDR release and are interleaved with no marker, so regenerating the file from the documented cldr-json source silently drops them.

We are using CLDR43 in this package, adding other tags needs to be discussed as it affects Vantage

(Also, it can affect to the displayName utility and not work as expected for those languages)

PR #1235, opened 2026-09-01
feat(universal-testing-utils): request-derived mocks for defineApiContract

kibertoad:
## Description

`ApiContractMockttpHelper` and `ApiContractMswHelper` could only mock a **fixed** body, via `mockResponse`. The request-derived variant existed only as `MswHelper.mockValidResponseWithImplementation`, which the compatibility matrix ties to the legacy `buildRestContract`/`buildSseContract` definitions. So a `defineApiContract` consumer whose mock needs to read the incoming request had no option but to drop down to raw mockttp or msw.

Found while auditing a service that had finished migrating to `defineApiContract`: its e2e specs hand-roll `mockttp` with `thenCallback` handlers that build responses out of the request body, and `universal-testing-utils` had nothing to offer them.

### `mockResponseWithImplementation`

Added to both `ApiContract*` helpers. `responseStatus` still selects the contract entry, which is what types `handleRequest`'s return value and supplies the Zod schema the result is validated against.

```ts
// mockttp
await helper.mockResponseWithImplementation(postUserContract, {
  responseStatus: 200,
  handleRequest: async (request) => {
    const body = await request.body.getJson()
    return { id: `id-${body.name}` }
  },
})

// msw
helper.mockResponseWithImplementation(postUserContract, {
  responseStatus: 200,
  handleRequest: async ({ request }) => ({ id: `id-${(await request.json()).name}` }),
})
```

Only JSON entries are addressable. An SSE-only or blob-only status collapses the params type to `never`, so it is rejected at compile time; a runtime guard with a pointer back to `mockResponse` covers untyped callers. Both cases are tested.

### Per-call status via `response()`

Both helpers gain a static `response()`, mirroring `MswHelper.response()`:

```ts
handleRequest: () => callCount++ === 0
  ? ApiContractMockttpHelper.response({ id: 'first' }, { status: 500 })
  : { id: 'second' }
```

The `Symbol` brand behind that wrapper moved out of `MswHelper` into a new `responseWrapper.ts` so all three helpers share **one** brand. Without that, a wrapper made by `MswHelper.response()` would not be recognised by the new helpers. `MswHelper.response()` and `MockResponseWrapper` keep their existing signatures and export paths, so its public surface is unchanged.

### Review follow-up (249aa6fc)

- The reply now carries the media type the contract declared it under. `resolveJsonSchema` dropped the content-map key, so an `application/problem+json` entry went out as `application/json` and the client rejected its own contract's response.
- `contentType` selects between JSON media types, matching the field `mockResponse` already takes. Required when a status entry declares more than one, since the handler's result has to match a specific schema.
- A per-call status override resolves against its own contract entry, so a body wrapped with `response({ status })` is validated against the schema for the status actually sent.
- A request negotiating the SSE branch of a dual JSON/SSE status gets a 406 pointing at `mockResponse`, instead of JSON it cannot read.
- Handler and validation failures are reported through `console.error` and a labelled 500. mockttp and msw otherwise bury a throwing route callback in an opaque 500 that reaches the test as a client-side parse error.
- The handler's request argument is typed from the contract's `requestBodySchema`, so `getJson()` and `request.json()` no longer need a cast at every call site.
- The wrapper brand switched to `Symbol.for`, so two copies of the package in one tree still recognise each other's wrappers.

## Verification

- `pnpm run lint` (biome + tsc): clean
- `pnpm test`: 203 passed, covering the content-map fixtures the new method previously never touched (non-default JSON media types, multi-JSON selection, dual JSON/SSE negotiation, no-body and blob-only statuses) and asserting the emitted content-type header
- `pnpm run build`: clean; confirmed the new methods and `MockResponseWrapper` appear correctly in the emitted `.d.ts`

## AI Assistance Tracking

### DIFF
--- a/packages/app/universal-testing-utils/README.md
+++ b/packages/app/universal-testing-utils/README.md
@@ -255,14 +255,14 @@ await helper.mockResponse(contract, { responseStatus: 201, responseJson: { id: '
 
 `mockResponse` takes a fixed body. When the response has to depend on what the caller sent, use `mockResponseWithImplementation` and return the body from a handler.
 
-`responseStatus` still selects the contract entry, which is what types `handleRequest`'s return value and supplies the Zod schema the result is validated against. Only JSON entries are addressable this way: SSE and blob responses stay static, via `mockResponse`.
+`responseStatus` still selects the contract entry, which is what types `handleRequest`'s return value and supplies the Zod schema the result is validated against. The handler's request argument is typed from the contract too, so `request.body.getJson()` (mockttp) and `request.json()` (msw) hand back the output of the contract's `requestBodySchema` rather than `unknown`. Only JSON entries are addressable this way: SSE and blob responses stay static, via `mockResponse`.
 
 ```ts
 // mockttp: the handler receives the mockttp CompletedRequest
 await helper.mockResponseWithImplementation(postUserContract, {
   responseStatus: 200,
   handleRequest: async (request) => {
-    const body = await request.body.getJson()
+    const body = await request.body.getJson() // typed by requestBodySchema
     return { id: `id-${body.name}` }
   },
 })
@@ -271,7 +271,7 @@ await helper.mockResponseWithImplementation(postUserContract, {
 helper.mockResponseWithImplementation(postUserContract, {
   responseStatus: 200,
   handleRequest: async ({ request }) => {
-    const body = await request.json()
+    const body = await request.json() // typed by requestBodySchema
     return { id: `id-${body.name}` }
   },
 })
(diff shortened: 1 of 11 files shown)


CarlosGamero (approved):

packages/app/universal-testing-utils/README.md
```diff
@@ -250,6 +251,61 @@ await helper.mockResponse(contract, { responseStatus: 201, responseJson: { id: '
...
+### mockResponseWithImplementation
+
+`mockResponse` takes a fixed body. When the response has to depend on what the caller sent, use `mockResponseWithImplementation` and return the body from a handler.
+
+`responseStatus` still selects the contract entry, which is what types `handleRequest`'s return value and supplies the Zod schema the result is validated against. Only JSON entries are addressable this way: SSE and blob responses stay static, via `mockResponse`.
+
+```ts
+// mockttp: the handler receives the mockttp CompletedRequest
+await helper.mockResponseWithImplementation(postUserContract, {
+  responseStatus: 200,
```
CatchMe2, line 263: Would it make sense to add support for setting `responseContentType`? Contract can have multiple content types per status code defined.

packages/app/universal-testing-utils/src/api-contracts/types.ts
```diff
@@ -65,6 +68,26 @@ export function resolveContractEntry(
...
+    return responseEntry
+  }
+
+  if (!responseEntry.content) {
+    return undefined
+  }
+
+  return Object.values(responseEntry.content).find((descriptor): descriptor is z.ZodType =>
+    isJsonBody(descriptor),
+  )
```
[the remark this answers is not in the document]
kibertoad: Fixed in 249aa6fc.

`resolveJsonSchema` is now `resolveJsonTarget` and returns `{ mediaType, schema }`: the content-map key for content entries, `application/json` only for bare-schema entries. Both helpers set the response content-type from it, and the msw path moved off `HttpResponse.json` (which forces the header) onto `new HttpResponse`.

Regression test in both suites, `replies with the declared JSON media type, not application/json`, against a new `problemJsonContentApiContract` fixture that declares only `application/problem+json`. It asserts the emitted header and then round-trips the response through `sendByApiContract`, which is where the old behaviour failed: the client rejected its own contract's response with `UNEXPECTED_RESPONSE_ERROR`.

Related: a status entry declaring several JSON media types now takes a `contentType` param to pick between them, see the reply on the README thread.

packages/app/universal-testing-utils/README.md
kibertoad: Yes, added in 249aa6fc. Named `contentType`, to match the field `mockResponse` already takes for the same job.

It is required when the status entry declares more than one JSON media type and optional otherwise, and the descriptor it selects is what types `handleRequest`'s result:

```ts
await helper.mockResponseWithImplementation(multiJsonContentApiContract, {
  responseStatus: 200,
  contentType: 'application/problem+json',
  handleRequest: () => ({ title: 'Invalid', detail: 'Something went wrong' }),
})
```

Omitting it on an ambiguous entry is a type error, and a runtime throw for untyped callers: `Status 200 declares more than one JSON content type (application/json, application/problem+json); pass contentType to select one`. Before this it silently served the first JSON descriptor while the types accepted any of them, so returning the wrong variant compiled and then failed schema validation at request time.

`contentType` only names JSON entries. Pointing it at an SSE or blob descriptor throws, since a handler can only produce JSON. A status declaring both JSON and `text/event-stream` still serves its JSON side, but a request negotiating SSE through `Accept` now gets a 406 pointing at `mockResponse` rather than a JSON body it cannot read.

Independently of `contentType`, the reply now carries whatever media type the contract declared, so a lone `application/problem+json` entry is no longer flattened to `application/json`.

PR #1253, opened 2026-09-02
docs: scheduling library extraction analysis (proposal review only, not for merge)

kibertoad:
> **Proposal review only. This PR is not intended to be merged.** It exists so `docs/scheduling-library-analysis.md` can be commented on inline. Once the direction is agreed, the work is tracked in implementation PRs and this one is closed.

## The question

Maestro and CTENG both run recurring work, with nothing shared. Should the scheduling logic become a library in this repo, and if so, which layers?

## What the analysis concludes

**Two mechanisms exist today, different at every layer.** CTENG stores schedules in Postgres with a `next_run_at` column and a periodic runner ticks and dispatches whatever is due. Maestro registers BullMQ job schedulers with a custom RRULE `repeatStrategy`. They differ in recurrence model, next-occurrence engine, storage, trigger delivery and execution tracking. Building a third mechanism inside CTENG, which the task as posed assumed, is the one option the document argues against.

**Share the recurrence engine first.** A schedule definition plus "next occurrence after T in timezone Z" is pure logic with no infrastructure dependency, and it is where both services carry their hardest code (DST, timezone conversion, start boundaries) and most of their tests. Low risk, immediate value, no runtime change.

**The definition is RFC 5545 rather than a model of our own.** Both products' capabilities were read off the code and mapped part by part: everything they express needs six of RRULE's fourteen rule parts (`FREQ`, `INTERVAL`, `BYDAY`, `BYHOUR`, `BYMINUTE`, `BYMONTHDAY`), and none of the awkward ones. CTENG's 1-to-4 times per day are a recurrence set of up to four rules, and its local-minute hourly schedule is `FREQ=HOURLY;BYMINUTE=M` under a `TZID`. So RRULE is the lingua franca, narrowed by a published profile, with a codec in core: the encoder mostly exists in maestro today, and the decoder that the UI needs is one bounded decision tree over those six parts, guaranteed by a `decode(encode(form))` property test. Out-of-profile rules decode to `custom` and render read-only. Each product keeps its own form schema as its API contract.

**Trigger delivery goes behind one adapter interface, with pg-boss and BullMQ equally supported.** The document deliberately does not pick a backend: that waits on a broader org architectural decision about which paths are recommended. It records the trade-off as input (Postgres commits are synchronously durable and let a schedule be registered or cancelled in the caller's transaction; Redis persistence is best-effort and maestro's schedules vanish with the Redis instance, the failure mode its CUJ-002 journey doc records as revenue-generating workflows silently not running; BullMQ fires faster, is in production, and needs no new infrastructure). Adopting the library commits a service to neither backend, and a later switch is a cutover rather than a rewrite.

**No library-owned schedule table.** Each service's domain row stays the canonical registry; the pending job or job scheduler is only the runtime record. `reconcile(desired)` diffs the two and ships in the first release, which is what gives maestro a recovery path for a lost schedule where today the answer is "re-activate the workflow".

**The plan assumes pg-boss gets native recurrence.** `schedule()` is cron only today, so the adapter would otherwise drive a self-perpetuating chain of deferred jobs, the ugliest part of the design. The change that removes it (pluggable recurrence kinds) is implemented and open as [timgit/pg-boss#886](https://github.com/timgit/pg-boss/pull/886) for 12.30.0, and the document plans on it being adopted: the adapter is a direct mapping onto `boss.schedule`, `boss.unschedule` and `boss.getSchedules`. It is built last in the outline, so if 12.30.0 is not out by then the chain ships as a bridge and switches internally later, with no caller change and no migration of running schedules under three rules the document spells out.

## What needs review

1. **RRULE as the canonical model, with a profile?** The profile table in "Engine and model: trade-offs" is the thing to check: it claims six rule parts cover both products, derived from `rruleUtils.ts#translateFilters` on one side and `scheduleSchemas.ts` plus `computeNextRunAt.ts` on the other. If either team knows of a schedule shape that is not in that table, the recommendation changes. The cost being accepted is the decoder (roughly 150 lines and its fixtures) that an invented model would not need.
2. **Both adapters supported, backend choice deferred?** The alternative is picking one now. Supporting two is a standing cost: two container setups in CI, two runbooks, and a parity contract to maintain. The document accepts that cost deliberately rather than pre-empting the org decision. Is that the right call, and should this proposal feed into that decision explicitly?
3. **Engine implementation: `rrule-rust` or a pure-JS port?** `rrule-rust` is proven in maestro and ships prebuilt binaries for 14 targets, but it is a native module that CTENG's image and lockfile would start carrying. Recommended behind a library-agnostic API with a shared fixture corpus as the escape hatch.
4. **Build order.** Engine and codec, then CTENG on the engine, then facade plus BullMQ adapter, then maestro on it, then the pg-boss adapter. Maestro's step is a pure refactor with no infrastructure change and it closes the CUJ-002 gap (a schedule lost in Redis, with no reconciliation job today) via `reconcile`. Does anything need to move earlier?
5. **`catchUp: 'skip' | 'once' | 'all'`** now has a per-backend semantics table in "Keeping the two adapters interchangeable", with `all` as a pg-boss-only capability that BullMQ rejects at call time. What is still open is the default, and it has to be settled before the first release, because today's behaviour differs per backend by accident. Which default do the products want?
6. **DST policy.** RFC 5545 says nothing about a gap or a repeated hour, so the engine owns it. Maestro bumps by an hour and rethrows if that is not enough; CTENG shifts forward into the gap and takes the first instance of a duplicated hour. One of these has to win, and the fixture corpus is where it gets pinned.
7. **Ownership.** A shared scheduler sits on revenue-critical paths in two teams' services. Who owns it in `shared-ts-libs`?

One incidental finding, worth fixing in maestro regardless of this proposal: `monthDayPosition: 'current'` compiles to `byMonthday: [new Date().getUTCDate()]`, the day the compile happens to run on, so a recompile silently moves the schedule, and days 29 to 31 skip short months.

Sections worth reading closely: "Engine and model: trade-offs" for the profile and codec argument, "Layer 2: trigger delivery" for the adapter interface and the parity rules, "Upstream: native recurrence lands in 12.30.0" for what the adapter assumes and what it falls back to, and "Risks and things to settle early" for what is knowingly unresolved.

No changeset: no published package code changes.

## AI Assistance Tracking


docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,558 @@
...
...o make. The scheduling record is mission-critical: a lost entry means a revenue-generating workflow silently stops running, which is CUJ-002's failure mode. Redis persistence is best-effort by design (RDB snapshots lose everything since the last dump, AOF with the default `everysec` fsync still loses up to a second of acknowledged writes, and a failover promotes a replica that may be behind the primary), while Postgres commits are synchronously durable and the service's Postgres already carries backups, point-in-time recovery, replication and monitoring that the Redis instance does not match. Keeping the schedule next to the domain data means one durability story, one backup to restore, and the option of enqueuing the next occurrence in the same transaction as the domain write. What it costs: a new dependency and a `pgboss` schema in any service that does not already run it, fire precision bounded by the poll interval instead of Redis latency, and (until [#886](https://github.com/timgit/pg-boss/pull/886) lands) the chain below. Speed is not the deciding factor for schedules that fire minutes to months apart, but the operational surface is a real cost and lands on whoever adopts it.
```
kibertoad: this is a wall of text that is hard to read. structure it better

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,558 @@
...
+
+- `get()` returns the structured definition and the next fire time, the same as the pg-boss adapter, by carrying the definition in the job scheduler's template data. Confirmed against the bullmq 6 job-scheduler API that `background-jobs-common` develops against: `getScheduler(key)` returns `template.data` alongside `next`, `pattern` and `tz`. This requires `upsertJobScheduler`, not the legacy repeatable path, which stores the pattern only.
+- The pattern-compare-then-remove logic for edits moves into the adapter so the dedup edge case is handled once.
+- Redis is the firing source of truth here, so `reconcile` matters more on this adapter than on pg-boss, where the pending row is covered by the service's own backups. It ships in the first release for both.
+- No transactional enqueue: a fire cannot be committed with the service's domain write. This is inherent to the backend and is a documented capability gap, not a defect to fix in the adapter.
+- Housekeeping cron schedules stay on BullMQ's own repeat handling, matching what pg-boss's native `schedule()` does on the other side.
+
+#### Parity is a contract, not an accident
```
kibertoad: get rid of this llmism and check document for other clear llmisms that make document hard to read

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,558 @@
...
+
+`list(owner)` on pg-boss is `findJobs({ data: { owner } })`, a top-level key match. It is adequate for admin tooling; customer-facing lists come from the product tables.
+
+### Layer 3: execution tracking and domain dispatch (do not share)
+
+CTENG's `schedule_execution` model (conflict keys, queued drains, Autopilot correlation, TTL recovery) and maestro's `RunWorkflowJob` (start a workflow run, entitlement checks, delay histograms) are domain code. The library exposes `onFire` and `onDispatchFailed` and leaves the rest to the service. Because pg-boss caps `active` time at 24 hours (`expireInSeconds`), the fire handler should enqueue domain work rather than run it inline.
+
+## pg-boss research: what it supports and whether it fits
+
+Checked against pg-boss 12.29.0 (released 2026-08-30), its type definitions, `src/plans.ts`, `src/timekeeper.ts` and the docs at pgboss.io. Requirements: Node 22.12 or newer, PostgreSQL 13 or newer. Runtime dependencies: `pg`, `cron-parser` 5, `serialize-error`. Both services satisfy the Node floor.
```
kibertoad: this is irrelevant, we know this

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,558 @@
...
+| `reconcile(desired)` | `findJobs` per id, `send` for missing, `cancel` for stale | Full, library code |
+| Missed occurrence while no worker ran | job fires on first fetch after `startAfter`; handler computes next from now | Equivalent to CTENG's "run once, advance" |
+| Fixed housekeeping crons | native `schedule()` | Full |
+
+**Does anything need to change upstream first?**
+
+No. Every operation maps to a public, documented pg-boss 12 API, so the chain adapter can ship without waiting on anyone. But the chain is a workaround: the library re-implements, in application code, the "recurring job" concept that pg-boss already has for cron, and pays for it with `singletonKey` discipline, a mandatory `reconcile`, and a `get()` that scrapes job payloads. The next section proposes public API changes that would remove most of that. Two constraints stay regardless: `expireInSeconds` caps handler time at 24 hours (dispatch should enqueue domain work, not run it), and `useListenNotify` is incompatible with PgBouncer transaction pooling (accept polling precision or give pg-boss a direct connection).
+
+## Upstream pg-boss API proposals
```
kibertoad: no need to enumerate different options, we already implemented the best one, incorporate that fact into the doc and make assumption it will be adopted eventually, plan around that

docs/scheduling-library-analysis.md
kibertoad: Restructured into **What it gives** (durability, one backup story, transactional enqueue), **What it costs** (new dependency plus `pgboss` schema, poll-interval precision, the chain before 12.30.0) and **How it works** as four numbered steps. The Redis-persistence detail is one bullet now instead of a clause chain, and the chain mechanics moved out to the upstream section where the fallback is described.

docs/scheduling-library-analysis.md
kibertoad: Heading is now "Keeping the two adapters interchangeable". Swept the rest of the document for the same reflexes and removed: "stated as a trade-off rather than a verdict", "worth naming", "tested rather than argued", "neither extreme", "honest limit", "not a defect to fix in the adapter", "not a stepping stone", "not a dialect", "does not survive contact with the numbers", plus the `X, not Y` constructions that were not carrying the point. Also folded the standalone `Status:` line into the header and dropped the "revised recommendation" framing, since the current position is all a reader needs.

docs/scheduling-library-analysis.md
kibertoad: Cut. The section is now "pg-boss: the primitives the adapter uses" and it is one table mapping each primitive to what the adapter does with it, plus the two constraints worth designing around (`expireInSeconds` capping `active` at 24 hours, `useListenNotify` versus PgBouncer transaction pooling). Gone: the version and release date, the Node and Postgres floors, the runtime dependency list, the native-cron and operations tables, the per-block verdicts, and the "fit against the model" table. Also dropped the duplicate mention of the 24-hour cap from the Layer 3 section.

docs/scheduling-library-analysis.md
kibertoad: Done. The document now plans on #886 being adopted rather than enumerating alternatives.

- The six numbered proposals and the "does anything need to change upstream first" section are gone. What replaces them is one section, "Upstream: native recurrence lands in 12.30.0": what #886 changes, and what the adapter becomes on top of it (`schedule`/`reschedule` are `boss.schedule`, `unschedule` is `boss.unschedule`, `get` is `boss.getSchedules`, the schedule row is the registry, no chain and no `singletonKey` bookkeeping).
- The competing `repeat`-on-`send` shape is deleted, along with the three-outcome decision table and the "if it is declined, push for X" branches.
- The chain survives only as a subsection, "If 12.30.0 is not released when the adapter is due", since the adapter is built last in the outline and may never need it. The three rules that keep the internal switch invisible to callers stay, because they are what makes planning on adoption safe.
- The remaining small asks (`getSchedule`/`previewSchedule`, transactional `work()`, `findJobs` filters, doc fixes) are a four-bullet follow-up list rather than numbered proposals with problem and cost sections.

Net: 156 lines shorter.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+## Target design
+
+### Layer 1: recurrence definition and next-occurrence engine (share)
+
+A `ScheduleDefinition` that is an RFC 5545 recurrence (`DTSTART;TZID` plus one to four `RRULE`s), a codec between it and each product's form, and one function:
+
+```ts
+nextOccurrence(definition, { after: Date, startAt?: Date }): Date | undefined
+encode(form): ScheduleDefinition        // total, canonical serialisation
+decode(definition): Form | 'custom'     // deterministic inside the published profile
```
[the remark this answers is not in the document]
kibertoad: Fixed. `ScheduleDefinition` is a branded string now, canonical RFC 5545 text and nothing else, and the codec is bound to one product's form schema:

```ts
type ScheduleDefinition = string & { readonly brand: unique symbol }

codec<Form>(schema: ZodType<Form>): {
  encode(form: Form): ScheduleDefinition
  decode(definition: ScheduleDefinition): Form | 'custom'
}
```

So `decode` cannot hand a caller the other product's shape, and `encode` is the only thing that mints a definition.

The double-encode reading came from the pg-boss step, which said `schedule(job)` "encodes the definition". It passes it through now: `boss.schedule(queue, { kind: 'rrule', expression: job.definition }, ...)`. Option A's wording that called `ScheduleDefinition` an `RRuleSet` is gone as well; the `RRuleSet` is a parse artefact inside the engine and never crosses the API.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+Read off maestro's compiler (`rruleUtils.ts#translateFilters`) and CTENG's schema and engine (`scheduleSchemas.ts`, `computeNextRunAt.ts`):
+
+| RRULE part | Maestro uses | CTENG needs | In the profile |
+| --- | --- | --- | --- |
+| `FREQ` | `MINUTELY`, `HOURLY`, `DAILY`, `WEEKLY`, `MONTHLY` | `HOURLY`, `DAILY`, `WEEKLY` | those five |
+| `INTERVAL` | minutely, hourly, weekly, monthly (`repeatEvery`) | not used | yes, integer >= 1 |
+| `BYDAY` | weekly selection, and a filter on the minutely and hourly windows | `days: 1..7` (Mon to Sun, ISO order) | yes, weekday set, no ordinal prefixes |
+| `BYHOUR` | one hour (daily, weekly, monthly) or a contiguous range (windows) | implied by `times[]` | yes, list of hours |
+| `BYMINUTE` | one minute; never combined with `FREQ=MINUTELY` | the hourly minute, and each entry of `times[]` | yes, list of minutes, forbidden with `FREQ=MINUTELY` |
+| `BYMONTHDAY` | `1`, `-1`, or the start date's day-of-month | not used | yes, single value, 1 to 28 or -1 |
```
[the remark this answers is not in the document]
kibertoad: Specified. The bullet now says what happens to rules that already store 29 to 31: `decode` still reads them, the profile validator marks them out of profile so the form renders read-only, the encoder never emits them, and a one-off audit lists the affected workflows for their owners to re-pick between a fixed day and `-1`.

Not normalising is the deliberate part. "Fires on the 31st" and "fires at month end" are different intents, and the compile date does not record which one was meant, so choosing for the user would move schedules exactly as silently as the current bug does. February in a leap and a non-leap year is a fixture, for 29, 30, 31 and `-1`.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+  payload: unknown                 // what the service needs at fire time
+  nextRunAt: Date | undefined
+}
+
+interface DeliveryAdapter {
+  schedule(job: ScheduledJob): Promise<void>            // register first occurrence
+  reschedule(job: ScheduledJob): Promise<void>          // recreate with a new definition
+  unschedule(id: string): Promise<void>
+  get(id: string): Promise<ScheduledJob | undefined>
+  list(filter: { ids?: string[]; owner?: string }): Promise<ScheduledJob[]>
```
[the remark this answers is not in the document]
kibertoad: Both findings fixed, in the interface itself.

`owner` is a field on `ScheduledJob` rather than only a filter argument, and each adapter has to store it somewhere it can filter on: the top level of the pg-boss schedule data, template data on BullMQ. The pg-boss step spells that out.

`list` is bounded and documented as live schedules only:

```ts
list(f: { ids?: string[]; owner?: string; limit: number; cursor?: string }):
  Promise<{ jobs: ScheduledJob[]; cursor?: string }>
```

`reconcile` reads the backend through it, which is why the bound belongs in the contract rather than in one adapter. The upstream follow-up for `findJobs` filters is reworded to match: without `states`, `limit`, `orderBy` and a cursor, the chain adapter has to filter and page in memory to keep that promise, which is the reason for the ask.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+  onFire(handler: (job: ScheduledJob, firedAt: Date) => Promise<void>): void
```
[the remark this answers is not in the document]
kibertoad: Fixed. `onFire` takes a context, and `onDispatchFailed` is in the interface instead of being mentioned only in Layer 3:

```ts
type FireContext = {
  fireId: string                   // `${id}@${occurrence.toISOString()}`, same on every delivery
  firedAt: Date                    // the occurrence, not the delivery time
  db?: TransactionRunner           // only where the backend can commit with the domain write
}

onFire(handler: (job: ScheduledJob, ctx: FireContext) => Promise<void>): void
onDispatchFailed(h: (job: ScheduledJob, ctx: FireContext, error: Error) => Promise<void>): void
```

The transactional claim was overstated and is now narrower. `ctx.db` is pg-boss only and covers the handler's own writes: domain rows and any job the handler enqueues commit together. Completing the fired job in that same transaction needs the upstream `work({ transactional: true })` ask that was already on the follow-up list, so until it lands a fire is at-least-once on pg-boss too. The doc states that, and states where the recurrence advance sits relative to `onFire` for each strategy: pg-boss native advances the schedule row before the job is forwarded to the queue, while the chain commits the successor send before `onFire` runs. `fireId` rather than a transaction is what handlers dedupe on.

The pg-boss "what it gives" bullet is "transactional writes" now, meaning register, reschedule and unschedule commit with the domain row they belong to, which is the part that actually holds under native recurrence.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+
+#### BullMQ adapter
+
+Wraps BullMQ job schedulers as maestro does today, with the engine's `nextOccurrence` supplied as `repeatStrategy`. Maestro adopts the library through it and deletes its three hand-rolled `QueueManager` subclasses without changing what fires when. It is supported on the same terms as the pg-boss adapter, with no deprecation.
+
+- `get()` returns the structured definition and the next fire time, the same as the pg-boss adapter, by carrying the definition in the job scheduler's template data. Confirmed against the bullmq 6 job-scheduler API that `background-jobs-common` develops against: `getScheduler(key)` returns `template.data` alongside `next`, `pattern` and `tz`. This requires `upsertJobScheduler`, not the legacy repeatable path, which stores the pattern only.
```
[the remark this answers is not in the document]
kibertoad: Correct, and that predicate would have paged on healthy schedules the moment the native path shipped. The risks bullet is per backend now: on the chain a live schedule has a pending job, on BullMQ a scheduler with a `next`, and on pg-boss native only a `schedule` row with a future `next_run_at`, with no pending job at all between occurrences. `list()` is what normalises the three, so the alert stays one diff against the domain table.

The "where the definition lives per adapter" table had the same assumption baked in, reading "pending job payload, via `findJobs` by key" for pg-boss. It now reads schedule row `expression` natively, pending job payload on the chain, which is also what makes `get()` work between fires without a pending job.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+- Housekeeping cron schedules stay on BullMQ's own repeat handling, matching what pg-boss's native `schedule()` does on the other side.
+
+#### Keeping the two adapters interchangeable
+
+Two equally supported backends stay equal only if the library enforces it:
+
+- **One conformance suite, run against every backend.** The same test file, parameterised over adapters (Redis container, Postgres container), covering register, reschedule, unschedule, read, list, missed occurrences, DST boundaries and reconcile. This is the mechanism that keeps "adopting the library does not commit you to a backend" true.
+- **The facade never leaks a backend identifier.** No pg-boss job id, no BullMQ scheduler key, in any return type. Callers key on their own schedule id.
+- **The option surface is what both adapters can honour**, `catchUp` included. An option only one backend can implement goes in the capability matrix and is rejected at call time by the other, rather than silently behaving differently.
```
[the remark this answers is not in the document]
kibertoad: Settled, as a table rather than a rule that quietly could not hold:

| `catchUp` | pg-boss native | pg-boss chain | BullMQ |
| --- | --- | --- | --- |
| `skip` | `missed: 'skip'` | handler still chains the successor, skips `onFire` when the occurrence is older than one poll interval | adapter skips `onFire` on the same rule |
| `once` | `missed: 'once'` | overdue job fires, successor computed from now | the scheduler's own behaviour |
| `all` | `missed: 'all'`, capped at 1000 | engine enumerates the missed occurrences, capped at 1000, each sent under `${id}#catchup:${iso}` | rejected at call time |

The chain can honour `all` because the engine enumerates occurrences anyway for `previewOccurrences`, and the separate key namespace gives each replayed occurrence its own idempotency without touching the series' own `singletonKey` slot. That is what matters for the rule you were poking at: both pg-boss strategies honour all three values, so the internal switch stays invisible to callers. `all` on BullMQ is rejected at call time and sits in the capability matrix as a pg-boss-only row.

On ordering: replays go out ascending, each carries its own `fireId` and `firedAt`, and delivery order is not guaranteed with more than one worker running, so a handler that cares about sequence orders on `firedAt`.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,442 @@
...
+- `getSchedules()` reports `kind`, `expression`, `nextRunAt` and `lastRunAt`.
+
+What the adapter becomes on top of it: `schedule` and `reschedule` are `boss.schedule`, `unschedule` is `boss.unschedule`, `get` is `boss.getSchedules`. The schedule row is the registry, so there is no chain, no `singletonKey` bookkeeping and no payload scraping, and `reconcile` narrows to a consistency check against the service's own table. Ours to keep: the recurrence engine, registered as the `rrule` parser, and `previewOccurrences`.
+
+### If 12.30.0 is not released when the adapter is due
+
+The adapter ships the same public surface over a self-perpetuating chain of deferred jobs, then switches to native recurrence internally:
+
+1. `schedule(job)` computes the first occurrence and calls `send(queue, { id, definition, payload }, { startAfter: fireAt, singletonKey: id })`.
+2. The handler computes the next occurrence from the definition in the job data, sends the successor with the same `singletonKey`, then invokes `onFire`. Under the `short` policy at most one job per key sits in `created`, so a retried handler cannot double-book.
```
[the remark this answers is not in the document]
kibertoad: Right: `short` dedupes the successor and does nothing about a retry of the current handler. Chain step 2 says so now. The successor send has already committed when `onFire` runs, so a handler that fails after that point is retried against a successor that exists, and `ctx.fireId` (`${id}@${occurrence.toISOString()}`, identical on every delivery of one occurrence) is what keeps the domain work single. The window is called out as a test case.

The guarantee is stated once in Layer 2 rather than per adapter, because the same window exists under native recurrence and on BullMQ: delivery is at-least-once on every backend, and the library does not dedupe domain work on a handler's behalf.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,562 @@
...
+2. **CTENG adopts the engine.** `computeNextRunAt` delegates to `nextOccurrence`. Storage and runner unchanged. Run both implementations side by side in tests over a year of dates in the timezones CTENG customers use, then delete the local implementation. This is where CTENG's `times[]` first becomes a list of rules, still behind its own storage.
+3. **Facade and BullMQ adapter.** Build the `Scheduler` facade, the `DeliveryAdapter` interface and `reconcile`, then the BullMQ adapter as a thin wrapper carrying maestro's pattern-compare-then-remove logic. No pg-boss yet. Settle the `catchUp` option set here, since both adapters have to honour it in full and step 5 must not widen it, and the fan-out and grouping of a multi-rule recurrence, since BullMQ needs it as much as pg-boss does.
+4. **Maestro on the BullMQ adapter.** Replace `RunWorkflowJobScheduler`, `DetectStuckWorkflowRunsJobScheduler` and `TasksAboutToExpireJobScheduler` with facade calls. Add the periodic `reconcile` fed from Live scheduled workflows. The CLI script switches to `scheduler.list()`. No runtime behaviour change; CUJ-002's missing-scheduler failure mode gains a recovery path.
```
armababy, line 544: 🔶 Small correction to "no runtime behaviour change": the engine already changes in step 1, so we do get different behaviour at DST edges even though we stay on BullMQ. The doc says the DST policy becomes whatever rrule-temporal does and gets "pinned by the fixture corpus", but doesn't say what that behaviour is or how it differs from our one-hour bump today. Could that be spelled out here? And could the step 1 side-by-side include our actual prod schedulers from Redis (pattern, tz, startDate), not just fixtures? Otherwise we only find out after the swap.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,562 @@
...
+7. **Backend changes are available, not scheduled.** Maestro on pg-boss is one config swap plus a cutover (register on pg-boss and unregister from BullMQ on next activation or edit, then `reconcile` the remainder in a controlled batch), and CTENG on BullMQ is the mirror image. Neither is planned here. Both wait on the org-level decision about which paths are recommended, and the point of steps 1 to 6 is that whichever way it goes, the work is a cutover rather than a rewrite. Independently of it: delete maestro's dead `SCHEDULER_REDIS_*` config.
```
armababy, line 547: 🟢 This reads as if parking the maestro move settles it, but the doc's own reason for pg-boss is that our schedules vanish with Redis, and after step 4 reconcile is the only thing covering that for us. Worth a sentence here saying the maestro durability gap stays open until the org call and reconcile is the stopgap, so nobody reads "available, not scheduled" as done.

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,562 @@
...
+- **Catch-up is a backend option now, and its default is the wrong one for both products.** `missed: 'skip' | 'once'` landed with #898, and `skip` is the default, which sends nothing for an occurrence that came due while no pass ran. CTENG fires a missed occurrence on its next tick and BullMQ fires one on resume, so both products behave as `once` today and both have to ask for it explicitly. What `once` covers is a gap in the timekeeper, measured from the pass timestamp the `version` row already carries and bounded below by each schedule's own `created_on`: a deploy window, an outage, a deployment running with scheduling switched off. What it does not cover is a job that was sent and then dead-lettered, which is the queue's business, or a schedule missing from the table altogether, which is `reconcile`'s. Settle per product which policy each schedule kind registers with, and keep the fire handler out of it: on pg-boss it cannot tell a caught-up delivery from a due one.
```
armababy, line 559: 🟢 The catch-up section says both products register once, but step 4 doesn't mention catch-up for maestro at all. Since the pg-boss default is skip and BullMQ today effectively behaves like once (one overdue run on resume), leaving it implicit could change behaviour for us without anyone noticing. Could step 4 state it?

docs/scheduling-library-analysis.md
```diff
@@ -0,0 +1,562 @@
...
+- **Ownership.** A shared scheduler touches revenue-critical paths in two teams' services. Agree on a code owner in `shared-ts-libs` before step 3.
```
armababy, line 560: 🔶 The doc says ownership has to be settled before step 3, but nothing in it proposes who, and CODEOWNERS in shared-ts-libs is one repo-wide default line, so there's nothing to inherit. Same for who builds each step of the outline. Who's expected to answer this, and does the answer belong in the doc or somewhere else?

armababy (commented):
Took an early look from the maestro side, the rest of the team will do a proper pass later. Checked the maestro bits against our main and they hold up. The two things you said were worth fixing regardless are now [MET-2214](https://lokalise.atlassian.net/browse/MET-2214) (the monthly "current day" thing, reproduced, it really does drift with the compile date) and [MET-2215](https://lokalise.atlassian.net/browse/MET-2215) (the scheduler Redis vars, which turn out to be mandatory at boot, just never used).

On your "What needs review" list: for point 1 (does the profile cover both products), I compared our trigger schema with the profile table and didn't find a shape it misses, and legacy cron stays cron anyway, so nothing to add from us there. Points 2, 5, 6 and 7 are inline.

PR #1254, opened 2026-09-03
fix: preserve error cause details in Bugsnag reports and job error logs

CatchMe2:
## Summary
- A wrapped error's `cause` details (e.g. an HTTP response body captured on a lower-level `ResponseStatusError`) were silently dropped once the wrapping error was reported: Bugsnag's `reportErrorToBugsnag` only read the top-level error's `details`, and `pino.stdSerializers.err` used for job `errorJson`/log serialization discards `.cause` entirely.
- `@lokalise/error-utils`: `reportErrorToBugsnag` now walks the `cause` chain and attaches a `causeDetails` array to the Bugsnag "Context" metadata.
- `@lokalise/background-jobs-common`: swapped `pino.stdSerializers.err` → `pino.stdSerializers.errWithCause` at all `errorJson`/error-serialization call sites in both `AbstractBackgroundJobProcessorNew` and the legacy `AbstractBackgroundJobProcessor`, so the full cause chain (and its custom fields) survives into logged error JSON.
- Prompted by a Bugsnag `CALLBACK_DELIVERY_ERROR` in `polyglot-service` where the 400 response body from a translation callback was captured by `ResponseStatusError` but never surfaced in Bugsnag or Datadog once wrapped by `CallbackDeliveryError`.

## Test plan
- [x] `pnpm exec vitest run src/bugsnag.spec.ts` in `packages/app/error-utils` — 7/7 passing, including a new test asserting `causeDetails` from a nested cause chain
- [x] `pnpm exec vitest run` on `AbstractBackgroundJobProcessor.spec.ts` and `AbstractBackgroundJobProcessorNew.stalled.spec.ts` in `packages/app/background-jobs-common` — 31/31 passing
- [x] `turbo run lint` for `@lokalise/error-utils` and `@lokalise/background-jobs-common` — passing
- [x] `tsc --noEmit` for `background-jobs-common` — passing
- [x] Changeset added covering both packages

## AI Assistance Tracking

### DIFF
--- a/.changeset/preserve-error-cause-details.md
+++ b/.changeset/preserve-error-cause-details.md
@@ -1,0 +1,13 @@
+---
+'@lokalise/error-utils': patch
+'@lokalise/background-jobs-common': patch
+---
+
+Preserve `details` from wrapped errors' `cause` chain when reporting to Bugsnag and when
+serializing errors for logging in background job processors. Previously, an error's own `details`
+(e.g. an HTTP response body captured on a lower-level `ResponseStatusError`) were lost once that
+error was wrapped as the `cause` of another error, since Bugsnag reporting only read the top-level
+error's `details` and job processors serialized errors with `pino.stdSerializers.err`, which drops
+`cause` entirely. Bugsnag reports now include a `causeDetails` array with each cause's `details`,
+and job processors now use `pino.stdSerializers.errWithCause` to keep the full cause chain (and its
+custom fields) in logged `errorJson`.
--- a/.changeset/preserve-error-cause-details.md
+++ b/.changeset/preserve-error-cause-details.md
@@ -1,5 +1,5 @@
 ---
-'@lokalise/error-utils': patch
+'@lokalise/error-utils': major
 '@lokalise/background-jobs-common': patch
 ---
 
@@ -8,6 +8,16 @@ serializing errors for logging in background job processors. Previously, an erro
 (e.g. an HTTP response body captured on a lower-level `ResponseStatusError`) were lost once that
 error was wrapped as the `cause` of another error, since Bugsnag reporting only read the top-level
 error's `details` and job processors serialized errors with `pino.stdSerializers.err`, which drops
-`cause` entirely. Bugsnag reports now include a `causeDetails` array with each cause's `details`,
-and job processors now use `pino.stdSerializers.errWithCause` to keep the full cause chain (and its
-custom fields) in logged `errorJson`.
+`cause` entirely. Bugsnag reports now include an `err` metadata field built with
+`pino-std-serializers`' `errWithCause`, carrying the full cause chain along with `details`,
+`errorCode`, and any other custom error properties at every level (replacing the previous flat
+`errorDetails`/`errorCode`/`causeDetails` fields), and job processors now use
+`pino.stdSerializers.errWithCause` to keep the full cause chain (and its custom fields) in logged
+`errorJson`.
+
+`@lokalise/error-utils` now uses the native `Error.isError` and requires Node `>=24.3.0`
+(declared via `engines.node`), a breaking change for consumers on older Node versions.
+
+`AbstractBackgroundJobProcessorNew` no longer duplicates error serialization in the `context`
+passed to the error reporter (`errorJson`/`error` fields), since Bugsnag reporting now serializes
+the full error (including `cause`) itself.
--- a/.changeset/preserve-error-cause-details.md
+++ b/.changeset/preserve-error-cause-details.md
@@ -1,26 +1,23 @@
 ---
 '@lokalise/error-utils': major
-'@lokalise/background-jobs-common': patch
+'@lokalise/background-jobs-common': minor
 ---
 
-Preserve `details` from wrapped errors' `cause` chain when reporting to Bugsnag and when
-serializing errors for logging in background job processors. Previously, an error's own `details`
-(e.g. an HTTP response body captured on a lower-level `ResponseStatusError`) were lost once that
-error was wrapped as the `cause` of another error, since Bugsnag reporting only read the top-level
-error's `details` and job processors serialized errors with `pino.stdSerializers.err`, which drops
-`cause` entirely. Bugsnag reports now include an `err` metadata field built with
-`pino-std-serializers`' `errWithCause`, carrying the full cause chain along with `details`,
+Preserve `details` from wrapped errors' `cause` chain when reporting to Bugsnag. Previously, an
+error's own `details` (e.g. an HTTP response body captured on a lower-level `ResponseStatusError`)
+were lost once that error was wrapped as the `cause` of another error, since Bugsnag reporting only
+read the top-level error's `details`. Bugsnag reports now include an `err` metadata field built
+with `pino-std-serializers`' `errWithCause`, carrying the full cause chain along with `details`,
 `errorCode`, and any other custom error properties at every level (replacing the previous flat
-`errorDetails`/`errorCode`/`causeDetails` fields), and job processors now use
-`pino.stdSerializers.errWithCause` to keep the full cause chain (and its custom fields) in logged
-`errorJson`.
+`errorDetails`/`errorCode`/`causeDetails` fields).
 
-`@lokalise/error-utils` now uses the native `Error.isError` and requires Node `>=24.3.0`
-(declared via `engines.node`), a breaking change for consumers on older Node versions.
+`@lokalise/error-utils` now uses the native `Error.isError` and requires Node `>=24.3.0` (declared
+via `engines.node`), a breaking change for consumers on older Node versions.
 
-`AbstractBackgroundJobProcessor` and `AbstractBackgroundJobProcessorNew` no longer duplicate error
-serialization in the `context` passed to the error reporter (`errorJson`/`error` fields), since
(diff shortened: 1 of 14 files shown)


packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
```diff
@@ -361,7 +360,6 @@ export abstract class AbstractBackgroundJobProcessorNew<
           jobId,
           jobName: job.name,
           'x-request-id': job.data.metadata.correlationId,
```
kibertoad, line 362: Removing the `error: JSON.stringify(isError(error) ? ... : error)` field here drops the only carrier for non-`Error` throws.

The reported error a few lines above is `isError(error) ? error : new Error(...onFailed non-error exception)`. In the non-`Error` branch there is no real error object to serialize, so the justification ("Bugsnag reporting now serializes the full error itself") does not apply.

Scenario: an `onSuccess`/`onFailed` hook rejects with a non-`Error` value, e.g. `throw 'redis timeout'` or `Promise.reject({ code: 'LOCK_LOST' })`. Previously the thrown value reached the reporter as the serialized `error` context field. Now the report carries only the synthetic message (the constructor name) plus `jobId`/`jobName`/`x-request-id`, so every such event across every processor groups into one issue with no payload, and the actual value survives only in the local `resolveGlobalErrorLogObject` log line.

Suggest keeping the serialized value for the non-`Error` branch.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
```diff
@@ -435,7 +434,6 @@ export abstract class AbstractBackgroundJobProcessor<
           jobId,
           jobName: job.name,
           'x-request-id': job.data.metadata.correlationId,
```
kibertoad, line 436: Same non-`Error` information loss as in `AbstractBackgroundJobProcessorNew.ts`: the removed `error: JSON.stringify(isError(error) ? ... : error)` field was the only place the thrown value appeared when a hook rejects with a string or a plain object, since the reported error is then the synthetic `new Error(...onFailed non-error exception)`.

Separately, `errorReporter` here is an injected `ErrorReporter` (`processors/types.ts:109` and `:137`) and is never defaulted to the Bugsnag reporter. The removal therefore couples this package's observability to `@lokalise/error-utils`' implementation: a consumer wiring a log or Datadog based `ErrorReporter` that formats `context` now gets no serialized error chain at all, in either processor. That is a behaviour removal for existing consumers, so `@lokalise/background-jobs-common` looks like a `minor` rather than a `patch`.

.changeset/preserve-error-cause-details.md
```diff
@@ -0,0 +1,23 @@
...
+serializing errors for logging in background job processors. Previously, an error's own `details`
+(e.g. an HTTP response body captured on a lower-level `ResponseStatusError`) were lost once that
+error was wrapped as the `cause` of another error, since Bugsnag reporting only read the top-level
+error's `details` and job processors serialized errors with `pino.stdSerializers.err`, which drops
+`cause` entirely. Bugsnag reports now include an `err` metadata field built with
+`pino-std-serializers`' `errWithCause`, carrying the full cause chain along with `details`,
+`errorCode`, and any other custom error properties at every level (replacing the previous flat
+`errorDetails`/`errorCode`/`causeDetails` fields), and job processors now use
+`pino.stdSerializers.errWithCause` to keep the full cause chain (and its custom fields) in logged
+`errorJson`.
```
kibertoad: This sentence is stale and contradicts the third paragraph of the same changeset. It promises that "job processors now use `pino.stdSerializers.errWithCause` to keep the full cause chain (and its custom fields) in logged `errorJson`", but commits 3 and 4 deleted every `errorJson`/`error` context field from both `AbstractBackgroundJobProcessor` and `AbstractBackgroundJobProcessorNew`. The only remaining `errWithCause` call in those files is inside the `purgeJobData` failure message, which is not `errorJson` and is not a log field.

So the published release note tells a reader to expect a richer `errorJson` in their logs, and then three paragraphs later tells them the field was removed. A consumer reading only the first paragraph would conclude their log-based dashboards get more data, when they in fact get less.

Also, given that field removal is a behaviour change consumers may assert on, `@lokalise/background-jobs-common` looks like it should be `minor` rather than `patch`.

packages/app/error-utils/src/bugsnag.spec.ts
```diff
@@ -61,11 +88,7 @@ describe('bugsnag', () => {
 
       // When
       reportErrorToBugsnag({
-        error: new InternalError({
-          errorCode: 'TEST_ERROR_CODE',
-          message: 'test',
-          details: { hello: 'world' },
-        }),
+        error: new CustomErrorWithCode('test', { hello: 'world' }, 'TEST_ERROR_CODE'),
```
kibertoad, line 91: This test is still named `'internal error'` (line 84), and the one below is still named `'public non recoverable error'` (line 114), but both now construct the identical `new CustomErrorWithCode('test', { hello: 'world' }, 'TEST_ERROR_CODE')` (line 91 and line 121) and assert the identical `err: { errorCode, details }`. They are duplicates of each other, and nothing in the suite exercises node-core's real `InternalError` / `PublicNonRecoverableError` any more.

Those two classes are the main reason this code exists, and `@lokalise/node-core` is a peer dependency (`>=11.0.0`), so their shape is not pinned by this repo. The new implementation depends on `errWithCause`'s `for (const key in err)` loop picking up their `details` and `errorCode` class fields. That works today (both are own enumerable properties assigned in the constructor), but the regression guard against a node-core change is gone: if either field ever moved to a prototype getter or became non-enumerable, `details` would silently stop reaching Bugsnag and this suite would stay green.

Suggest keeping at least one test that reports an actual `InternalError`, and one for `PublicNonRecoverableError`, rather than replacing both with the local stub.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessorNew.ts
CatchMe2: Good catch — for real `Error`s this is intentional (the reporter already extracts `details`/`code`/`cause` via `errWithCause`, so re-serializing into `context` was pure duplication). But you're right that the non-Error branch loses information: the `error` passed to the reporter there is a synthetic placeholder, not the actual thrown value, so there's nothing for the reporter to extract it from.

I considered moving this into the reporter via `cause` instead of `context`, but `pino-std-serializers`' `errWithCause` only serializes `.cause` when it's itself Error-like (`typeof cause.message === 'string'`), and the native `Error(msg, { cause })` constructor makes `cause` non-enumerable — so a raw string/object cause would be silently dropped, not serialized (verified this empirically). There's no way today to make the reporter handle a non-Error throw without widening its interface (`ErrorReporter.error` is typed `Error` in `@lokalise/node-core`).

Restored `context.nonErrorValue = JSON.stringify(error)` for the non-Error branch only — dropped for real `Error`s, where it's now redundant. Fixed in 8574eaa7, with a regression test covering a non-Error throw from the failed hook.

packages/app/background-jobs-common/src/background-job-processor/processors/AbstractBackgroundJobProcessor.ts
CatchMe2: Same fix applied here (8574eaa7) — restored `context.nonErrorValue = JSON.stringify(error)` for the non-Error branch only, dropped for real `Error`s where the reporter already handles serialization itself. See the reply on the `...New.ts` thread for the reasoning (and why `cause` doesn't work as a carrier here).

On the injected-`ErrorReporter`/patch-vs-minor point: separate, still open, haven't settled it yet.

.changeset/preserve-error-cause-details.md
CatchMe2: Fixed in 640b8005 — dropped the stale "job processors now use `pino.stdSerializers.errWithCause` to keep the full cause chain ... in logged `errorJson`" sentence, since that field was removed. Also bumped `@lokalise/background-jobs-common` to `minor`, per your point on the other thread: the `context` field removal is a real behavior change for a consumer using a different `ErrorReporter` implementation, not something a `patch` should cover.

packages/app/error-utils/src/bugsnag.spec.ts
CatchMe2: Agree on the naming/duplication — fixed in 640b8005, merged the two into one test (`'error with own errorCode and details fields'`) and renamed the remaining generic-details-only test for consistency.

On using real `InternalError`/`PublicNonRecoverableError` instances instead of the local stub: intentionally not doing that. `reportErrorToBugsnag` is meant to work with any error shape that happens to carry `details`/`errorCode`-like own properties — it doesn't (and shouldn't) know about node-core's error hierarchy, which is itself being phased out in favor of `@lokalise/errors`. Pinning the test suite to either one would couple a duck-typed, deliberately-generic code path to a specific error library's internals, which is exactly the coupling we're trying to get away from — and we'd have the same "silent regression" exposure either way if that library's shape changed. The local stub is the fixture that matches what the code actually promises to support.

kibertoad (approved):

PR #1258, opened 2026-09-04
feat(frontend-http-client): add sse-fallback transport adapter

kibertoad:
`@opinionated-machine/sse-fallback` runs SSE with a transparent polling
fallback — polling as the correctness backbone, SSE as the latency layer —
but owns no HTTP. This adds the HTTP half.

`createFallbackTransport(wretch, { contract })` supplies the core's
`fetchSnapshot` / `openStream` pair on top of an API contract:

- the poll asks for the contract's JSON branch and validates the snapshot
  against its schema; a violation rejects the poll rather than reaching the
  version gate, where a wrong watermark would silently drop every later event
- the stream asks for the SSE branch, forwards `Last-Event-ID` on reconnect,
  and yields raw text chunks (comment frames included) so the core's
  stale-connection watchdog stays byte-level
- refusals resolve with their status, so `unretryableStatuses` and
  `onAuthChallenge` can act on them; only an unusable outcome rejects
- `Cache-Control: no-cache` on both channels, because a cached poll keeps
  "succeeding" while reporting stale state
- the header source is resolved fresh per request, which is what lets the one
  retry `onAuthChallenge` grants carry a refreshed token
- no deadline of its own: the core bounds every wait through its `signal`

SSE payloads are validated against the contract's event schemas.
`eventValidation: 'report'` (default) reports drift to diagnostics and still
delivers; `'drop'` withholds the frame from the core so the watermark does not
advance and a poll repairs the gap, which requires `streamMode: 'events'`
because that is where framing happens before the core sees anything.

Also adds `buildFallbackParams` (contract-typed, validated subscription
params), `SseFramer`, `FallbackSnapshotOf` / `FallbackEventsOf`, and a typed
error hierarchy. The transport seam is matched structurally, so this takes no
dependency on the core and nothing moves in lock-step; a type test keeps the
vendored declarations honest. `test/verifyWithCore.mjs` is an opt-in
acceptance gate that runs the adapter against the real core over HTTP.

## AI Assistance Tracking

### DIFF
--- a/packages/app/frontend-http-client/README.md
+++ b/packages/app/frontend-http-client/README.md
@@ -509,6 +509,7 @@ await subscription.waitFor('uploadFinished')
 | A non-2xx response **resolves** with its status | The core owns what a status means: `unretryableStatuses`, and handing a 401 to `onAuthChallenge` |
 | Only an unusable outcome rejects — network failure, schema violation, wrong representation | A rejected poll is a recoverable poll failure; a snapshot the version gate cannot trust is not |
 | Raw text chunks by default, comment frames included | Keeps the core's `staleConnectionTimeoutMs` watchdog byte-level, so a silently dead connection is caught |
+| No cross-subscription state | One transport serves any number of subscriptions; nothing a stream learns is carried into another, so a per-tenant cursor can never reach the wrong stream |
 | A snapshot it refuses has its body released | The core aborts a poll on timeout or on stopping, never one the transport rejected — an unread body (an SSE stream reached by a bad `Accept` negotiation never ends) would hold a socket per failing poll |
 | In `streamMode: 'events'`, an `id:` or `retry:` with no frame to ride on is carried into the next connect | The core's own parser reads both per chunk in `'chunks'` mode; framing here would otherwise drop a cursor advance or a server backoff hint |
 | No deadline of its own | The core bounds every wait (`pollTimeoutMs`, `connectTimeoutMs`) through the `signal` it passes |
@@ -546,14 +547,27 @@ Passing `contract` turns on the checks this package exists for:
 | `eventValidation` | Effect | Requires |
 |---|---|---|
 | `'report'` (default when the contract declares SSE events) | Reports to `diagnostics.onEventSchemaError`; the frame is still delivered | — |
-| `'drop'` | The frame never reaches the core, so the version watermark does not advance past it and the deadman poll repairs the gap | `streamMode: 'events'` |
+| `'drop'` | The frame never reaches the core **and the stream ends**, so the watermark stays below the hole and the repair snapshot lands above it | `streamMode: 'events'` |
 | `'off'` | No checking | — |
 
 `'drop'` needs `streamMode: 'events'` because that is where framing happens before the core sees
 anything. With raw chunks the core frames the stream itself, so by the time a payload could be
 withheld its version has already been gated — and the repair poll would read the snapshot as a
-duplicate and drop it. `'events'` costs byte-level liveness (comment frames are consumed by the
-framing), which is why `'chunks'` + `'report'` is the default. A cursor advance the core would
+duplicate and drop it.
+
+Ending the stream is the other half of `'drop'`, not an implementation detail: withholding one
+frame while delivering the next lets the core's watermark advance past the hole, and the repair
+poll then arrives at or below that watermark, where the reconciler treats it as a stale duplicate
+and synthesizes nothing. The withheld event would be lost for good. The cost is a reconnect per
+rejected payload — which is proportionate, since a stream emitting bodies its own contract rejects
+is broken, and polling carries the subscription meanwhile.
+
+`'events'` has two costs of its own, which is why `'chunks'` + `'report'` is the default: comment
+frames are consumed by the framing, so the stale-connection watchdog drops from byte-level to
+event-level; and an `id:` or `retry:` arriving with no `data:` reaches the core only if a later
+frame on the same connection carries it out, whereas the core's own parser reads both per chunk.
+Neither loses data — an unreported cursor replays events the version gate then dedupes, and an
+unreported `retry:` leaves the core on its own backoff. A cursor advance the core would
 otherwise miss is not part of that cost: an `id:` or `retry:` that no frame carried is inherited by
 the next connect, except past a frame `'drop'` withheld — replaying from there would skip for good
 the very event the repair poll is owed.
--- a/packages/app/frontend-http-client/README.md
+++ b/packages/app/frontend-http-client/README.md
@@ -509,6 +509,8 @@ await subscription.waitFor('uploadFinished')
 | A non-2xx response **resolves** with its status | The core owns what a status means: `unretryableStatuses`, and handing a 401 to `onAuthChallenge` |
 | Only an unusable outcome rejects — network failure, schema violation, wrong representation | A rejected poll is a recoverable poll failure; a snapshot the version gate cannot trust is not |
 | Raw text chunks by default, comment frames included | Keeps the core's `staleConnectionTimeoutMs` watchdog byte-level, so a silently dead connection is caught |
+| A snapshot it refuses has its body released | The core aborts a poll on timeout or on stopping, never one the transport rejected — an unread body (an SSE stream reached by a bad `Accept` negotiation never ends) would hold a socket per failing poll |
+| In `streamMode: 'events'`, an `id:` or `retry:` with no frame to ride on is carried into the next connect | The core's own parser reads both per chunk in `'chunks'` mode; framing here would otherwise drop a cursor advance or a server backoff hint |
 | No deadline of its own | The core bounds every wait (`pollTimeoutMs`, `connectTimeoutMs`) through the `signal` it passes |
 | Header source resolved per request | What makes the one retry `onAuthChallenge` grants actually carry a refreshed token |
 | Abort ends stream iteration quietly; a mid-stream death throws | The core aborts on purpose (stale watchdog, stop); a real failure has to surface so it backs off and polls |
@@ -551,7 +553,10 @@ Passing `contract` turns on the checks this package exists for:
 anything. With raw chunks the core frames the stream itself, so by the time a payload could be
 withheld its version has already been gated — and the repair poll would read the snapshot as a
 duplicate and drop it. `'events'` costs byte-level liveness (comment frames are consumed by the
-framing), which is why `'chunks'` + `'report'` is the default.
+framing), which is why `'chunks'` + `'report'` is the default. A cursor advance the core would
+otherwise miss is not part of that cost: an `id:` or `retry:` that no frame carried is inherited by
+the next connect, except past a frame `'drop'` withheld — replaying from there would skip for good
+the very event the repair poll is owed.
 
 An event the contract does not declare is reported to `diagnostics.onUndeclaredEvent` and never
 dropped: it usually means a newer server, not a broken one.
@@ -605,8 +610,8 @@ snapshot being a retried poll failure rather than a poisoned watermark; and
 | `FallbackSnapshotValidationError` | A snapshot body violated the contract schema; carries `issues` |
 | `FallbackUnexpectedSnapshotError` | The poll branch answered with something that cannot be a snapshot — an undeclared content type, an SSE stream (the `Accept` negotiation went wrong), a binary body, no body, or invalid JSON. Carries `contentType` and `bodyPreview` |
 | `FallbackEventValidationError` | An SSE payload failed its schema, or was not JSON. Never thrown — reported to `diagnostics.onEventSchemaError` |
-| `FallbackParamsValidationError` | `buildFallbackParams` was given params the contract's request schemas reject |
-| `FallbackUnsupportedParamError` | A query parameter is a list or a structured value; a subscription request's query is a flat string map |
+| `FallbackParamsValidationError` | `buildFallbackParams` was given params the contract's request schemas reject. Headers are checked too, but only the ones supplied — the rest come from the transport's `headers` option |
+| `FallbackUnsupportedParamError` | A query parameter is a list or a structured value; a subscription request's query is a flat string map. A schema that parses a query string into a `Date` is not one of these: the value you supplied is sent, exactly as `sendByApiContract` would |
 
 ## Credits
 
--- a/packages/app/frontend-http-client/README.md
+++ b/packages/app/frontend-http-client/README.md
@@ -394,6 +394,220 @@ sendPostWithProgress({
 abortController.abort()
 ```
 
+## Resilient live data — SSE with a polling fallback
+
+Push channels fail silently. A connection dies without an error event, a proxy kills an idle
+stream, a room rebalance drops a message — and the UI waiting for "upload finished" is stuck
+until the user reloads. [`@opinionated-machine/sse-fallback`](https://github.com/kibertoad/opinionated-machine/tree/main/packages/sse-fallback)
+solves that by making **polling the correctness backbone and SSE the latency optimization**:
+it subscribes to the SSE branch of a dual-mode contract, keeps a deadman timer, and polls the
+JSON branch of the same route whenever the stream goes quiet. A single version gate reconciles
+the two channels, so application code sees one uniform event stream.
+
+That core owns no HTTP. This package supplies the HTTP half: `createFallbackTransport` gives it
+a `fetchSnapshot` / `openStream` pair built on your configured `wretch` client and your API
+contracts, with the same schema validation, header handling and path prefixing as
+`sendByApiContract`.
+
+```
+                    ┌─────────────────────────────────────────────┐
+   your code  ◄──── │  createResilientSubscription (sse-fallback) │
+   one event        │  version gate · deadman poll · reconnect ·  │
+   stream           │  hydration · degradation · budgets          │
+                    └──────────────────────┬──────────────────────┘
+                                           │ FallbackTransport
+                    ┌──────────────────────┴──────────────────────┐
+                    │  createFallbackTransport (this package)     │
+                    │  Accept negotiation · Last-Event-ID ·       │
+                    │  Zod validation · fresh auth headers        │
+                    └─────────────────────────────────────────────┘
+```
+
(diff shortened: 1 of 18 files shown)


packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.spec.ts
```diff
@@ -0,0 +1,922 @@
...
+
+      const snapshot = await transport.fetchSnapshot(getRequest, signalOf())
+      expect(snapshot.body).toEqual({ version: 1, status: 'ok' })
+
+      const stream = await transport.openStream(
+        { path: '/uploads/u-1/events', method: 'get' },
+        signalOf(),
+      )
+      await collectChunks((stream as { chunks: AsyncIterable<string> }).chunks)
+      expect(onEventSchemaError).not.toHaveBeenCalled()
```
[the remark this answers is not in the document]
kibertoad: Fixed in 2f7d3ca. The payload is now `{ version: 1, result: 42 }`, and the test asserts one `onEventSchemaError` call carrying a `FallbackEventValidationError` for `uploadFinished` — so it fails if the frames stop being checked against `streamContract`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
```diff
@@ -0,0 +1,676 @@
...
+      contentType,
+      strictContentType,
+    )
+    if (!entry) {
+      throw new FallbackUnexpectedSnapshotError(
+        `The poll of "${path}" returned status ${response.status} with content-type "${contentType ?? '<none>'}", which "${snapshotContract.summary}" does not declare for that status.`,
+        { channel: 'poll', path, status: response.status, contentType },
+      )
+    }
+    rejectUnusableSnapshot(entry, path, response.status)
```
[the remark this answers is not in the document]
kibertoad: Fixed in f271645, with one deliberate difference from the suggestion.

The body is released via `response.body?.cancel()` in a `releaseBody` helper that swallows a cancellation failure, so the original rejection is what surfaces. `cancel()` rather than draining matters here: the case the message itself describes — a poll answered on the SSE branch — is a body that never ends, so `text()` would hang a request the caller has already given up on.

`noContent` is covered too, not skipped: `allowNoBody` only says a body is *permitted* to be absent, and a server that sends one anyway would leave it unread.

The retry framing is worth sharpening — the core's `executePoll` never aborts a poll it rejected (its `pollAbort` fires only on `pollTimeoutMs` or on stopping), so this was a socket held for the life of the page, once per failing poll, not just across a retry.

Covered by three tests under `releasing a body it walks away from`, each asserting the source stream's `cancel` ran; all three fail without the change.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/src/sse-fallback/framer.ts
```diff
@@ -0,0 +1,130 @@
...
+      this.eventName = value
+    } else if (field === 'data') {
+      this.dataLines.push(value)
+    } else if (field === 'id') {
+      // Per spec an id containing NUL is ignored outright, rather than
+      // truncated — a truncated cursor would replay from the wrong point.
+      if (!value.includes(NULL_CHARACTER)) {
+        this.frameId = value
+        this.cursor = value
+      }
```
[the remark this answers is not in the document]
kibertoad: Both fixed in 2f7d3ca — these were the sharpest findings of the four.

**Empty `event:`** — the event-type buffer is only allowed to override `message` when it is non-empty, so `dispatch()` now omits the name when it is `''`. Reporting `''` as a name of its own sent the frame looking for a schema by that name and counted it as an undeclared event.

**`id:` in an unterminated frame** — split into the spec's two pieces: `idBuffer` (the last event ID *buffer*, set by any `id:` line and never reset between frames) and `cursor` (the last event ID *string*, which now takes the buffer's value as the first step of `dispatch()`). So an `id:`-only frame still advances the cursor — that step runs before the empty-data-buffer abort — while an `id:` in a frame the connection cut short does not.

That second one had teeth beyond the framer: f271645 added a stream carry that inherits the framer's trailing cursor into the next connect in `streamMode: 'events'`, so an unterminated `id:` would have made the reconnect resume past an event that was never delivered — the exact "reconnect past an undelivered event" risk in your walkthrough. Tests added for both cases (`reads an empty event name as no name…`, `holds the cursor at an id whose frame the stream cut short`).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/src/sse-fallback/params.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    })
+
+    expect(params.queryParams).toEqual({ verbose: false })
+  })
+
+  it('passes a static header map and a request body through', () => {
+    const params = buildFallbackParams(startImportContract, {
+      body: { fileId: 'f-1' },
+      headers: { 'x-tenant': 'acme' },
+    })
```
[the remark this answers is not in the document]
kibertoad: The core of this is fixed in f271645 — `headers` was indeed the one part never checked — but two details are deliberately not what you proposed, because a request's headers come from two layers.

**Not demanded in full.** The transport's own `headers` option is resolved fresh for every poll and every reconnect, which is what lets `onAuthChallenge` carry a refreshed token, and `buildFallbackParams`' own docs send rotating credentials there. `requestHeaderSchema` describes the *request*, not this layer's contribution to it — so demanding all of it here would reject the setup the docs recommend for a contract that declares `authorization`. The schema is therefore `.partial()`ed first: what is supplied is validated, what is absent is assumed to come from the transport. `FallbackContractParams['headers']` is optional/partial for the same reason, so the type and the runtime agree.

**Validated, not substituted.** The caller's map is what gets sent rather than Zod's output: header values are strings on the wire either way, and swapping in the parsed object would let unknown-key pruning silently drop a header the contract does not declare — a worse failure than the one being fixed.

Tests added under `validation > headers`: an invalid supplied value raises `FallbackParamsValidationError` with `part: 'headers'`; a header the transport supplies is not demanded; an undeclared header survives.

---
_Generated by [Claude Code](https://claude.ai/code)_

Drodevbar (approved):

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
```diff
@@ -0,0 +1,806 @@
...
+    const framer = new SseFramer({ lastEventId: stream.resumedFrom, retry: stream.retry })
+    // Both start where the connection opened: with no frames yet, that is
+    // exactly where the core's cursor is.
+    let cursorAtLastFrame = stream.resumedFrom
+    let cursorAtLastDelivery = stream.resumedFrom
+    try {
+      for await (const chunk of chunks) {
+        for (const frame of framer.push(chunk)) {
+          cursorAtLastFrame = frame.lastEventId
+          if (validateFrame(frame, path) && eventValidation === 'drop') continue
```
casamitjana: Dropping only this frame does not preserve the gap. The loop continues and can yield version N+1, which advances the core watermark past the invalid version. A later repair poll at N or N+1 is then treated as a duplicate, so `snapshotToEvents` never reconstructs the dropped event. Once validation fails, stop delivering frames from this connection until reconciliation has run, or remove the claim that `drop` guarantees poll repair.

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
```diff
@@ -0,0 +1,806 @@
...
+  }
+
+  /**
+   * Residue of the last framed connection — see {@link StreamCarry}. One slot
+   * rather than a map: reconnects for a stream are consecutive, so a slot
+   * covers them, while a transport shared across several streams simply finds
+   * nothing to inherit instead of inheriting another stream's cursor. It is
+   * only ever written in `streamMode: 'events'`, and read exactly once.
+   */
+  let carry: StreamCarry | undefined
```
casamitjana: This transport-wide slot is shared by every subscription. Two streams with the same method/path/query—but different request headers, bodies, or simply separate subscription lifecycles—can consume each other’s trailing cursor. A tenant-scoped stream can then reconnect with another tenant’s `Last-Event-ID` and skip events it never received. This state needs subscription-local identity; the request-derived key and a single shared slot are not safe.

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
```diff
@@ -0,0 +1,806 @@
...
+      carry = {
+        key: stream.key,
+        deliveredCursor: cursorAtLastDelivery,
+        // Only an advance that happened *after* the last frame is inheritable.
+        // A cursor sitting past a frame the core never received (a dropped
+        // payload) is one the next connect must not skip over — that frame is
+        // exactly what the repair poll, or a replay, still owes us.
+        trailingCursor:
+          cursorAtLastDelivery === cursorAtLastFrame ? framer.lastEventId : cursorAtLastDelivery,
+        retry: framer.pendingRetry,
```
casamitjana: A standalone `retry:` is applied one reconnect late. When this connection ends, the core has seen no frame carrying the value, so it chooses the next delay from its existing backoff. The value is only seeded into the new framer and reported if that next connection later dispatches data. SSE applies `retry:` as soon as the field is parsed, so this representation cannot preserve the directive across a body that closes immediately.

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
kibertoad: You're right, and the carry is removed in 82aef49 rather than repaired.

I went looking for a safe key first. Adding headers and body closes the tenant case you name, but not the one after it: two subscriptions to the *same* stream — same tenant, same path, same headers, separate lifecycles — are indistinguishable at this seam, and handing one the other's trailing cursor skips events it never received just the same. There is no request-derived key that identifies a subscription, and the seam hands the transport no subscription identity to use instead. Scoping it by documenting "one transport per subscription" would have been worse: the transport is built from a wretch client plus a contract precisely so it can serve many, and a rule like that fails silently when violated.

So the state is gone. `frameChunks` still seeds the framer's Last-Event-ID, but from `opts.lastEventId` — the cursor the core passes for that connection — so nothing survives a call now. I added a row to the transport-guarantees table making "no cross-subscription state" an explicit property rather than an accident.

What that costs is what the carry was buying: in `streamMode: 'events'`, an `id:` or `retry:` arriving with no `data:` behind it reaches the core only if a later frame on the same connection carries it out. That is now documented on `FallbackStreamMode` and in the README as a cost of the mode, next to the byte-level liveness it already gives up. Neither loses data — an unreported cursor replays events the version gate then dedupes — and `chunks`, the default, has neither problem because the core's own parser reads both per chunk.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
kibertoad: Agreed, and it's gone with the carry in 82aef49 — `SseFramerOptions.retry` and `SseFramer.pendingRetry` are removed too, so nothing pretends to preserve the directive.

Your diagnosis is the sharper one: it isn't that the value arrives late, it's that it can't arrive at all through this seam. A `retry:` only reaches the core riding on a frame, and the connection it applies to has already ended by then. A body that closes immediately after `retry:` — the load-shedding case where the directive actually matters — can never deliver it.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts
kibertoad: Confirmed against the core and fixed in 82aef49 by taking the first branch — `drop` now ends the stream.

I traced it to be sure the repair really is unreachable, and it is. `gateAndDeliver` on version N+1 with the watermark at N-1 does register a gap, but it still delivers the frame and advances the watermark to N+1. The repair poll's snapshot then hits `watermarkComparison <= 0` in `handleSnapshot` and lands in `handleStaleSnapshot`, which sets `stale` and returns without touching `snapshotToEvents`. The state layer is repaired via `repairState` — which is presumably why this looked fine — but the withheld *event* is never reconstructed. So the doc claim was false in exactly the way you describe.

Ending the stream on the first rejected payload leaves the watermark below the hole, which is what makes the repair snapshot land above it and synthesize the events it covers. The core reads the end as a disconnect and reconnects from the cursor it still holds, and the deadman poll fills the gap meanwhile.

The reconnect-per-rejected-payload cost seems right rather than merely acceptable: a stream emitting bodies its own contract rejects is broken, and `drop` is the mode you pick when you would rather poll than trust it. Documented on `FallbackEventValidationMode` and in the README as the load-bearing half of `drop` it turns out to be, not an implementation detail.

The existing test was asserting the old behaviour — it fed one bad frame and one good one and expected the good one through. It now feeds valid/invalid/valid and asserts only the frame before the rejection is delivered; it fails against `continue`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/README.md
```diff
@@ -394,6 +394,244 @@ sendPostWithProgress({
...
+| Only an unusable outcome rejects — network failure, schema violation, wrong representation | A rejected poll is a recoverable poll failure; a snapshot the version gate cannot trust is not |
+| Raw text chunks by default, comment frames included | Keeps the core's `staleConnectionTimeoutMs` watchdog byte-level, so a silently dead connection is caught |
+| No cross-subscription state | One transport serves any number of subscriptions; nothing a stream learns is carried into another, so a per-tenant cursor can never reach the wrong stream |
+| A snapshot it refuses has its body released | The core aborts a poll on timeout or on stopping, never one the transport rejected — an unread body (an SSE stream reached by a bad `Accept` negotiation never ends) would hold a socket per failing poll |
+| In `streamMode: 'events'`, an `id:` or `retry:` with no frame to ride on is carried into the next connect | The core's own parser reads both per chunk in `'chunks'` mode; framing here would otherwise drop a cursor advance or a server backoff hint |
```
casamitjana: This still describes the carry removed in `82aef497`. In `streamMode: events`, a standalone `id:` or `retry:` is no longer carried into the next connection; the limitation is described correctly below. Remove or rewrite this row. The later paragraph around the event-mode costs also still says these values are inherited across reconnects and needs the same cleanup.

packages/app/frontend-http-client/test/verifyWithCore.mjs
```diff
@@ -0,0 +1,359 @@
...
+  })
+
+  await new Promise((resolve) => setTimeout(resolve, 300))
+  upstream.push('id: 2\nevent: progress\ndata: {"version":2,"percent":"half"}\n\n')
+  upstream.push(`id: 3\nevent: progress\ndata: ${JSON.stringify({ version: 3, percent: 75 })}\n\n`)
+
+  await new Promise((resolve) => setTimeout(resolve, 300))
+  assert.equal(schemaErrors.length, 1, `expected one schema error, saw ${schemaErrors.length}`)
+  assert.equal(schemaErrors[0].event, 'progress')
+  assert.equal(subscription.getState()?.percent, 75, 'the valid delta must have been applied')
```
casamitjana: This acceptance case still asserts the behavior removed in `82aef497`. With `eventValidation: drop`, `frameChunks` returns on the invalid version-2 frame, so version 3 from the same connection must not be applied. This script should instead prove the repair path: return a newer snapshot from the poll, wait for that state, and assert that the post-error SSE frame was not delivered. As written, the opt-in acceptance gate should fail against the new implementation.

packages/app/frontend-http-client/README.md
kibertoad: Right, and worse than stale — that row contradicted the "no cross-subscription state" row two lines above it. Fixed in aa537b8: the row is gone, and the stream-mode paragraph's trailing sentences about the values being inherited across reconnects went with it, leaving the two costs that are real.

One more you didn't flag, same cause: the acceptance-gate summary further down still said the gate asserts `'drop'` "withholding a bad payload while a valid delta still applies" — the removed behaviour again. It now matches what the gate actually does.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/frontend-http-client/test/verifyWithCore.mjs
kibertoad: Correct — and confirmed by running it rather than by reading it. I built the core from source, linked it in, and reproduced the failure on `51d9c91` before changing anything:

```
AssertionError [ERR_ASSERTION]: the valid delta must have been applied
undefined !== 75
    at test/verifyWithCore.mjs:350
```

That gate was the one thing that should have caught 82aef497 changing behaviour under it, and I never ran it — I only ran the unit suite, which I had updated. My mistake.

Rewritten along the lines you describe, in aa537b8. It now hydrates at version 1, pushes the invalid version 2 and the valid version 3 down one connection, and asserts version 3 is **absent** and the watermark still at 1 — then flips the poll's answer to a version 4 snapshot and waits for the state to reach it. That last step is the part worth having: it only passes because the watermark stayed below the gap, so it is exactly what the old behaviour turned into a stale duplicate. `'drop'` ending the stream is now demonstrated end to end rather than only argued for in a commit message.

The reconnect after the stream ends falls through to the JSON rule, whose content type the core counts as a connect failure — which is what degrades it onto the poll, so the repair path is the real one and not a shortcut.

All seven cases pass against the real core:

```
  ✓ SSE delivery + subscribe-first hydration
  ✓ deadman poll repairs an event the live stream never sent
  ✓ Last-Event-ID resumes the stream after a drop
  ✓ onAuthChallenge recovery picks up a freshly resolved token
  ✓ poll-only mode never opens a stream
  ✓ a schema-invalid snapshot is a retried poll failure, reported to diagnostics
  ✓ eventValidation 'drop' holds the watermark below the gap so a poll repairs it
```

---
_Generated by [Claude Code](https://claude.ai/code)_

casamitjana (approved):
Thanks for addressing the review findings and tightening the acceptance coverage.

PR #1276, opened 2026-09-10
feat!: drop the legacy contract definition API

CatchMe2:
Removes the deprecated contract API. `defineApiContract` is the only way to define a contract now, and the `*ApiContract` functions are the only way to consume one. The short legacy names (`sendByContract`, `buildFastifyRoute`, `injectByContract`, `MockttpHelper`, `MswHelper`) are gone rather than aliased, so a stale call site fails with a missing export instead of silently changing behavior.

Major bump for api-contracts, backend-http-client, frontend-http-client, fastify-api-contracts and universal-testing-utils. Consumers need `@lokalise/api-contracts@>=9.0.0`.

## Removed

- api-contracts: `buildGetRoute`, `buildPayloadRoute`, `buildDeleteRoute`, `buildRestContract`, `buildContract`, `buildSseContract`, `mapRouteToPath`, `describeContract` and the legacy definition types.
- backend-http-client and frontend-http-client: `sendByContract`, `sendByGetRoute`, `sendByPayloadRoute`, `sendByDeleteRoute`, the streamed variants, and `connectSseByContract`.
- fastify-api-contracts: `buildFastifyRoute`, `buildFastifyRouteHandler`, `buildFastifyNoPayloadRoute`, `buildFastifyPayloadRoute`, `injectByContract`, `injectGet` through `injectDelete`.
- universal-testing-utils: `MockttpHelper`, `MswHelper` and their param types. Use `ApiContractMockttpHelper` and `ApiContractMswHelper`.

## Also

- `pathResolver` must return a path starting with `/`. Literals and template strings infer fine. A helper that returns plain `string` needs to return `` `/${string}` ``.
- api-contracts gets framework-agnostic server types (`InferServerRequest`, `InferServerResponse` and friends). fastify-api-contracts builds its handler types on them, names unchanged.
- Flattened the `src/api-contracts/` folders in fastify-api-contracts and universal-testing-utils.

Migrate our services to `defineApiContract` before this is released.

## Test plan

- [x] lint, tests at 100% coverage and build pass for all five packages
- [x] repo-wide lint has three failures that also fail on `main` locally (prisma-utils, fastify-bullboard-plugin, translation-quality-checks), unrelated

## AI Assistance Tracking


packages/app/fastify-api-contracts/src/buildFastifyApiRoute.ts
```diff
@@ -20,6 +20,13 @@ import {
   validateApiResponseHeaders,
 } from './sseUtils.ts'
 
+declare module 'fastify' {
+  interface FastifyContextConfig {
+    /** The contract the route was built from. */
+    apiContract: ApiContract
```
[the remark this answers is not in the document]
CatchMe2: It is what it was implemented like before, I wouldn't change it for now.

packages/app/frontend-http-client/src/client.ts
```diff
@@ -117,7 +100,7 @@ async function sendResourceChange<
     return Promise.reject(queryParams.error)
   }
 
-  const resolvedHeaders = await resolveHeaders(params.headers as Record<string, string>)
+  const resolvedHeaders = (params.headers ?? {}) as Record<string, string>
```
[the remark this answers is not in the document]
CatchMe2: It was working fine, I wouldn't change it now.

packages/app/backend-http-client/MIGRATION.md
```diff
@@ -9,9 +9,10 @@ This release introduces streaming response support for GET requests, allowing me
 #### Added Methods
 
 - `sendGetWithStreamedResponse()` - Direct path-based GET requests with streamed responses
-- `sendByGetRouteWithStreamedResponse()` - API contract-based GET requests with streamed responses
 
-These methods use `sendWithRetryReturnStream` under the hood to return a `Readable` stream instead of consuming the entire response body.
+This method uses `sendWithRetryReturnStream` under the hood to return a `Readable` stream instead of consuming the entire response body.
+
+For contract-based requests, declare the response as `blobResponse()` in the contract and call `sendByContract` (alias `sendByApiContract`); the returned `BlobResponseHandle` exposes `stream()` for incremental consumption.
```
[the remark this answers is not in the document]
CatchMe2: Done

kibertoad (commented):
Recall-oriented review of the full branch. I built, linted, typechecked and ran the test suites for all five packages in a scratch worktree: everything passes (1620 tests, no lint or type errors), so what follows comes from reading the diff rather than from a red build. Finding 2 is confirmed by compiling a scratch file against the branch.

14 findings, most severe first. 13 are inline. The remaining one has no anchor in the diff and is recorded here.

### 14. Test-plan claim: coverage is not 100% for all five packages

`packages/app/universal-testing-utils/vitest.config.ts:14`

The test plan says "lint, tests at 100% coverage and build pass for all five packages". Running `test:ci` on this branch gives:

- `universal-testing-utils`: 94.5% statements, 88.18% branches, 95.09% lines. Uncovered: `ApiContractMockttpHelper.ts` line 73 and lines 153-161, `ApiContractMswHelper.ts` lines 153-160, `types.ts` line 99.
- `api-contracts`: 96.29% branches.

Nothing regressed: `main` sits at 93.37% statements for `universal-testing-utils`, and the thresholds in `vitest.config.ts` (lines 81, branches 79) are unchanged. The claim is inaccurate, and it hides the fact that the retained mock-helper error paths have no test.

packages/app/fastify-api-contracts/src/injectByApiContract.ts
```diff
@@ -45,13 +44,50 @@ export function injectByApiContract<const TApiContract extends ApiContract>(
...
   app: AnyFastifyInstance,
   apiContract: ApiContract,
   // biome-ignore lint/suspicious/noExplicitAny: params shape depends on the contract
   params: any,
 ): Promise<LightMyRequestResponse> {
   const path = buildRequestPath(apiContract.pathResolver(params.pathParams), params.pathPrefix)
+  const headers = typeof params.headers === 'function' ? await params.headers() : params.headers
 
-  return dispatchInjectByMethod(app, apiContract.method, path, params)
+  switch (apiContract.method) {
```
kibertoad: Inlining `dispatchInjectByMethod` dropped its `default:` arm, which threw ``Unsupported HTTP method: ${method}``. An unrecognised method now falls off the end of the switch and the function resolves to `undefined`.

TypeScript proves the switch exhaustive over `ApiContract['method']`, so nothing fails at compile time. At runtime the gap opens for a contract cast with `as ApiContract`, a JS test helper, or a `head`/`options` method added to `ApiContract` later without touching this switch. The caller's next line, `response.statusCode`, then throws `TypeError: Cannot read properties of undefined`, pointing at the assertion instead of at the contract.

Restore the `default` throw.

packages/app/fastify-api-contracts/src/buildFastifyApiRoute.ts
kibertoad: `apiContract` is declared required, so this augmentation makes it required on every Fastify route config in any program that loads this package's types. Compiled against this branch:

```
app.get('/x', { config: { rateLimit: { max: 5 } } }, handler)
// TS2322: Property 'apiContract' is missing in type '{ rateLimit: { max: number; }; }'
//         but required in type 'FastifyContextConfig'.
```

The other direction is a runtime hazard: a hook doing `req.routeOptions.config.apiContract.method` type-checks, then throws on any route not built by `buildFastifyApiRoute`, because the property really is absent there.

The required declaration predates this PR (it lived in the now-deleted `fastifyApiContracts.ts`), but the block is being moved and retyped here, so this is the moment to write `apiContract?: ApiContract`.

packages/app/frontend-http-client/src/client.ts
kibertoad: `resolveHeaders` is gone, so a function passed as `headers` is no longer invoked. The same substitution lands at line 155 (`sendGet`) and line 347 (`sendDelete`).

`wretch.headers(fn)` merges its argument with `Object.assign`, and a function has no own enumerable properties, so the request goes out with no headers at all instead of the resolved ones. A JS caller, or a TS call site using `as any`, loses its Authorization header without an error and gets a 401.

The public param types (`FreeHeadersParams.headers?: Record<string, string>`) do reject a function, so typed callers are safe. The changeset still does not list this removal; either add it there or keep the resolution.

packages/app/backend-http-client/MIGRATION.md
```diff
@@ -214,17 +204,12 @@ if (result.result) {
...
-  {
-    requestLabel: 'Download with retry',
-    retryConfig: {
-      maxAttempts: 3,
-      statusCodesToRetry: [500, 502, 503],
-      retryOnTimeout: true,
-    },
-  }
-)
+const result = await sendGetWithStreamedResponse(client, '/items/123', {
```
kibertoad, line 209: The rewritten retry example passes `maxAttempts`, `statusCodesToRetry` and `retryOnTimeout`. `RetryConfig` in `src/api-contract/retry.ts` declares `maxRetries`, `statusCodes`, `delay`, `maxDelay` and `maxJitter`, and `src/client/types.ts:32` types `retryConfig?: RetryConfig | true` against it.

Copying this block produces an excess-property error on `maxAttempts`. Cast past it and the request retries with the defaults (2 retries, `[408, 425, 429, 500, 502, 503, 504]`) rather than the configured policy. The spec added in this PR uses the correct names (`statusCodes`, `maxRetries`, `delay`, `maxJitter`), so the doc is the only thing left stale, and these lines were rewritten here.

packages/app/fastify-api-contracts/src/apiHandlerTypes.ts
```diff
@@ -0,0 +1,152 @@
...
+/** Union of the response content-types the contract's success entries declare (error responses excluded). */
+export type InferContractResponseContentTypes<TContract extends ApiContract> =
+  InferServerResponseContentTypes<TContract>
+
+/**
+ * Context passed to every `ApiContract` handler as the third argument.
+ *
+ * `expectedContentType` is the response content-type the client prefers, negotiated from the
+ * request's `Accept` header (with `q=` quality values and wildcards) against the content-types
+ * the contract's success entries declare (`2xx` codes, `'2xx'`, `'default'`) — error responses
```
kibertoad: 18 added lines in this PR contain U+2014, which `~/.claude/CLAUDE.md` bans outright: "Never emit [...] as punctuation. Never. [...] One in a diff is a defect, the same as a failing test."

```
git diff origin/main...HEAD -- . ':(exclude)pnpm-lock.yaml' | grep -cP '^\+.*\xe2\x80[\x93\x94]'
# 18
```

By file: 9 in this one (lines 44, 46, 47, 86, 101, 103, 132, 146, 147), 7 in `packages/app/api-contracts/README.md`, 1 in `packages/app/fastify-api-contracts/README.md`, 1 in `packages/app/api-contracts/src/commonTypes.ts:27`. The rule covers text moved as much as text written fresh, so the line copied into `commonTypes.ts` counts.

Replace each with a colon, a comma, parentheses, or a full stop.

packages/app/backend-http-client/README.md
```diff
@@ -13,9 +13,7 @@ The library provides methods to implement the client side of HTTP protocols. Pub
     keepAliveMaxTimeout: 300_000,
     keepAliveTimeout: 4000,
     ```
-- `sendByApiContract()`, the recommended method for making type-safe HTTP requests from an `ApiContract` definition (created with `defineApiContract`);
-- `sendByContract()` _(deprecated — use `sendByApiContract` instead)_;
-- `sendByContractWithStreamedResponse()` _(deprecated — use `sendByApiContract` instead)_;
+- `sendByContract()` (also exported as `sendByApiContract()`), the recommended method for making type-safe HTTP requests from an `ApiContract` definition (created with `defineApiContract`);
```
kibertoad: This bullet presents the alias as the primary name. The code reads the other way round: `src/api-contract/sendByApiContract.ts:340` is `export const sendByContract = sendByApiContract`, and line 86 of this same README says "`sendByApiContract` is the fully type-safe way [...] `sendByContract` is the same function under a shorter name".

`MIGRATION.md:15` repeats the inverted form ("call `sendByContract` (alias `sendByApiContract`)"). Pick one canonical name and use it in both files.

packages/app/backend-http-client/src/client/httpClient.ts
```diff
@@ -213,48 +203,6 @@ async function sendResourceChange<
...
-    result,
-    options.throwOnError ?? (DEFAULT_OPTIONS.throwOnError as DoThrowOnError),
-    options.validateResponse ?? DEFAULT_OPTIONS.validateResponse,
-    options.responseSchema,
-    options.requestLabel,
-    options.isEmptyResponseExpected,
-  )
-}
-
 export function sendPost<
```
kibertoad, line 206: Anchored here because the line I mean, `httpClient.ts:37`, sits outside the diff.

`sendNonPayload` was deleted at this spot, but `sendGet` (line 37) and `sendDelete` (line 119) still carry two verbatim copies of its body. The `executeRequest` plus `resolveResult` block is identical in both apart from `method: 'GET'` vs `'DELETE'` and the `isEmptyResponseExpected` default (`false` vs `true`).

A later change to the `REQUEST_ID_HEADER` propagation or to the `Object.hasOwn(options, 'timeout')` handling now has to be made twice, and making it once is invisible in review. `sendNonPayload` only looked dead because the legacy senders went with it: keep it, and have `sendGet` and `sendDelete` delegate with their method and their default.

packages/app/fastify-api-contracts/src/injectByApiContract.ts
```diff
@@ -45,13 +44,50 @@ export function injectByApiContract<const TApiContract extends ApiContract>(
...
 // Implementation
-export function injectByApiContract(
+export async function injectByApiContract(
   app: AnyFastifyInstance,
   apiContract: ApiContract,
   // biome-ignore lint/suspicious/noExplicitAny: params shape depends on the contract
   params: any,
 ): Promise<LightMyRequestResponse> {
   const path = buildRequestPath(apiContract.pathResolver(params.pathParams), params.pathPrefix)
+  const headers = typeof params.headers === 'function' ? await params.headers() : params.headers
```
kibertoad: This is the third hand-rolled copy of header-factory resolution in the repo. The other two are `resolveRequestHeaders` in `packages/app/frontend-http-client/src/api-contract/sendByApiContract.ts:92` and `resolveHeaderSource` in `packages/app/frontend-http-client/src/sse-fallback/createFallbackTransport.ts:186`. This PR deletes a fourth, `resolveHeaders` in `frontend-http-client/src/client.ts`.

`HeadersParam<T>` already lives in `@lokalise/api-contracts` (`clientTypes.ts`). Exporting a `resolveHeadersParam` next to it gives all three call sites one implementation, and one place to decide what happens when the factory rejects.

packages/app/fastify-api-contracts/src/sseTypes.ts
```diff
@@ -116,15 +101,8 @@ export type SSESession<Events extends SSEEventSchemas = SSEEventSchemas, Context
...
-export type SSESelection = {
-  statusCode: number | string
-  contentType: string
-  events: SSEEventSchemas
-}
+/** One SSE representation a contract declares: status key, media type and event schemas. */
+export type SSESelection = ServerSseSelection
 
 /** True when `TUnion` has two or more members. */
 type IsUnion<TUnion, TFull = TUnion> = TUnion extends unknown
```
kibertoad: This private `IsUnion` duplicates the one in `packages/app/api-contracts/src/typeUtils.ts`, which the PR now routes `serverTypes.ts:8` through.

The copy exists because `packages/app/api-contracts/src/index.ts` has no `export * from './typeUtils.ts'`, so `IsUnion`, `Prettify`, `ValueOf` and `KeysOfUnion` are unreachable from here. Adding `typeUtils.ts` to that barrel lets this declaration go and stops two copies of a distributive conditional drifting apart.

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
```diff
@@ -336,3 +336,5 @@ export async function sendByApiContract<
   // biome-ignore lint/suspicious/noExplicitAny: return type is inferred from TIsStreaming and TCaptureAsError
   return { result: parsedResponse } as any
 }
+
+export const sendByContract = sendByApiContract
```
kibertoad: This alias, together with `buildFastifyRoute`, `injectByContract`, `MockttpHelper` and `MswHelper`, rebinds a removed name to an incompatible implementation. The PR description gives the reason: "so we can drop the `Api` infix later".

In a release that is already major across five packages, that trade runs the wrong way. It doubles the exported surface now to save a rename later, and it sets a same-name, different-semantics trap. Old `sendByContract` returned `{ result: <body> }` for a legacy route definition; this one returns `{ result: { statusCode, headers, body } }` for an `ApiContract`. Old `buildFastifyRoute` took a handler returning the body; the new one requires `{ status, body }`. Wherever such a call sits behind `any` (mock factories, generic route registrars, JS tests) the swap is silent.

Either do the rename in this PR, or drop the old names so consumers get a missing-export error instead of a rebind.

packages/app/frontend-http-client/src/client.ts
```diff
@@ -77,7 +60,7 @@ function handleBodyParseError<RequestBodySchema extends z.ZodSchema>(
   return Promise.reject(bodyParseResult.error)
 }
 
-async function sendResourceChange<
+function sendResourceChange<
```
kibertoad: `async` was dropped here, and from `sendGet` (line 128) and `sendDelete` (line 316). The declared return type is still `Promise<...>`, but a synchronous throw now escapes on the caller's stack instead of surfacing as a rejection.

The parse-failure paths still use `Promise.reject`, so those are fine. The exposed path is the wretch builder chain (`wretch.headers(...)[method](body, path)`), which now runs synchronously: if a middleware or addon throws while building the request, `sendPost(w, params).catch(handle)` never reaches `.catch`.

`async` only became removable because the `await resolveHeaders(...)` went away. Keeping it costs nothing and preserves the contract the signature advertises.

packages/app/api-contracts/src/serverTypes.ts
```diff
@@ -0,0 +1,237 @@
...
+  [K in keyof TApiContract['responsesByStatusCode'] & HttpStatusCodeRange]: ExpandStatusRangeKey<K>
+}[keyof TApiContract['responsesByStatusCode'] & HttpStatusCodeRange]
+
+/**
+ * Maps a `responsesByStatusCode` key to the statuses a handler may return for it, mirroring
+ * the runtime lookup precedence (exact → range → `'default'`): a concrete key stays as-is; a
+ * range key expands to its status class minus the exactly-declared codes; `'default'` expands
+ * to every status not covered by an exact or range key.
+ */
+export type InferServerStatusesForKey<
```
kibertoad, line 94: `InferServerStatusesForKey` (this line), `ServerSseSelection` (line 207) and `SseMessage` (line 29) are new public exports with no assertion in `serverTypes.spec.ts`, which covers the other five.

`InferServerStatusesForKey` encodes the exact, range, `'default'` precedence and is exercised only indirectly through `InferServerResponse`, so a regression in the `Exclude<HttpStatusCode, ExactStatusCodes<...> | RangeStatusCodes<...>>` branch for a contract declaring only `'default'` would pass. `ServerSseSelection.statusCode` is typed `number | string`, wider than the concrete union `InferServerSseSelections` produces, and nothing asserts the two stay compatible.

packages/app/api-contracts/src/defineApiContract.spec.ts
```diff
@@ -48,7 +48,7 @@ describe('defineApiContract', () => {
         visibility: 'public',
         summary: 'Original',
         method: 'get',
-        pathResolver: () => '/users',
+        pathResolver: () => '/users' as const,
```
kibertoad, line 51: This `as const` is the tell that the new `` `/${string}` `` constraint does not infer for a contract config assembled before the `defineApiContract` call: with no contextual type, the arrow's return widens to `string`.

The changeset says "Inline literals and template strings infer correctly". That holds for resolvers written inline in the call. A consumer who builds the config as a separate object, a factory, or a shared const hits `Type '() => string' is not assignable to type 'RoutePathResolver<...>'` for a plain literal.

Either widen the changeset and README note to name that case, or keep `pathResolver` returning `string` and rely on the runtime `ensureStartsWithSlash` in `buildRequestPath`, which is still in place.

CatchMe2: I don't see a usecase with having shared pathResolver, especially when it arg depends on pathParamsSchema. I'd leave it as is.

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
CatchMe2: @kibertoad @CarlosGamero WDYT about this one? Should we drop the old name? Actually, I'm not sure about this rename, I kinda like `sendByApiContract` over `sendByContract`. It is explicit and follows the api-contract naming (defineApiContract return ApiContract type).

kibertoad: fair point, ApiContract is generally clearer, I mostly hate the ApiController part :D

CarlosGamero: Personally I also prefer sendByApiContract

packages/app/fastify-api-contracts/src/injectByApiContract.ts
CatchMe2: Restored in e18c14ab. The default arm is back, now as an exhaustiveness check: the method is assigned to `never`, so a new `ApiContract` method fails to compile here instead of falling through. A test drives a `head` contract through a cast and asserts the throw.

packages/app/fastify-api-contracts/src/buildFastifyApiRoute.ts
CatchMe2: Done in f04e403d, `apiContract?: ApiContract`. README and the builder doc comment now say it is absent on routes not built by `buildFastifyApiRoute`, and the changeset notes that hooks must narrow first.

packages/app/frontend-http-client/src/client.ts
CatchMe2: Restored in 08f6d215. `resolveHeaders` is back at all three call sites. I also widened `FreeHeadersParams.headers` to `HeadersSource` so the factory is typed, like `sendByApiContract`, and added a test that sends a GET with an async factory and matches the resolved header on the server.

packages/app/backend-http-client/MIGRATION.md
CatchMe2: Fixed in 5aa59d7e: `maxRetries`, `statusCodes`, `delay: exponentialDelay({ baseDelayMs: 100 })`, `retryOnTimeout`, plus the import line. The earlier `maxAttempts` blocks in the v10 section describe that version and are left as is.

packages/app/fastify-api-contracts/src/apiHandlerTypes.ts
CatchMe2: Done in 57d3455e. `git diff origin/main...HEAD` now has zero added lines with U+2013 or U+2014. Colons, semicolons, parentheses or full stops depending on the sentence; the arrow chain in the README is spelled out.

packages/app/fastify-api-contracts/src/sseTypes.ts
CatchMe2: Done in 2831ea17. `typeUtils.ts` is exported from the api-contracts barrel and this copy is gone. Same for the `IsUnion` in universal-testing-utils and the `Prettify` in frontend-http-client. The testing-utils one differed on `never` input, so I checked `MockImplementationParams` under both definitions for SSE-only and blob-only statuses; same result.

packages/app/frontend-http-client/src/client.ts
CatchMe2: Back in 08f6d215. Restoring `resolveHeaders` brought the `await` back, so `sendResourceChange`, `sendGet` and `sendDelete` are `async` again and a builder-chain throw surfaces as a rejection.

packages/app/backend-http-client/src/api-contract/sendByApiContract.ts
CatchMe2: Dropped in 32689dc4. `sendByContract`, `buildFastifyRoute`, `injectByContract`, `InjectByContractParams`, `MockttpHelper` and `MswHelper` are gone; only the `*ApiContract` names remain. READMEs, changeset and PR description updated.

packages/app/backend-http-client/src/client/httpClient.ts
CatchMe2: Restored in 620b4d8f. `sendNonPayload` is back with the original body; `sendGet` and `sendDelete` delegate with their method and their `isEmptyResponseExpected` default.

packages/app/fastify-api-contracts/src/injectByApiContract.ts
CatchMe2: Done in 620b4d8f. api-contracts exports `resolveHeadersParam` next to `HeadersParam` (both now live in `headersParam.ts`), with its own spec. The two `sendByApiContract`s, `createFallbackTransport`, the plain frontend sends and `injectByApiContract` all call it; no local copies remain.

packages/app/api-contracts/src/serverTypes.ts
CatchMe2: Added in 620b4d8f. `InferServerStatusesForKey` has cases for an exact key, a range key minus exact codes, `'default'` minus everything covered, and a contract declaring only `'default'` (equals `HttpStatusCode`). `InferServerSseSelections` is asserted assignable to `ServerSseSelection`, and `SseMessage` is checked as the shape every `SseStreamMessage` satisfies plus per-event data typing.

PR #1277, opened 2026-09-10
chore: require ioredis 6 and drop ioredis 5 from the tree

kibertoad:
## What

Moves `@lokalise/background-jobs-common` and `@lokalise/healthcheck-utils` to `redis-semaphore@5.8.0`, requires ioredis 6, and removes ioredis 5 from the dependency tree entirely.

## Why

`redis-semaphore@5.7.0` declares its peer as `ioredis: ^4.1.0 || ^5`. Both packages were already resolving it against ioredis 6, so the pairing was unsupported by the library's own metadata. `5.8.0` is the first release to add `^6` to that range.

## Changes

**redis-semaphore and the ioredis peer**

- `@lokalise/background-jobs-common`: `redis-semaphore` `^5.7.0` to `^5.8.0`
- `@lokalise/healthcheck-utils`: `redis-semaphore` `^5.6.2` to `^5.8.0`
- Both narrow their `ioredis` peer from `^5.4.1 || ^6.0.0` to `^6.0.0` (breaking, hence the major bumps)

**Getting ioredis 5 out of the tree**

One copy survived the above: `@splitsoftware/splitio` pins `ioredis: ^4.28.0 || ^5.0.0` and arrives transitively through the `@lokalise/fastify-extras` dev dependency. A workspace-level override redirects it:

```yaml
overrides:
  ioredis: ^6.0.0
```

The override is what does the work, and bumping fastify-extras does not replace it. Its latest release (34.0.1) still depends on `@splitsoftware/splitio: ^11.6.0`, and splitio's own latest (11.11.2) still carries `ioredis: ^4.28.0 || ^5.0.0`. Resolving with fastify-extras at `^34.0.1` and the override removed puts `ioredis@5.11.1` straight back in the lockfile under `@splitsoftware/splitio-commons`.

**@lokalise/fastify-extras 32.0.0 to 34.0.1**

Bumped anyway, on its own merits, in the three packages that pull it in for tests: `auth`, `aws-config`, `fastify-bullboard-plugin`. Dev dependency only, so no published surface changes and no changeset. With the override in place, `@splitsoftware/splitio-commons` resolves against `ioredis@6.0.0` and the lockfile contains no `ioredis@5` reference at all.

`minimumReleaseAgeExclude` gains `redis-semaphore@5.8.0`, which published earlier today.

## Verification

Everything below ran under pnpm 11.9.0, the version this repo declares.

- `pnpm install --frozen-lockfile` from a clean tree: passes, and `node_modules/.pnpm` contains only `ioredis@6.0.0`. Zero `ioredis@5` matches in `pnpm-lock.yaml`.
- `pnpm run build`: 33/33 packages
- `pnpm run lint`: 33/34. `@lokalise/universal-testing-utils` fails on msw response-resolver types in `MswHelper.ts` and `ApiContractMswHelper.ts`. Reproduced identically on a clean `origin/main` worktree, so it is pre-existing and unrelated; msw stays at 2.15.0 in this diff.
- `@lokalise/auth` tests: 66/66 passed
- `@lokalise/aws-config` tests: 149/149 passed
- `@lokalise/fastify-bullboard-plugin` tests: 17/17 passed
- `@lokalise/healthcheck-utils` tests: 14/14 passed
- `@lokalise/background-jobs-common` tests: green in CI on both Node 24.x and 26.x. Locally two tests fail (`ignores missing job error during automatic purging`, on `Connection is closed.` during `dispose()`), but a clean `origin/main` worktree run against the same Redis reproduces them with identical 263/2/1 counts, and CI passes on this branch, so they are a local environment artifact rather than anything this PR introduces.

All 15 CI checks pass.

## AI Assistance Tracking


pnpm-workspace.yaml
```diff
@@ -1,6 +1,11 @@
 packages:
   - "packages/dev/*"
   - "packages/app/*"
+overrides:
```
CatchMe2, line 4: How about bumping fastify-extras dep instead? v34.0.0 supports ioredis v6

kibertoad: Checked this, and the bump alone doesn't get us there.

v34.0.0 does support ioredis 6, but its peer range is `ioredis: ^5.7.0 || ^6.0.0`, which is what v32.0.0 already declares. The copy of ioredis 5 in our tree isn't coming from that peer, it comes from `@splitsoftware/splitio`, a hard dependency of fastify-extras. v34.0.1 still depends on `@splitsoftware/splitio: ^11.6.0`, and splitio's own latest (11.11.2) still pins `ioredis: ^4.28.0 || ^5.0.0`.

I resolved it both ways to confirm. With fastify-extras at `^34.0.1` and the override removed:

```
ioredis@5.11.1:
'@splitsoftware/splitio-commons@2.12.1(ioredis@5.11.1)':
  '@splitsoftware/splitio-commons': 2.12.1(ioredis@5.11.1)
```

With the override in place, the only entry left is `ioredis@6.0.0`, and `node_modules/.pnpm` holds a single copy.

That said, I don't mind bumping fastify-extras, can do so in the same PR

kibertoad: Bumped it anyway in 6050c8fb: `^32.0.0` to `^34.0.1` across `auth`, `aws-config` and `fastify-bullboard-plugin`. Dev dependency only, so no published surface changes and no changeset.

To be clear on the ioredis side, the override still has to stay. With fastify-extras now at 34.0.1 the lockfile has zero `ioredis@5` matches and `@splitsoftware/splitio-commons` resolves against `ioredis@6.0.0`, but that is the override doing it, not the bump.

Green on the three affected packages: auth 66/66, aws-config 149/149, fastify-bullboard-plugin 17/17. Full build 33/33.

CatchMe2 (approved):

PR #1284, opened 2026-09-14
fix(api-contracts): infer streaming param for dual-mode contracts

CatchMe2:
## Summary

Writing the missing dual-mode (JSON + SSE) tests for `sendByApiContract` in both HTTP clients exposed a type bug: `streaming: true` did not typecheck against a dual-mode contract, and the result type never narrowed to the SSE body.

**Root cause.** `ClientRequestParams` wrapped the conditional `streaming` member in `Prettify` together with the other params. TypeScript can infer a type parameter through a conditional type by trying both branches, but not when that conditional sits inside a mapped type over an intersection that also depends on the still-unfixed `TApiContract`. With no inference candidate, `TIsStreaming` fell back to its default (`false` for dual-mode) and the call was rejected with `Type 'true' is not assignable to type 'false'`.

**Fix.** Move `StreamingParam` outside `Prettify` so it is a direct inference site. The http client signatures are unchanged. The only visible side effect is that hovering the params type shows an intersection rather than one flat object.

## Changes

- `api-contracts/src/new/clientTypes.ts`: `StreamingParam` is now intersected outside `Prettify`.
- `api-contracts/src/new/clientTypes.spec.ts`: five flat-shape assertions now compare against `Prettify<ClientRequestParams<...>>` (same strict key/value check). New type test infers `TIsStreaming` through a generic caller mirroring `sendByApiContract`: dual infers `true`/`false`, SSE-only defaults to `true`, JSON-only defaults to `false`, and the flag is required for dual and rejected otherwise.
- `backend-http-client` and `frontend-http-client` `sendByApiContract.spec.ts`: new `dual-mode (JSON + SSE)` block. Non-streaming returns the parsed JSON body and does not send `Accept: text/event-stream`; streaming sends that header and yields typed events; a contract-defined 400 JSON error is still parsed in streaming mode.
- Changeset: patch for `@lokalise/api-contracts`.

## Follow-ups (not in this PR)

- At runtime the `streaming` flag does not constrain response resolution: with `streaming: false`, a `text/event-stream` reply still resolves to the SSE entry, so the caller gets an async iterator where the type says JSON. Treating that as `UnexpectedResponseError` is a behavior decision.
- Both READMEs reference `anyOfResponses` in the dual-mode example; that helper does not exist.

## Test plan

- [x] `pnpm test` and `pnpm lint` in `api-contracts` (847 tests), `backend-http-client` (519), `frontend-http-client` (494), with `api-contracts` rebuilt first since the clients resolve it via `dist`.

## AI Assistance Tracking


CarlosGamero (approved):
🙏

PR #1287, opened 2026-09-16
feat(pyroscope-profiling): continuous profiling for local load-test analysis

kibertoad:
## Changes

A load test says the cache refresh takes nine seconds, and a trace says which span it was in. Neither
names the function. This adds `@lokalise/pyroscope-profiling`, which turns continuous CPU, wall-clock
and heap profiling on from environment variables and ships a CLI that prints where the time went, so
answering that question takes one container, one environment variable and a terminal.

### The problem

Profiling a Node service here exists in exactly one place, inside CTENG's
`src/infrastructure/profiling/`
([content-type-app-engine#1734](https://github.com/lokalise/content-type-app-engine/pull/1734)). Two
things are wrong with that. No other service can use it, and it only ships profiles: reading one
means a deployed environment, a Grafana and a flame graph, so a question a laptop could settle in two
minutes needs a browser and an eye for rectangle widths.

The loop this is built for is the local one. Run the load test the repo already has, then ask which
function spent the time:

```
$ pnpm exec pyroscope-analyze --service my-service --from now-10m

frame (self)                                       self  share
---------------------------------------------  --------  -----
./src/cache/upsert.ts:upsertCacheFieldEntries    16.62 s  98.2%
:Garbage Collection:0                           214.6 ms   1.3%
```

With tracing on, every sample carries the route or job it was taken under, so one run can be cut per
journey (`--select 'span_name="POST /v1/content/refresh"'`) instead of averaged across all of them.
Shipping to a Grafana stack still works and is what deployed environments do; it is no longer the
only way to read a profile.

### What is in it

- `startProfiling` / `stopProfiling`, off unless `PYROSCOPE_ENABLED` says otherwise and never on
  under `NODE_ENV=test`, with the SDK imported lazily so a process running without profiling never
  loads its native binding, and every failure logged and swallowed.
- `pyroscopeProfilingPlugin` (`/fastify`): starts the profiler, labels the samples taken during each
  request with its route, flushes the last profile window on close.
- `PyroscopeSpanProcessor` (`/opentelemetry`): labels samples with the local root span they were
  taken under, which is what makes `--select 'span_name=...'` and Grafana's trace-to-profiles link
  work at all.
- `withProfilingLabels` for background jobs, which have no span to cut a profile by.
- `pyroscope-analyze`, a dependency-free CLI over Pyroscope's query API: flat self-time table by
  default, `--tree` for call paths, `--against-from` for a per-frame diff between two runs, `--json`
  for a script or a CI gate, `--folded` for flamegraph.pl.
- Compose files for Pyroscope alone, with Grafana and Tempo as overlays, and `examples/` in the
  tarball so they are reachable from `node_modules`.

### Why it looks like this

- Self time rather than cumulative in the default table. A Node wall profile's cumulative shares
  collapse across every `await`, so nearly every frame near the root reads as 100%. Self time is the
  only column left that means anything, and `--tree` covers the case where the path matters.
- No ASCII flame graph. Width-encoded rectangles round a 3% frame and a 0.3% frame to the same
  character at terminal resolution, and a number can go into a PR description.
- Sampling rates are deliberately not configurable through this package. The SDK reads them from the
  environment, and anything passed to its `init()` outranks that, so a second name for one setting
  would silently win. The README documents the variables instead.
- `ProfilingLogger` is structural rather than a `@lokalise/node-core` import, which keeps the
  package's runtime dependencies down to the SDK and `fastify-plugin`. A test pins that
  `CommonLogger`, `app.log` and the SDK's own `Logger` all still satisfy it.
- Three entry points with both peers optional, so a worker or a script can depend on this without
  installing Fastify or the tracing SDK.

### Where the risk is

`stopProfiling` runs inside an `onClose` hook and is the only thing here on a shutdown path.
Pyroscope's exporter posts with no timeout of its own and swallows its errors, hence the five-second
ceiling around it. `PyroscopeSpanProcessor.onStart` runs on every root span, so it holds a bounded
map and swallows everything: a profiler that cannot be labelled must not cost a span.

### Verified against a real Pyroscope 2.3.1

Running it rather than typechecking it corrected three things the source PR's docs had wrong or did
not mention:

- Node ingests under `wall:...`, not the `process_cpu:...` that Grafana's trace-to-profiles docs
  show, so the provisioned Tempo datasource would have queried a profile type that does not exist.
- Heap profiles only ship on the periodic flush, never on `stopProfiling`. A process that blocks its
  event loop produces no heap profile at all, and no wall profile until it exits.
- A fresh Pyroscope drops ingests for its first 60 seconds, which is the most common reason an
  otherwise correct local setup reports no samples.

Also confirmed end to end in a Linux container: the five profile types, the four default labels plus
custom ones, `withProfilingLabels` surviving an `await`, and a label-filtered flame graph. On Windows
the wall profiler cannot start, because `@datadog/pprof` compiles the SIGPROF path out on `_WIN32`
and Pyroscope's wall profiler always asks for labels; the `TypeError: Contexts are not supported.` is
logged and swallowed, the process keeps serving, and the README names WSL2 and the service's own
container as the two ways to close the loop there.

### Not in this PR

No changes to any consuming service. Adopting this in CTENG means replacing
`src/infrastructure/profiling/` with the package, which is a follow-up against that repo.

## Checklist

## AI Assistance Tracking


packages/app/pyroscope-profiling/bin/analyze.mjs
```diff
@@ -0,0 +1,429 @@
...
+  --min-share <percent>   Frames below this share of the total are left out of
+                          the tree. Default 1.
+  --json                  Machine-readable output instead of a table.
+  --folded <file>         Also write collapsed stacks, the input format
+                          flamegraph.pl and inferno take.
+`
+
+const UNIT_BY_PROFILE_TYPE = (profileType) => {
+  if (profileType.includes(':nanoseconds:')) return 'time'
+  if (profileType.includes(':bytes:')) return 'bytes'
```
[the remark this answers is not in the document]
kibertoad: Changed, though not for the reason given. `memory:inuse_objects:count:inuse_space:bytes` ends in `bytes` with no trailing colon, so `includes(':bytes:')` never matched it and `--type objects` already printed counts. The check was right by accident: the only component it could match with colons on both sides happens to be the sample unit for every type in the table. It now parses that component outright, so a full profile type id passed to `--type` cannot land on the wrong formatter, with a test on `objects`.

`1f278d82`

packages/app/pyroscope-profiling/bin/analyze.mjs
```diff
@@ -0,0 +1,429 @@
...
+  if (Number.isNaN(parsed)) throw new Error(`Cannot read a time from "${when}"`)
+  return parsed
+}
+
+const buildSelector = (service, select) =>
+  select ? `{service_name="${service}", ${select}}` : `{service_name="${service}"}`
+
+async function fetchFlamegraph(url, profileType, selector, start, end) {
+  const endpoint = `${url.replace(/\/$/, '')}/querier.v1.QuerierService/SelectMergeStacktraces`
+  const response = await fetch(endpoint, {
```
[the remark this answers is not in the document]
kibertoad: Bounded at 30s, with `--timeout <seconds>` to raise it for a range wide enough that the query itself is slow. A `TimeoutError` comes back as `Pyroscope did not answer within 30s` rather than as the abort, so the failure names the cause.

`1f278d82`

packages/app/pyroscope-profiling/examples/docker-compose.grafana.yml
```diff
@@ -0,0 +1,26 @@
...
+# own UI is on http://localhost:4040 either way.
+services:
+  grafana:
+    image: grafana/grafana:12.3.1
+    ports:
+      - '${GRAFANA_PORT:-3000}:3000'
+    environment:
+      GF_AUTH_ANONYMOUS_ENABLED: 'true'
+      GF_AUTH_ANONYMOUS_ORG_ROLE: Admin
+      GF_AUTH_DISABLE_LOGIN_FORM: 'true'
```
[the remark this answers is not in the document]
kibertoad: All three publish on `127.0.0.1` now. Nothing in the stack loses reach: Grafana talks to Pyroscope and Tempo over the compose network by service name, and the two documented Windows paths still work, since WSL2 forwards its own `localhost` to the Windows loopback and a containerised service is on the compose network anyway.

`1f278d82`

packages/app/pyroscope-profiling/examples/docker-compose.tracing.yml
```diff
@@ -0,0 +1,29 @@
+# Overlay that adds Tempo, so a span links to the flame graph taken while it was
+# open. Needs both files before it:
+#
+#   docker compose -f docker-compose.pyroscope.yml -f docker-compose.grafana.yml #                  -f docker-compose.tracing.yml up -d
```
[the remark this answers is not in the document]
kibertoad: Fixed. It had already moved from an inline `#` to a backslash continuation, which has the same end: strip the `#` markers to copy it and the second line still opens a comment. One line now.

`1f278d82`

packages/app/pyroscope-profiling/README.md
```diff
@@ -0,0 +1,748 @@
...
+  the container id under plain Docker. The first is useful; the second changes
+  on every run, so a profile filtered by it is only readable during the run that
+  produced it.
+
+For a browser locally, [`examples/`](examples) has Grafana as an overlay on the
+same Pyroscope:
+
+```bash
+docker compose -f docker-compose.pyroscope.yml -f docker-compose.grafana.yml up -d
+```
```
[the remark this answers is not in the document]
kibertoad: Both say it. The package README gets a `cd node_modules/@lokalise/pyroscope-profiling/examples` ahead of the first Compose command and points the Tempo overlay at the same directory; `examples/README.md` names the directory the paths are relative to.

`1f278d82`

packages/app/pyroscope-profiling/src/fastify.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      ENABLED_CONFIG,
+      { appEnv: 'staging' },
+      app.log,
+    )
+
+    await app.close()
+  })
+
+  it('falls back to the environment, with the app name as the default profile name', async () => {
+    const app = await buildApp({ appName: 'my-service' })
```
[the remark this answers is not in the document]
kibertoad: Stubbed in `beforeEach` and unstubbed after each test. `NODE_ENV=test` already forced `isEnabled` to false under vitest, but `PYROSCOPE_APPLICATION_NAME` had nothing holding it: exported in a shell, it replaced the `appName` the assertion is about. Verified by running the file with both variables set.

`1f278d82`

packages/app/pyroscope-profiling/bin/analyzeCommand.mjs
CatchMe2, line 1: How about using some lib to build the CLI instead of reinventing the wheel? We could use smth like https://github.com/cacjs/cac (it has good typescript support)

packages/app/pyroscope-profiling/src/fastify.ts
```diff
@@ -0,0 +1,161 @@
...
+    done()
+  })
+}
+
+export const pyroscopeProfilingPlugin = fp(plugin, {
+  fastify: '5.x',
+  name: 'pyroscope-profiling-plugin',
+})
+
+export type { ProfilingConfig, ProfilingContext, ProfilingLogger } from './types.ts'
```
CatchMe2: Any reason for reexporting it from `fastify`? Cannot we import these types directly from `types.ts` file?

packages/app/pyroscope-profiling/package.json
```diff
@@ -0,0 +1,85 @@
+{
+    "name": "@lokalise/pyroscope-profiling",
+    "version": "0.0.0",
```
CarlosGamero, line 3: 🟡 We will need to release this manually and add OIDC config on npm for the changeset to work

packages/app/pyroscope-profiling/package.json
```diff
@@ -0,0 +1,85 @@
...
+        "lint": "biome check . && tsc",
+        "lint:fix": "biome check --write",
+        "test": "vitest run",
+        "test:ci": "vitest run --coverage",
+        "prepublishOnly": "pnpm run build",
+        "package-version": "echo $npm_package_version",
+        "postversion": "biome check --write package.json"
+    },
+    "dependencies": {
+        "@pyroscope/nodejs": "0.6.3",
```
CarlosGamero, line 56: Should we pin it? I am fine doing it for what reasons as we did for open telemetry, asking just tout of curiosity

packages/app/pyroscope-profiling/package.json
```diff
@@ -0,0 +1,85 @@
...
+        "fastify-plugin": "^6.0.0"
+    },
+    "peerDependencies": {
+        "@opentelemetry/sdk-trace-base": ">=2.0.0 <3.0.0",
```
CarlosGamero: 🟢 Nit: Same as `^2.0.0`

packages/app/pyroscope-profiling/package.json
```diff
@@ -0,0 +1,85 @@
...
+            "optional": true
+        }
+    },
+    "devDependencies": {
+        "@biomejs/biome": "^2.5.12",
+        "@lokalise/biome-config": "workspace:^",
+        "@lokalise/node-core": "^14.9.1",
+        "@lokalise/tsconfig": "workspace:^",
+        "@opentelemetry/api": "1.9.1",
+        "@opentelemetry/sdk-trace-base": "2.11.0",
```
CarlosGamero: 🟢 For dev-deps we can maybe use `^`

CarlosGamero (commented):
Looked at the profiler and label-scope bookkeeping, the fastify plugin, the analyze CLI, and the API surface checked against a real @pyroscope/nodejs 0.6.3 tarball. Four items inline, none blocking. Skipped what CodeRabbit already raised.

packages/app/pyroscope-profiling/src/wallLabels.ts
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Attaches `labels` on top of whatever scopes are already open and returns a
+ * handle for {@link closeLabelScope}, or `undefined` when profiling is off.
+ *
+ * Throws what the profiler throws (the Windows wall profiler refuses labels
+ * outright), leaving no scope behind, so a caller can fall back to running its
+ * work unlabelled.
+ */
+export function openLabelScope(labels: WallLabels): number | undefined {
```
CarlosGamero, line 82: 🔶 Runtime label names skip the sanitization that init tags get. `buildTags` rewrites `service.name` to `service_name` because one bad label name gets the whole series rejected at ingest, and the exporter only says so at debug level. `withProfilingLabels({ 'tenant-id': '42' }, fn)` hits that same silent failure with no guard, so the job's samples never land while everything logs clean. Should the keys here go through the same `toLabelName`?

packages/app/pyroscope-profiling/bin/analyzeCommand.mjs
```diff
@@ -0,0 +1,611 @@
...
+    console.log(USAGE)
+    return options.help ? 0 : 1
+  }
+
+  const now = Date.now()
+  const settings = resolveSettings(options, now, env)
+
+  const flamegraph = await fetchFlamegraph(settings, settings.from, settings.until)
+  const total = Number(flamegraph.total ?? 0)
+  if (total === 0) {
```
CarlosGamero, line 592: 🔶 An empty main range returns before `fetchAgainst` runs, so a `--json` diff invocation gets output with no `against` field. A gate reading `.against` can't tell "no comparison requested" from "current run produced nothing while the baseline had samples", and the second one is the strongest regression signal it could get. `printJson`'s comment promises the same shape either way. Worth fetching the against range before bailing out here?

packages/app/pyroscope-profiling/bin/analyzeCommand.mjs
```diff
@@ -0,0 +1,611 @@
...
+  for (const row of rows) console.log(line(row))
+}
+
+/** The defaults, and the two shorthands `--type` and `--from` accept. */
+export function resolveSettings(options, now, env = process.env) {
+  const type = options.type ?? 'wall'
+  // Own property only: `PROFILE_TYPES.toString` is a function, truthy, and
+  // would be passed on as if it were a profile type id.
+  const profileType = Object.hasOwn(PROFILE_TYPES, type) ? PROFILE_TYPES[type] : type
+  const from = parseWhen(options.from ?? 'now-15m', now)
```
CarlosGamero, line 399: 🟢 Nothing checks that `from` precedes `until`. A transposed pair like `--from now --until now-15m` queries an inverted range, comes back empty, and the tool prints the flush-interval hint with exit 2, pointing the user at ingest instead of at their arguments.

packages/app/pyroscope-profiling/src/fastify.ts
```diff
@@ -0,0 +1,161 @@
...
+  })
+
+  app.addHook('onResponse', (request, _reply, done) => {
+    release(request)
+    done()
+  })
+
+  // A client that disconnects mid-request never reaches onResponse, and the
+  // scope would then be held until the open-scope bound reclaimed it.
+  app.addHook('onRequestAbort', (request, done) => {
```
CarlosGamero, line 150: 🟢 This abort path has no test in fastify.spec.ts. It's the only thing keeping a client disconnect from holding a scope open until the 1024 bound reclaims it, and a regression here (a hook signature change in a Fastify major, `release` throwing) would pass the suite and only show up in production as stale `span_name` labels reapplied to unrelated samples.

packages/app/pyroscope-profiling/bin/analyzeCommand.mjs
kibertoad: Switched to `node:util`'s `parseArgs`, which is what `drizzle-utils` already uses for its bin, so no new dependency on a published package. It covers every case the hand-rolled parser did: `--top 5` and `--top=5`, an unknown option refused rather than ignored, a positional refused, and a flag standing where a value belongs (`--folded --json`) refused rather than taken as the filename. The option table replaces the two `Set`s and the two parsing helpers.

The one gap is `--service=`, which `node:util` reads as an empty string, so the empty-value check stays.

`cac` would buy subcommands and a generated help, neither of which this has: one command, and a usage string that documents what each option means for reading a profile rather than just its type.

packages/app/pyroscope-profiling/src/fastify.ts
kibertoad: No good reason. Dropped from both `/fastify` and `/opentelemetry`: the root entry point publishes them, and importing a type from there pulls in nothing at runtime. One home per type. The README now says which entry point each comes from.

packages/app/pyroscope-profiling/package.json
kibertoad: Noted. Nothing to change in the PR then, so flagging it here for whoever cuts the first release: the changeset is a `major` on `0.0.0`, and it needs the npm OIDC trusted publisher for `@lokalise/pyroscope-profiling` in place first.

packages/app/pyroscope-profiling/package.json
kibertoad: Kept pinned, same reason as the OpenTelemetry packages. `@pyroscope/nodejs` pulls in `@datadog/pprof`, which is a native addon: a floating patch can change which prebuilt binding gets installed and what the sampler does, and the failure mode is a profile that reads differently rather than a build that breaks. Being on `0.x` it has no major to protect us either, so the version that gets tested is the version that ships, and a renovate PR makes the change visible.

packages/app/pyroscope-profiling/package.json
kibertoad: Right, they are identical here. Changed to `^2.0.0`. The long form in `opentelemetry-fastify-bootstrap` is `>=1.9.0 <1.10.0`, which a caret cannot express, which is probably where it came from.

packages/app/pyroscope-profiling/package.json
kibertoad: Floated `@opentelemetry/sdk-trace-base` to `^2.11.0`. It backs the `^2.0.0` peer range, so tests are better off following the newest 2.x.

Left `@opentelemetry/api` pinned at `1.9.1` and `typescript` at `7.0.2`, matching `opentelemetry-fastify-bootstrap` and the rest of the repo.

CarlosGamero (approved):

packages/app/pyroscope-profiling/src/wallLabels.ts
kibertoad: Yes, and they now do. `toLabelName` moved to `src/labelNames.ts` and `openLabelScope` runs every key through it, so `withProfilingLabels({ 'tenant-id': '42' }, fn)` sets `tenant_id`. Two keys that sanitize onto one name leave the last one, same as at init.

No `warn` for that collision here, unlike `buildTags`: this runs per request and per job and has no logger, and a log line on the label path would be the thing that costs more than the labelling.

Values are left alone. The `{},=` substitution exists because `init()` encodes the app name and the init tags into one `appName{key=value}` string; a wall label value goes into the pprof sample as a string and carries no structural characters.

packages/app/pyroscope-profiling/bin/analyzeCommand.mjs
kibertoad: Moved the `fetchAgainst` call above both empty checks, so `against` is in the JSON whenever `--against-from` or `--against-until` was passed, `total: 0` or not. A gate can now read a baseline with samples and a current range with none as the regression it is.

The text report gets the same signal in one line: when the compared range held something, `reportEmpty` says so and drops the flush-interval hint, which is the wrong place to send the reader in that case.

packages/app/pyroscope-profiling/bin/analyzeCommand.mjs
kibertoad: Added `assertRange`, which `resolveSettings` calls for `--from`/`--until` and `fetchAgainst` calls for the pair it derives. It names both flags and prints the two timestamps it resolved them to, so `--from now --until now-15m` exits 1 with the arguments in the message instead of exiting 2 with the flush-interval hint.

Equal bounds are refused too: a zero-length range has the same empty answer and the same wrong explanation.

packages/app/pyroscope-profiling/src/fastify.ts
kibertoad: Covered now. It needs a real socket: `light-my-request` never sets `req.aborted`, and Fastify only runs `onRequestAbort` when it is true, so nothing driven by `inject` can reach the hook. The test listens on `127.0.0.1:0`, holds a request open in the handler, aborts the client `fetch` and waits for the labels to come back to `{}`.

Checked it fails for the right reason: with the `release(request)` call taken out of the hook, it ends on `{ span_name: 'POST /v1/content/:id' }` still applied.

